PR #8, opened 2015-08-15
Fixes for GCC 3.2.3

AKuHAK2:


sp193:
Sorry if I might be missing stuff. I am now very busy with the starting of school. :/

PR #9, opened 2015-08-30
Added some fixes and improvments from Megaman's fork

AKuHAK2:
Just some fixes from Megamans fork that I forgot to add 2 years ago. Needs to check ee/libgs/src/texture.c and ee/rpc/sior/src/sior_rpc.c

### DIFF
--- a/Defs.make
+++ b/Defs.make
@@ -17,7 +17,7 @@
 # Definitions for the EE toolchain.
 #
 
-EE_TOOL_PREFIX = ee-
+EE_TOOL_PREFIX ?= ee-
 EE_CC = $(EE_TOOL_PREFIX)gcc
 EE_CXX = $(EE_TOOL_PREFIX)g++
 EE_AS = $(EE_TOOL_PREFIX)as
@@ -31,7 +31,7 @@ EE_STRIP = $(EE_TOOL_PREFIX)strip
 # Defintions for the IOP toolchain.
 #
 
-IOP_TOOL_PREFIX = iop-
+IOP_TOOL_PREFIX ?= iop-
 IOP_CC = $(IOP_TOOL_PREFIX)gcc
 IOP_AS = $(IOP_TOOL_PREFIX)as
 IOP_LD = $(IOP_TOOL_PREFIX)ld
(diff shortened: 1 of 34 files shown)


sp193:
Most of the changes look okay, but I would like you to explain why the following was done:
1. the large number of changes to the SIF stuff (what's with the DMA_ALIGN stuff?)
2. the call to restore the IMR was removed from ps2_screenshot.

For number 1, I understand that at some point one of the toolchains (AFAIK, not the official R5900 support committed to GCC) used had the stack alignment set to 8 instead of 16. But this compiler has the alignment set correctly and your "buffer" ends up having the same alignment and base address as the original structure, so I am not sure what the improvement here is (other than making the code more confusing and less like the original).

For number 2, the change you made was like this:
   // Put back prew imr and set finish event
 
-  GsPutIMR( prev_imr );
+  asm __volatile__("sync.l\n"); 
   *PS2SS_GS_CSR = PS2SS_CSR_FINISH;
 
Is it really okay for the IMR to not be restored, unlike what the comment says? I did not study how ps2_screenshot works and so I do not know if this change is definitely correct or not.

For the "check if correct" comments in the code, I am quite sure that the code will still work without the castings (as you have changed them to not use). If not, then I will refactor the code to explicitly use the right variable type.

AKuHAK2:
oh about Megaman's fixws i cant say something more reliable. I just checked out my sources and realized that this changes PROBABLY isnt related to new gcc and binutils. Of course it can be mistake from my side and DMA stuff is only needed if you use 4.9 gcc and newest binutils. So simple answer: I don't know :) If you need I can make another pull request without these changes.

sp193:
I've played with his official support for the R5900 and the SIF alignment stuff isn't required. That toolchain works very much like the current one, except that it has no support for the 2nd pipeline and MMI.
But I do remember seeing one toolchain before that had the stack alignment set to 8 bytes instead. Whatever it was, it isn't any of the compilers that we will be using and hence the changes to the SIF libraries should not be required.
Please remove the changes to the SIF libraries because they result in an unnecessary overhead. I'll very much appreciate if you'll do that and send a new pull request. :)

AKuHAK2:
Ok no problem :) I will do it

PR #10, opened 2015-09-25
Added some changes from Megaman's fork and many fixes for samples

AKuHAK2:
I changed order in sample Makefiles cause we need to recognize EE-PREFIX before compilng. I also was trying to fix some samples to be compilable. Still draw/texture sample cannot be compiled and im not pretty sure if font sample compiles in normal way. I removed all DMA_CHAIN stuff.

### DIFF
--- a/Defs.make
+++ b/Defs.make
@@ -17,7 +17,7 @@
 # Definitions for the EE toolchain.
 #
 
-EE_TOOL_PREFIX = ee-
+EE_TOOL_PREFIX ?= ee-
 EE_CC = $(EE_TOOL_PREFIX)gcc
 EE_CXX = $(EE_TOOL_PREFIX)g++
 EE_AS = $(EE_TOOL_PREFIX)as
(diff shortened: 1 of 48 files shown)


sp193:
I'll try to make a manual pull and merge, so that I can slip in some changes for you:
1. For Vu0DotProduct() under libvux/vuhw.c, I think that ret doesn't need to be initialized but the inline assembly expression should be adjusted to treat ret as output instead (GCC should then stop complaining about the uninitialized variable).
2. For GsLoadImage() under libgs/texture.c, the change is good so you need not worry. I'll remove the comment... but then I am not sure if we want to allow void pointers to be incremented. Some compilers don't allow that because it seems to be undefined behaviour under the C spec (meaning that it's non-standard C), although GCC allows that. Perhaps I'll just change the code there to avoid this ambiguity altogether.
2. The change to sior/sior_rpc.c needs to be corrected. The return value from sio_gets() is a pointer. I guess that it's NULL on failure, so dereferencing it is incorrect.

Please standby!

AKuHAK2:
Ok thanks!
I agree with all of your comments.
BTW I tested samples: cube, teapot, graph, doublebuffer, draw. mpeg, padx, pad, mtap works fine. memorycard work fine on real console but failes on pcsx2.
tcpips and ps2snd failed in both (cant load some irx files properly).
All audsrv stuff failes in real console but works in pcsx2.

sp193:
Please try again because there was a logic glitch in the change to kernel/sifrpc.c. OPL stopped working because of this, although strangely it will still sometimes work.

PR #15, opened 2015-12-30
Stop using long type

uyjulian:
using the long type causes a lot of issue between 32 and 64 bit

### DIFF
--- a/common/include/sys/time.h
+++ b/common/include/sys/time.h
@@ -20,7 +20,7 @@ typedef unsigned long long clock_t;
 #endif
 
 #ifndef __time_t_defined
-typedef unsigned long time_t;
+typedef unsigned long long time_t;
 #define __time_t_defined
 #endif
 
--- a/common/include/tamtypes.h
+++ b/common/include/tamtypes.h
@@ -22,17 +22,17 @@ typedef volatile unsigned short 	vu16;
 
 #ifdef _EE
 typedef unsigned int		u32;
-typedef unsigned long int	u64;
+typedef unsigned long long int	u64;
 typedef unsigned int		u128 __attribute__(( mode(TI) ));
 
 typedef volatile unsigned int		vu32;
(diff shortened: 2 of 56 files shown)


sp193:
Aren't the "common" types like u32 supposed to solve that IOP<-> EE code compatibility?
Not that our SDK is really great for moving code between the two...

sp193:
Why is it necessary to change long int to long long int anyway?
Especially if the code isn't going to be moved over to the other processor?

uyjulian:
@sp193 The "long" type changes to 32-bit integer on 32-bit processor, and 64-bit integer on 64-bit processor.

The "long long" and "int" types don't change.

sp193:
If you put it that way, then on a 16-bit processor, int is 16-bit. If a specific word size is desired, then perhaps it's better to use the defined data types like u32.
The EE has 128-bit registers and has a 64-bit instructions, but our current GCC compiler treats its word size as being 32-bit. long long is supposed to be at least 64-bits in length, so wouldn't assuming that it's always 64-bit not much better than assuming that long is going to be 64-bit?

sp193:
That is, if the code is going to be moved between the IOP and the EE. If the code is staying on either one, I don't see why it's wrong to write the code to be processor-specific.

sp193:
But if anyone else has any other views on this, I'm all ears.
I, as an individual, am just unsure if it's really a great idea because what you're suggesting here will mean that the programming standard of the PS2SDK will change; new code that gets committed should follow this new format, otherwise they will be "wrong".
If I'm not mistaken, your intent is to promote the code interchangeability between the EE and IOP, but I feel that it's best to not specify specific data types (i.e. long long int for 64-bit) if a specific width is required because it still wouldn't guarantee that the right data type will be used.

uyjulian:
I'm closing this since sp193's branch integrated most of my changes.

PR #21, opened 2016-08-07
Fixed samples and makefiles

AKuHAK2:
I added imports and exports into common iop makefiles for better use in external projects.

### DIFF
--- a/ee/draw/samples/cube/Makefile.sample
+++ b/ee/draw/samples/cube/Makefile.sample
@@ -12,7 +12,7 @@ EE_OBJS = cube.o
 EE_LIBS = -ldraw -lgraph -lmath3d -lmf -lpacket -ldma
 
 all: $(EE_BIN)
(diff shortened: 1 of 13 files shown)


sp193:
I don't know if we should or should not remove the -miop option from CFLAGS and LDFLAGS, for the v3.2.x toolchain. It's supposed to enable all options that are relevant to the IOP target, although (all?) the required options are also specified via Makefile rules. I suppose that it would be best to keep it there, until the existing tools are phased out... if ever.

The rules for the export and import tables should go into Rules.Make. I guess that the sample Rules.Make file doesn't have that because it was never used; there are no IOP samples.

uyjulian:
The repository containing the commit referenced in this PR seems to have disappeared "unknown repository".

sp193:
I have manually pulled over the remaining changes, as you can see over the recent commits. Thank you for your contributions!

PR #23, opened 2016-10-03
Get gcc/g++ to compile without newlib.

BoGanon:
I sanitized the standard headers so they're more POSIX standard. I removed io_common.h as part of that. I also added a few standard headers, a few standard definitions, and moved declarations to their standard spots. Then fixed up the problems from doing that, :smiley:. Some ports or external projects might need to add some include directives to compile. The functions in locale.h and raise() in signal.h are still undefined. I also changed sjis_conversion in string.c so it can be saved with utf-8 encoding safely. I then had to add a couple nonstandard headers in common/sys. With these changes, configure passes all tests and g++  can compile libsupc++ using ps2sdk's headers, :+1:.

To compile gcc-3.2.3:
In "gcc-3.2.3/libstdc++v3/Makefile.in" remove "src" from the SUBDIRS variable. That disables libstdc++ from compiling and installing.

Then the following steps to setup ps2sdk's headers.
```bash
# Toolchain directory where binutils is installed
prefix=/opt/ps2dev/ee
cd $prefix/ee
# Now in /opt/ps2dev/ee/ee
ln -s "$PS2SDKSRC/ee/libc/include" include
cd include
ln -s "$PS2SDKSRC/common/include/sys" sys
# Need errno.h with standard headers
ln -s "$PS2SDKSRC/common/include/errno.h" errno.h
# gcc's fixinc fixes it but libstdc++ still fails to compile
ln -s "$PS2SDKSRC/ee/math/include/floatlib.h math.h
```
The ps2sdk headers should be all set up, now. Then just configure and compile gcc-3.2.3 like in stage2 of the ps2toolchain script.

You might want or need the "gcc-3.2.3-remove-sysinclude.diff" patch in my ps2toolchain repository which removes the buggy "sys-include" directory mess from being created or needed during configure.

Some more info on libsupc++: http://wiki.osdev.org/Libsupcxx#Compiling_libsupc.2B.2B


BoGanon:
After thinking about it, merging this right now feels a bit premature. I want to create a proper math.h header and see if libstdc++ will compile. Then rework the way ps2sdk sets up the include directories. Then add an additional "install-headers" or similar rule to the Makefiles. Then a bit of work to make sure newlib is compatible with gcc compiled with ps2sdk.

So, this will be ready when that's done.

PR #30, opened 2017-01-07
Cleanup

uyjulian:


sp193:
There is no isdigit() function on the IOP. Checks for character types were originally done (by SCE) that way with look_ctype_table(), although we are currently missing the macros to alias isdigit with a corresponding call to look_ctype_table() like in the PSPSDK.

But why do you want to change all the Makefiles, to move the setting of variables like IOP_BIN lower? I mean, I would understand if you wanted to standardize where those lines are located because the majority of them were located in that position, but this seems to be a total change to a new position?

There seems to be also a change in how export and import tables are generated. Previously, some temporary files were generated, compiled and then deleted, but now they seem to be left there?
If you edited a file like imports.lst now and ran _make_ without _make clean_, would Make still know that you edited imports.lst?

The number of tab stops of your editor is probably different from some of ours, so maybe that's also why some of your changes are changing the level of identation of some files. Even in the online Github viewer, the new lines are not aligned.

It's a rather large change in coding standards (structure of Makefiles), so I don't think that I can say yes/no to this pull request alone, sorry. :/
However, the rest seem to make sense to me. Like for example, you want to move the documents into standardized folders and to add missing/remove stray spaces.

uyjulian:
> There is no isdigit() function on the IOP. Checks for character types were originally done (by SCE) that way with look_ctype_table(), although we are currently missing the macros to alias isdigit with a corresponding call to look_ctype_table() like in the PSPSDK.  

The macros in ctype.h don't use look_ctype_table. I think that will be added at a later time.  

> But why do you want to change all the Makefiles, to move the setting of variables like IOP_BIN lower? I mean, I would understand if you wanted to standardize where those lines are located because the majority of them were located in that position, but this seems to be a total change to a new position?  

I pushed new commits IOP_BIN and EE_LIB are located higher.  

> There seems to be also a change in how export and import tables are generated. Previously, some temporary files were generated, compiled and then deleted, but now they seem to be left there?
If you edited a file like imports.lst now and ran make without make clean, would Make still know that you edited imports.lst?  

I pushed new commits to set the temporary files to be deleted using the . INTERMEDIATE target. 
 
> The number of tab stops of your editor is probably different from some of ours, so maybe that's also why some of your changes are changing the level of identation of some files. Even in the online Github viewer, the new lines are not aligned.  

There were some lines that skipped indentation, so I manually changed indentation of some files to their correct indentation.  
My editor (Sublime Text 2) auto-detects tab/spaces setting, but sometimes it gets confused if there is a mix of tabs and spaces.

sp193:
Thanks for your contributions!
Also, a belated Happy New Year to you.

uyjulian:
You too :-)

PR #33, opened 2017-01-15
General cleanup 2

uyjulian:


sp193:
I am not sure what size_t is defined as on both processors, but unless they are unsigned int, they won't have the same meaning on both the EE and IOP. Is it always guaranteed to be unsigned int, regardless of the version of GCC?

Regarding the integer to pointer cast warning, perhaps u8* would be more suitable as a fix? Unless you're certain that size_t is going to be always of the same size of a pointer.

uyjulian:
`size_t` is the native pointer size of each processor, so on the EE, it would be 64-bit, while on the IOP, it would be 32-bit.  
In commit 27cd887, the size_t changes were needed because otherwise, the `for` loop would iterate forever on non-32-bit systems (see https://github.com/uyjulian/pfsshell/commit/7bd15910871a9f0c8a787a5540b38adfd9427993#diff-90c1a01ee8e49069b0f5974be3d1e8bdL65 )

sp193:
Okay, thanks for the clarification. Aren't pointers on the EE is 32-bit though? I remember that it is mostly 64-bit, but has a 32-bit bus and has 128-bit GPRs.

So regarding the changes to ioman, wouldn't it be better to keep the u32 as u32, instead of using size_t?
The fixes for SMBMAN are desirable.

Sent from my Sony Xperia™ smartphone

uyjulian:
> Okay, thanks for the clarification. Aren't pointers on the EE is 32-bit though? I remember that it is mostly 64-bit, but has a 32-bit bus and has 128-bit GPRs.

I haven't checked yet, so I didn't make any changes to EE-side filexio.

> So regarding the changes to ioman, wouldn't it be better to keep the u32 as u32, instead of using size_t?  

For the changes to ioman, the functions `mount`, `devctl`, `readlink`, and `ioctl2` already have size_t as arguments. I just made the changes to `iop_device_ops_t` accordingly.

sp193:
So size_t was already there in some declarations? In that case, I believe that it would believe that it is better to use u32 instead (i.e. change size_t to u32 instead). I guess that the regular ioctl2 and devctl functions in *nix uses size_t (and the iomanx developers wanted to make them compliant), but these could be viewed as non-standard functions as well. After all, they even had sce* prefixed to them.

Ideally, the EE and IOP sides should have similar declarations, so that it is possible to transfer a program between the two more easily.

There should eventually be some solution to the ioman structures (ioman vs iomanx) as well, so eventually I would like to standardise the function declarations between the two (but I am still planning how to). After all, iomanx exists because there was no way to replace IOMAN and it is supposed to be a newer version of IOMAN (hence both forward and backward-compatible). This is what I was planning to do.

As for the use of size_t to perform pointer arithmetic, could you please use u8* instead? Then there would be no need to consider whether size_t is equal to the size of a pointer.

Sent from my Sony Xperia™ smartphone

uyjulian:
I dropped the `size_t` commits.  
I'll figure out a proper solution to the `size_t` issue in the next round of my cleanups.

sp193:
So you don't agree with what I wrote? If you have a better idea, we can discuss first before anything is implemented.

But we should remain binary-compatible with existing software. We do interface SCE modules (from ROM) and also from games (for projects like OPL).
And by doing so, we can also fall back to using the official documentation, when freely-available documentation is unavailable.

Sent from my Sony Xperia™ smartphone

uyjulian:
I think it is best to use `u32` types, but I'm out of free time now, so I'll do my next round of cleanups next weekend.

PR #34, opened 2017-02-04
General cleanup 3

uyjulian:
I cleaned up the buildsystem by quite a bit.
Running `make` with multiple jobs (example: `make -j8`) no longer crashes and burns randomly :-)


uyjulian:
Okay, should be good now.

sp193:
I don't know if you saw my comment on the libsd changes, but I was asking if you checked to see which definitions were the correct ones. For example, sceSdProcBatch returned an int in one version, while there was another that returned unsigned int.

You may also want to turn freesd.h into an internal header file, since we can include libsd.h and freesd.h no longer includes anything of value.

uyjulian:
I checked the definitions. Some of them had the incorrect type.
I'll probably remove `freesd.h` entirely.

I'll probably push my changes within 15 hours of this post.

sp193:
@AKuHAK the reason why tlbsrc and osdsrc were pre-built, was because there was a risk of the location of _start changing with the content. Back then, I did not have a clear understanding why that happened, but I found the right way to prevent that in 2015 (but the build process for it wasn't changed).

@uyjulian if you are sure that the current libsd definitions are the correct versions, then that's what I am looking for. There is no hurry to push out these updates, really! So please take your time. D:

uyjulian:
I checked the libsd definitions, and converted `freesd.h` header to an internal one.
Everything should be good now.

PR #35, opened 2017-02-09
Some makefile fixes for proper tool version recognition

AKuHAK:
Some more friendly setup for compatibility with modern toolchain. BTW maybe it is possible to port more changes from uyjulians [fork](https://github.com/ps2dev/ps2sdk/compare/master...uyjulian:uyjworking)?

### DIFF
--- a/ee/Rules.make
+++ b/ee/Rules.make
@@ -29,11 +29,16 @@ EE_C_COMPILE = $(EE_CC) $(EE_CFLAGS) $(EE_INCS)
 EE_CXX_COMPILE = $(EE_CXX) $(EE_CXXFLAGS) $(EE_INCS)
 
 # Extra macro for disabling the automatic inclusion of the built-in CRT object(s)
-ifeq ($(IOP_CC_VERSION),3.2.2)
+ifeq ($(EE_CC_VERSION),3.2.2)
 EE_NO_CRT = -mno-crt0
 endif
-ifeq ($(IOP_CC_VERSION),3.2.3)
+ifeq ($(EE_CC_VERSION),3.2.3)
 EE_NO_CRT = -mno-crt0
+endif
+ifneq ($(EE_CC_VERSION),3.2.2)
+ifneq ($(EE_CC_VERSION),3.2.3)
+EE_NO_CRT = -nostartfiles
+endif
 endif
 
 $(EE_OBJS_DIR)%.o: $(EE_SRC_DIR)%.c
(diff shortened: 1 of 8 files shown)


sp193:
Those changes that you have shown, are toolchain-specific. I know because some of them were from my fork of ps2sdk.
The only one that should be brought over, is likely the one for tlbsrc. I made that change (as a bug fix) because there was nothing for preventing _start() from moving around.

sp193:
For erl-loader/Makefile, the changes that you have made are already part of ee/Rules.make. So you may want to use those instead of duplicating them.

AKuHAK:
Yes I understand - I mean add something like IOP_CC_VERSION checks every time we met toolchain specific change.

AKuHAK:
But why erl hase his own makefile?

sp193:
Honestly, I don't know what to do about support for the new toolchains. You may have noticed that I have recently attempted to port the changes over from my fork for GCC v5, but it cannot be denied that no modern-day toolchain is complete and can compile the PS2SDK cleanly without workarounds (for the incompatible function declarations). The syntax for the VU assembler is also different, and for reasons I don't know why gas doesn't seem to accept $t0-$t3 anymore, even though GCC still accepts $t0-$t3 for backward-compatibility...

So while you can adapt the code to work on various GCC versions, it may not be without a great mess and may not even be worthwhile at the moment (particularly if the version of the toolchain is getting changed).

BTW, the changes to ATAD of my PS2SDK fork has a mistake, so it likely isn't complete. I plan to delete my ps2sdk repository in favour of this one, but there are still some fixes that I have not ported over, as you have observed.

sp193:
What do you mean by "erl has its own Makefile"? Each project has their own Makefiles, but they are not always exactly the same.

AKuHAK:
I mean exactly about erl - that it is compiled as elf so it has almost full makefile not skeleton like other modules.
Of course it can be tested but some variables didnt moved from ee-rules into low level makefiles. I remember this when I was trying to get rid of some strings in makefiles. For example IOP_OBJ_DIR cannot be defined in iop.rules but only in main makefiles (dunno know why).

sp193:
Okay, if you couldn't get it to work, then this will do for now. I think that @uyjulian is suggesting such a change for the IOP Rules.make: https://github.com/ps2dev/ps2sdk/pull/34/files#diff-69abad6f532cd47cc0028f9d52b719a0

Right now, I'm trying to ensure that the quality of the libsd changes are okay before accepting the pull request.
I'm also not convinced that it is a good idea to include the definition of the functions that we don't have (i.e. not implemented in libcdvd on the EE) or for IOP kernel versions that we don't usually have access to.
But if he wants to do that for completeness, I don't know... there is a conflict of interest here.

PR #40, opened 2017-05-02
Some keyboards make control endpoint halted when requesting iManufact…

belek666:
…urer and iProduct strings

### DIFF
--- a/iop/usb/keyboard/src/ps2kbd.c
+++ b/iop/usb/keyboard/src/ps2kbd.c
@@ -267,6 +267,7 @@ int ps2kbd_connect(int devId)
       currDev->packetSize = sizeof(kbd_data_recv);
     }
 
+#if 0
   if(dev->iManufacturer != 0)
     {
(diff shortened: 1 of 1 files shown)


sp193:
It appears that the code is already doing a valid check on whether iManufacturer and iProduct do exist. There may be a mistake in [this line](https://github.com/ps2dev/ps2sdk/blob/master/iop/usb/keyboard/src/ps2kbd.c#L362), whereby the language value is set to 0 instead of 0x0409 (English). Can you try this instead?

I haven't read the full USB specifications, but the few pages that I read about reading a string descriptor indicate that a language should be specified in wIndex.

belek666:
0x0409 in wIndex fixed problem. Strange that original ps2 keyboard works without this value and returns strings correctly. Thanks for finding what was really wrong.

sp193:
Great to hear that it worked!

I believe that a wIndex of 0 will cause a list of available languages to be listed, instead of returning a string in English. So even if the official keyboard accepts it, I don't think it's supposed to work that way.

If you can, please update this pull request to have the new change and I shall merge it.

PR #45, opened 2017-07-01
Fix C++ support, libkernel depends on libpatches

rickgaiser:
This fixes a missing dependency to sbv_patch_enable_lmb when using C++.

Is the dependency from libkernel to sbv/libpatches intended?
Is this the correct fix and are there any other Makefiles that need to be updated?

### DIFF
--- a/samples/Makefile.eeglobal_cpp_sample
+++ b/samples/Makefile.eeglobal_cpp_sample
@@ -31,7 +31,7 @@ EE_ASFLAGS := -G0 $(EE_ASFLAGS)
 EE_LIBS += -lstdc++ \
 	-Wl,--whole-archive $(PS2SDK)/ee/lib/libc.a -Wl,--no-whole-archive \
 	$(PS2DEV)/ee/ee/lib/libc.a \
-	-Wl,--whole-archive -lkernel -Wl,--no-whole-archive
+	-Wl,--whole-archive -lkernel -Wl,--no-whole-archive -lpatches
 
 # Externally defined variables: EE_BIN, EE_OBJS, EE_LIB
 


sp193:
If you use the new SifIopRebootBuffer or SifIopRebootBufferEncrypted, yes.
Otherwise, there is no need for libpatches when libkernel is used.

sp193:
Oh okay, I think I know what is happening. I guess this means that the
whole of libkernel is actually being linked to.

Hmm. This is going to be difficult to solve nicely because this old
workaround that we have used for C++ is finally causing problems. Unless
you consider the linking of everything within libkernel as being
acceptable. But it could be one solution because it was like that for eons.
What do you think?

sp193:
Or perhaps I could move the new IOP reboot functions into a separate
library. It would be strange because they will be separated from the IOP
reboot functions, but I guess that would also solve the ugliness with
linking libkernel and libpatches: - lkernel -lpatches - lkernel

In the original, the patch for LoadModuleBuffer was part of libkrnl, hence
they never had this problem.

rickgaiser:
This patch is a workaround for the old workaround. It would be better to fix the whole C++ linking problem, but I don't know how to fix that, so I'm ok with another workaround and linking in the whole libkernel.

Perhaps it is possible to detect C++ compilation problems sooner with travis-ci? I see the build is succesfull, but it does not compile the samples directory. The C++ example in the samples directory does compile to erl, but when changed to elf it does not compile. There is also other samples that do not compile at all so I think adding the samples directory would be a good check.

sp193:
Sorry, but you need to ask someone else about the Travis stuff. I only know
that it does automatic compilations to test for the conpilability of our
work, but I know nothing else about it.

If I am not mistaken, our CPP sample was not even a real CPP sample that
used the global file that you adjusted. :(
(or something along the lines of that, which was why it also stayed broken
for a long time)

As for the issue at hand, I will move the new IOP reboot code into another
library. Then it should make everyone happier. Even Makefiles.

rickgaiser:
Ok thank you.

PR #50, opened 2017-07-26
General cleanup 5

uyjulian:


sp193:
Am I right to say that the "alternative names" are for backward-compatibility?
Do you know how well this will work on platforms that do not support symlinks like Windows? If I am not wrong, ln (from Cygwin etc) will create a copy of each file in Windows.

As long as it works fine there too, I (for one) think these changes are okay.

uyjulian:
Yes, the "alternative names" are for backward compatibility. Cygwin supports symlinks fine.

sp193:
Thank you. I've merged this.

PR #54, opened 2017-11-21
Add d_type value in the dirent struct

fjtrujy:
I think that to add this new value in the struct is so useful. Usually this is the common architecture.

For instance, right now I'm working in Picodrive emulator which is shared code between a lot of platform.
I have need to create custom dirent struct because should feet with the rest of platform.

Thanks in advance, and sorry if this is a stupid idea.
Francisco Javier

### DIFF
--- a/ee/libc/include/dirent.h
+++ b/ee/libc/include/dirent.h
@@ -20,6 +20,7 @@ struct dirent
 {
 	/** relative filename */
 	char d_name[256];
+	unsigned int d_type;
 };
 
 typedef struct DIR


sp193:
I am not well-versed with this POSIX structure, but I can say that our
native I/O libraries for this homebrew SDK do not use that structure.

According to this:
https://stackoverflow.com/questions/12991334/members-of-dirent-structure
The POSIX standard only has the d_name attribute specified. So I am not
sure if adding d_type is the right way forward...

sp193:
Are you saying that this field is defined on Linux? To clarify, I think adding this field is a very simple fix. But if it is not a standard feature, we may end up supporting a bad coding practice.

uyjulian:
Bad idea. `dirent->d_type` is not defined in the POSIX standard.  
I would advise using `iox_stat_t->mode` instead. It appears that picodrive already has OS-specific code, so shouldn't it be trivial to make picodrive use the native PS2 I/O libraries? The PSP code should be very similar.

fjtrujy:
Hello,

Thanks for your quick reply.

As I said before I'm not an expert here, I'm just a very beginner guy, so not sure if the MR was going to be a good idea or totally the opposite.
I can understand that if the idea is to follow the POSIX standard, and it says that the dirent struct just contains the field `d_name`, we need to follow it and close the MR.

Regarding @uyjulian Picodrive for PS2 is not trivial stuff.
1st. The PSP version is not longer supported, so basically in that case Notaz forgot about that platform 9 years ago with version 1.51. 
2nd. It is true that there is a fork for PSP that right now is alive, but this guy took the version 1.51 and just update there the Pico base code, and not the front-end, and input stuffs. 
My idea is to have something that could be merged in the main Picodrive project.

Anyway I would like to explain why I have created this MR, and see if you can advice me (just if you want guys).

Picodrive is composed into 2 different section:
1st. Is where it has a lot of common things between the different architectures.
2nd. Where it has specifics  things per architecture. 

In this case, all the method regarding the folder/files manage is on the common files.
So in order to make it work I had 2 solution:
1st. Make a lot of pre-processor clauses, like `#ifdef _EE.... #else .....` on the common clases, something that following clean code rules, it should be forbidden.
2nd. Modify the original `dirent` structure, and then keep the common code as it was.

Following the 2nd option (I love the clean code), I created the MR. 

So, if I understood properly you @uyjulian and @sp193  are suggesting me to create something else that should fit with all the POSIX standard platforms, right? 

Again guys, thank you very much for your effort, you are amazing creating these SDKs, and sorry if I have bothered you.

Best regards,
Francisco Javier.

uyjulian:
The only places where I see `d_type` in picodrive's code is `platform/gizmondo/menu.c` and `platform/psp/menu.c`. It doesn't look like any common code has `d_type`.

fjtrujy:
Hello @uyjulian,
Thank you for your suggestion. Probably you're right.

I'm going to explain what I'm doing, from the latest versión that was compatible for PSP, it was version 1.51, I'm trying to merge step by step all the commit from main Picodrive repository. 
I'm doing that because I'm not an advanced user of PS2 development, so I'm learning following this approach. With my current skills would be impossible to create from the scratch from the current Master branch a PS2 port.

Right now I'm about the version 1.56, and there are a lot of places where he is using that.
Could be that in the current status of Master version 1.92 he (Notaz) has refactored already how he is dealing with the folders and file.

> Why do you need d_type?

He used to detect if a concrete path is a folder or it is a file, to navigate deep or basically select the file.

Thanks you very much again for your time.
It has bee so useful.

Francisco Javier

sp193:
Hey, there is no harm in asking! It is possible that what you saw, is a platform-specific thing, since @uyjulian saw that it's only occuring in the PSP and gizmondo ports. But because those are platform-specific files, you do not need to follow them exactly.

When I wrote the v1.51b port for the PS2, only the core files written in assembly were truly ports to the PS2. I think I replaced most of the platform-specific code.

fjtrujy:
Well guys,
Anyway I think that we can close the MR, as you said before is not following POSIX structure.


Thanks for your effort

PR #55, opened 2017-11-22
General cleanup 6

uyjulian:


ee/rpc/pad/src/erl-support.c
```diff
@@ -12,7 +12,11 @@
 
 #include <erl.h>
 
+#ifdef _PADX
```
rickgaiser: Shouldn't this be _XPAD instead of _PADX ?

rickgaiser:
I left a small note that I think should be changed, but other than that I think this change looks great. It makes the code a lot smaller and more managable with the same functionality.

uyjulian:
I fixed the typo.

PR #57, opened 2018-01-13
Workarounds in DEV9 & ATAD for bad GameStar adaptors

sp193:
This pull request contains workarounds in DEV9 & ATAD for bad GameStar adaptors.

As it appears that the current method of detecting a GameStar is not so effective, please **do not** merge this pull request yet.


sp193:
The new check will check register 0x20 for 0x0001. Officially, access to this register are done with 8-bit accesses. However, some compatibles appear to have a 0x01 for the lower 8 bits, but have a random value in the upper 8 bits.
Since official adaptors appear to have a 0x0001 there anyway, maybe it is okay to perform a 16-bit check instead.

I made ATAD print "Compatible adaptor detected." if the check passes. If any official adaptors triggers this patch, please create a new issue ticket.

PR #64, opened 2018-06-07
USBHDFSD - I/O Performance Enhancement

sp193:
Modified the FAT driver to allow for long reads and writes.
To make this work with its cache, I made the FAT driver do a cache writeback and invalidation before directly accessing the device.
The cache will hence be only used for caching accesses to the FAT itself.
The cluster size will now hence matter for performance, since the larger the clusters, the longer the possible continuous reads.

The code within usb_driver.c that does the actual sector I/O has also been changed to continue the I/O operation from the USB callback, instead of getting the main thread to do it - which adds overhead. @rickgaiser commented that doing this instead does improve performance.

Finally, I moved the code from configureNextDevice into mass_store_mount and the connect function itself.
I remember somebody writing somewhere that the device cannot be configured from the connect function, but I noticed quite a long time ago that it does work. I made the change and it still works... so this would allow open() to work quicker.

Special thanks to @rickgaiser for the ideas and guidance.

### DIFF
--- a/common/include/usbhdfsd-common.h
+++ b/common/include/usbhdfsd-common.h
@@ -26,8 +26,6 @@
 //Device status bits.
 /** CONNected */
 #define USBMASS_DEV_STAT_CONN	0x01
-/** CONFigured */
-#define USBMASS_DEV_STAT_CONF	0x02
 /** ERRor */
 #define USBMASS_DEV_STAT_ERR	0x80
 
--- a/iop/usb/usbhdfsd/src/fat_driver.c
+++ b/iop/usb/usbhdfsd/src/fat_driver.c
@@ -1008,12 +1008,6 @@ fat_driver * fat_getData(int device)
 	if (device >= NUM_DRIVES)
 		return NULL;
 
-	while (g_fatd[device] == NULL || g_fatd[device]->dev == NULL)
-	{
-		if (mass_stor_configureNextDevice() <= 0)
-			break;
-	}
-
 	if (g_fatd[device] == NULL || g_fatd[device]->dev == NULL)
 		return NULL;
 	else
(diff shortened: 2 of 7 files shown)


sp193:
Update: I was wrong. We cannot do a device configuration from the connect callback. Doing so, will result in a circular wait since the configuration step does a control transfer.
It must have worked previously for me, due to sheer luck. I've made a revert of that commit.

rickgaiser:
I've tested this change with opl. It stops loading in the GUI while trying to load the first game cover. Re-inserting the usb stick works, but then it stops loading again after trying to load the first cover.

rickgaiser:
Yes, I have the latest of both:
ps2sdk: 7f21832a: make clean all release
opl: 6c52ef6: make clean release

sp193:
Oh. It seems like we've hit a problem: since we no longer separate the destination buffer from the hardware, so we now expose the driver to the hardware errata.
Under the right conditions, the SCSI read function is passed an unaligned buffer.

If I open the pipe with sceUsbdOpenPipe instead of sceUsbdOpenPipeAligned, then it works. Do you have a better suggestion?
What sceUsbdOpenPipe does (AFAIK), is to internally limit the frame size to 62 bytes, if the address passed is unaligned and is either 63/64-bytes in length.

sp193:
To add on, using sceUsbdOpenPipe() is the official Sony workaround for this errata. They remarked that USB-compliant devices should have no problems if the frame length is 62 instead of 64, but sceUsbdOpenPipeAligned() was provided to allow the developer to avoid compatibility problems and performance loss , if desired (but the conditions for the bug must not be met).

Personally, I think it is quite acceptable. We'll trade off 3% of the frame, to avoid memcpy().

rickgaiser:
I just tested this again and everything seems to work fine now.

PR #84, opened 2018-11-30
Allow other compilers than gcc to build.

zappepappe:


sp193:
I understand that there can be different Make tools that are not quite the same, hence the previous set of patches from you. 

But isn't GCC just gcc on Linux, while other platforms may have some other toolchain instead (but the compiler is not named cc)? From what I understand, cc was used with Unix.
Which platform(s), would this patch offer compatibility for?

zappepappe:
cc is always the platforms standard C compiler. On FreeBSD that is Clang, on most Linux it is GCC. When CC = gcc you have to use gcc even on FreeBSD. While Clang is able to compile the SDK just fine.

sp193:
I see. I wonder why none of the Mac users had issues though. Thanks.

zappepappe:
Not sure if it's a symlink, hardlink or complete binary, but Apple has /usr/bin/gcc that is just a front for Clang. I believe they did so to not break everything that assumed GCC.

PR #86, opened 2018-12-30
Removed doxygen config filename in the Makefile

PGAnonamous:


PGAnonamous:
> You can squash commits then force push; see here for more details: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

Done, thank you for that information uyjulian. That is the first time I have gone through that process without it taking several hours to finally get right.

PR #87, opened 2019-01-17
Updated Multitap Page.

Based-Skid:
This Page Did Not Have Any Mention about The use of Memory Cards With The MultiTap.

Updated The information Regarding the XModule Requirement and Provided Some information on Software Replacement Modules.

Fixed the Incorrect Slot Number Listing For The D Port in the Diagram of the console ports.
Supplemented This Information With The Full Listing of MultiTap Ports, 
Provided the Full port and slot listing By its Letter

Replaced The Multitap detection Example Code With Code From The Mass Format Utility Project
https://github.com/Based-Skid/PS2-Memory-Card-Formatter/blob/master/main.c#L743

### DIFF
--- a/ee/rpc/multitap/README
+++ b/ee/rpc/multitap/README
@@ -75,7 +75,7 @@ Meaning that Mtap Port 2 is Still Memory Card Port 0 (Console Port 1)
 
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 //MTAP Controller Ports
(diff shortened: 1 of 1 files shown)


AKuHAK:
@Based-Skid could you please squash all commits into one before creating pull request?

sp193:
I did it for him.

Yes, we must not mix module sets of different versions. Even if it was to miraculously work, there may be problems.

Based-Skid:
I Was Just About to do this. Thanks @sp193

PR #89, opened 2019-01-19
Cleanup Page and Updated Sample Code.

Based-Skid:
Ive Updated The Page to Use Markdown and Have Also Updated the C Code.


sp193:
Thanks for updating the README, but what are mtaphelper.c & mtaphelper.h for?

If you meant to extend the functionality of libmtap, we should try to keep things as close as possible to the official documentation.

Based-Skid:
mtaphelper.c is a Simple set of Functions I wrote to Handle Opening and Closing the Multitap ports Automatically. Baasically you can Include mtaphelper.h and the C in your own project and Calling mtGO() and it will Handle Detection of the Multitap on all Ports Automatically.



after initializing libmtap you can call mtGO() and it will Check All the Mulitap ports for connections and close ports if mtap is not present in that port. mtGO() also Stores a Variable When its run the First Time so that it knows to Close MTAP ports before trying to open them again, This is so we dont accidentally crash or cause the PS2 to bug out by trying to open an already opened mtap port.

If for some Reason you didnt want to use mtGO() you can use mtDetect() as mtGO is just a for loop

the mtDetect() call takes a number between 0 and 3 that allows for you to handle the Multitap connection on a specific port with a single call. The Function also handles Closing the Port automatically if a Multitap is not Present.

eg
mtDetect(1); would Open MTAP Port 1 then it checks the connection and if the multitap is not present it closes the port.

Based-Skid:
On a Side note i have written you an Email about MCTOOLS in FMCBInstaller.

Im working on another Multitap application to Dump Memory Card Images and Im not Sure how Some of your Code For the Dumping Process works

sp193:
Yeah, I got it. Just that I have been really busy with things in life. :|

But I think you should write your own module. I can explain what MCTools is about and how it works later, but you will probably want to replace it because your requirements are vastly different.
It was just a big thing for me in 2011, as I was still new at PS2 development.

Issue #91, opened 2019-01-19
Fixing -Wnonnull-compare


AKuHAK:
I am not completely sure for this one warning:
```c
src/freemtap.c: In function 'update_slot_numbers_thread':
src/freemtap.c:158:33: warning: bitwise comparison always evaluates to false [-Wtautological-compare]
   if((resbits & EF_EXIT_THREAD) == 1)
                                 ^~
/ps2sdk/iop/system/mtapman/src/freemtap.c: In function 'update_slot_numbers_thread':
/ps2sdk/iop/system/mtapman/src/freemtap.c:158:33: warning: bitwise comparison always evaluates to false [-Wtautological-compare]
   if((resbits & EF_EXIT_THREAD) == 1)
                                 ^~
```
Maybe this should be fixed as descibed [here](https://github.com/Proxmark/proxmark3/issues/605#issuecomment-387993326)?:
```c
if((resbits & EF_EXIT_THREAD) == EF_EXIT_THREAD)
```

sp193:
It's not just a warning, but it is incorrect logic. So by merging this PR, we have fixed this bug in MTAPMAN.

PR #92, opened 2019-01-20
Fixed some warnings captured by modern toolchain

AKuHAK:
This will fix also #91 
I hope that changes in USB function are safe.
I am not really sure if these commits should be squashed or no so I left it to devs.


sp193:
I think it's supposed to be that way. Looks more correct anyway.

You didn't just fix warnings in mtapman BTW, that was incorrect logic that you fixed.

AKuHAK:
Yes this warning lead to major problem in mtapman logic. Just interesting what was fixed :D

PR #99, opened 2019-02-18
General Cleanup 7

uyjulian:


sp193:
If you remove the F_* definitions, then how can we include only the function that we want?

uyjulian:
I replaced those specifics with ifdefs to `KERNEL_NO_PATCHES` and `_XCDVD`.

sp193:
The original structure of those libraries and their Makefiles was to group the functions and data (typically one function per object) . This would allow only the required functions to be linked, rather than causing all functions within the file to be linked.

Unless, you have a substitute to this?

uyjulian:
Wouldn't only the required functions be linked, and the extra functions be removed on strip?

sp193:
No. You can try it yourself.
Stripping only removes things like symbols and relocations, but the code will remain.

I know that with newer versions of GCC, it can optimize and remove unused code. But the earlier tools we use, do not have such a capability.

uyjulian:
Ah. I'll drop those commits then.

PR #100, opened 2019-02-18
Set 'noreturn' attribute for longjump()

astrelsky:


sp193:
Why did you change longjmp() to return an integer?

uyjulian:
longjmp is supposed to return no value, as seen in the standard C specification:
`     void
     longjmp(jmp_buf env, int val);`

astrelsky:
I... I don't know. I realized the contradiction last night and will fix it
this afternoon.

astrelsky:
It appears that I saw the li a0, 1. Mov v0,a0 and then had a brain fart and
went in it's returning and int even though I was initially giving it the no
return attribute.

PR #105, opened 2020-01-15
iomanX: Added support for PSX's io_stat_t.

AKuHAK:
I just didn't see any possible side effects inside these changes. It seems that this modification only adds support for xfrom device from PSX DVR, so as for me this can be safely merged with the mainstream.

### DIFF
--- a/iop/kernel/include/iomanX.h
+++ b/iop/kernel/include/iomanX.h
@@ -31,7 +31,8 @@
 #define IOP_DT_FS	0x10
 #ifndef IOMAN_NO_EXTENDED
 /** Supports calls after chstat().  */
-#define IOP_DT_FSEXT	0x10000000	
+#define IOP_DT_FSEXT	0x10000000
+#define IOP_DT_PSX		0x20000000
 #endif
 
 /** File objects passed to driver operations.  */
--- a/iop/system/iomanx/src/iomanX.c
+++ b/iop/system/iomanx/src/iomanX.c
@@ -74,6 +74,9 @@ int AddDrv(iop_device_t *device)
 {
 	int i, res = -1;
     int oldIntr;
+	
+	if (!strcmp(device->name, "xfrom"))
+		device->type |= IOP_DT_PSX;
 
     CpuSuspendIntr(&oldIntr);
 
@@ -387,11 +390,11 @@ int mode2modex(int mode)
 {
 	int modex = 0;
 
-	if (FIO_SO_ISLNK(mode))
+	if (mode & FIO_SO_IFLNK)
 		modex |= FIO_S_IFLNK;
-	if (FIO_SO_ISREG(mode))
+	if (mode & FIO_SO_IFREG)
 		modex |= FIO_S_IFREG;
-	if (FIO_SO_ISDIR(mode))
+	if (mode & FIO_SO_IFDIR)
 		modex |= FIO_S_IFDIR;
 
 	/* Convert the file access modes.  */
@@ -409,11 +412,11 @@ int modex2mode(int modex)
 {
 	int mode = 0;
 
-	if (FIO_S_ISLNK(modex))
+	if (modex & FIO_S_IFLNK)
 		mode |= FIO_SO_IFLNK;
-	if (FIO_S_ISREG(modex))
+	if (modex & FIO_S_IFREG)
 		mode |= FIO_SO_IFREG;
-	if (FIO_S_ISDIR(modex))
+	if (modex & FIO_S_IFDIR)
 		mode |= FIO_SO_IFDIR;
 
 	/* Convert the file access modes.  */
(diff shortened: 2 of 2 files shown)


sp193:
I think we need to think carefully about this. It seems like a hack to again standardize the file mode flags, which somehow don't fit with the current code. But yet, this doesn't seem to be how Sony designed things.
A year or two ago, I believe Ragnarok2040 (@BoGanon) mentioned that the mode2modex stuff weas added to support LaunchELF - which I think makes sense. If you disassemble a newer version of the Sony IOMAN module, you will not find any code for converting the file mode bits.

If you went around the Sony PS2SDK, you may notice that the memory card file mode bits are different from most of the newer devices. These same bits are the FIO_SO* bits that we know of. I remember that the Sony naming made more sense - although I've already forgotten how they were named. I think it's specific to the memory card driver.

To me, it means that the interpretation of the file mode bits depends on the device driver. It may have been that the memory card modules were designed and developed a really long time ago, that Sony decided to keep the old bit definitions for both forward and backward compatibility. Newer modules would use the new functions and also support the new file mode bits.

In conclusion, if the original mode to modex conversion was a hack, I think this would be a hack of a hack.

sp193:
I can't really remember much about the XFROM module. I think it was based on libmc, which was why it inherited the use of the memory card file mode bits.
I opened two of them and saw that when it registers the xfrom: device, the "type" field is only set to 0x10 (IOP_DT_FS in our ioman.h).

The XFROM driver was made to support the new IOMAN (and its structures). Since our iomanX also identifies a "legacy" driver based on whether the type field has 0x10000000 set (which is not), the conversion doesn't work correctly.

After some thought, maybe it's for the best. Without the conversion code we have, it doesn't seem possible to use new drivers with the old IOMAN structures anyway.

Do you have any idea why he needed to replace macros like FIO_S_ISLNK with FIO_S_IFLNK?

The formatting seems a little messed up. But since it's his commit, perhaps it can be cleaned up later.

AKuHAK:
> 
> Do you have any idea why he needed to replace macros like FIO_S_ISLNK with FIO_S_IFLNK?
> 

Oh I didn't realize that - I thought that it was the same, just changed logic. My bad :) In that case, I have no idea why these macros are changed. And these changes doesnt look safe, maybe @balika011 can comment on this?

Regarding iomanx itself and @BoGanon works I recently dig out his forks (I plan to prepare commit for Doxygen support later) and realized that he changed a lot of things inside ioman module.
For example in [this](https://github.com/BoGanon/ps2sdk/commit/a0a7def3367d6f7d994f81b83e6334420c6aee90#diff-1a39dcf4f60a2bf324b878e593243c07) commit @BoGanon renamed IOP_DT_FSEXT to FILEXIO_DT_FSEXT ("rename confusing flag meant for fileXio").
He also made many changes inside iomanX.c which difficult to track cause he changed the formatting in the same commit.

I think the key thing is this [commit](https://github.com/BoGanon/ps2sdk/commit/134292444f846dda3148d441aad388d538409d64#diff-1a39dcf4f60a2bf324b878e593243c07):
>(IO) Fix 14-15 year old IOMANX misconception.
>
>Now IOMANX and ps2netfs won't do implicit file mode conversions.
>Other mode conversions were also removed.
>
>The two types of flags are for simple filesystems without access modes
>or permissions and for filesystems with access modes or permissions.
>
>See sys/io_stat.h for a list of devices that use simple flags.

In this commit @BoGanon completely removed this confusing conversion so all changes mad by @balika011 become useless. I am not completely sure that after this commit wLaunchELF will work normally.

sp193:
Even though I think the mode bits shouldn't have been converted in the first place, I agree that it's too late. Yes, if we undo the mode bits conversion, then existing software that expect the old uniform behaviour of the mode bits will stop working correctly.

That IOP_DT_FSEXT flag exists in the Sony modules too, from what I remember. Although it's not used for any mode bit conversions, it is used to determine whether the functions after chstat() are supported. I don't think it's supposed to be a homebrew thing, but we just don't know its official name.

I think @BoGanon may have broke something else in that commit, unless he planned to change the way iomanX was meant to be used. There is a need to do some conversion around the stat structures, which is what iomanX does in mode2modex (other than just changing the mode bits). Otherwise, things may get ugly when people mix and match modules that were compiled for different SDK versions; io_dirent_t represents the structure used by the older modules, which are made with the design of the early Sony SDKs. The newer SDKs use a larger structure, which is similar to iox_dirent_t.
This is another problem with our homebrew module - instead of being a full replacement for FILEIO & IOMAN and clearly defining the semantics for the I/O library functions, fileXio and iomanX are meant to be used alongside the boot ROM's FILEIO and IOMAN modules. As such, iomanX was designed to be compatible with even those old module semantics. "Newer" modules would be those cloned from the newer SDKs (somewhere around 2.1.1), such as the HDD modules (from 2.2.2?). The XFROMMAN module was built with a newer SDK, but did not support the extended IOMAN functions after chstat(), which is why it doesn't work with iomanX - since iomanX fails to convert its output for dread() and getstat() correctly.

Perhaps we could choose to either:
- clean up @balika011's commit and merge it. Since it's a hack that makes more sense for the way we've been always doing homebrew, I doubt it will go away so easily. The only part that still sticks out like a red flag to me, is the check for the "xfrom" device name. If there is a way to support modules like XFROM without explicitly making exceptions for specific device names, I am sure it will be a better design.
- Do nothing. We could choose to not use the Sony module with our iomanX module. Either a new, homebrew clone could be made or the Sony IOMAN should be used. I used the Sony IOMAN module with XFROMMAN, so I can say it is a workable option to get things done.

AKuHAK:
I hope we should close this pull request as some changes are not clear.

israpps:
Did this discussion continued elsewere?

Currently (If I understood correctly) PSX builtin XFROMMAN driver freezes before the `SifLoadModule()` ends it's job because of something related to this

PR #106, opened 2020-02-22
Various cleanup 8

uyjulian:


sp193:
What is the commit "Handle GP register for new Compiler" supposed to fix? That patch will prevent those macros from working entirely.

I remember that GCC was changed to prohibit the modification of $gp by default. If I am not mistaken, the logic behind that restriction was that $gp should be set by the kernel and not software, but that is not how things work here. I never found a way to inhibit that restriction.
If you were trying to shape the SDK to fit with that restriction, I am not sure if that is surely the right thing to do.

Technically, this patch could prevent such a compilation error because this would prevent the generation of code that modifies $gp. However, it also means that one cannot use $gp and projects must never use it either.

In the case of IOP modules like USBD, support for $gp preservation is lost too. You will inhibit the compilation error, but this functionality no longer works and nobody would know about it.

Is there no way to support the PS2 way of using $gp, whereby the software is free to change it?

uyjulian:
I dropped the relevant commit.

common/include/limits.h
```diff
@@ -72,11 +72,11 @@
 /* Minimum and maximum values a `signed long int' can hold.
    (Same as `int').  */
 #ifndef __LONG_MAX__
-#ifdef __R5900
+#if defined(__R5900) && (__GNUC__ <= 3)
```
rickgaiser: Are you aware of a similar problem in newlib? See this commit:
https://gitlab.com/ps2max/toolchain/newlib/-/commit/3c4ebfcea1fb06f0d5600b134198506ee2c274af

ee/libvux/src/vuhw.c
```diff
@@ -18,13 +18,23 @@ void Vu0IdMatrix(VU_MATRIX *m)
...
 	asm __volatile__(
+#if __GNUC__ > 3
+	"vmr32.xyzw  $vf18, $vf0    \n"
+	"sqc2        $vf0, 0x30(%0)   \n"
+	"vmr32.xyzw  $vf17, $vf18   \n"
+	"sqc2        $vf18, 0x20(%0)    \n"
+	"vmr32.xyzw  $vf16, $vf17   \n"
+	"sqc2        $vf17, 0x10(%0)    \n"
+	"sqc2        $vf16, 0x00(%0)    \n"
+#else
```
rickgaiser: tabs/spaces/aligning does not look right.
The same for all other assembly code.

ee/startup/src/crt0.s
```diff
@@ -174,3 +174,9 @@ _root:
    .align   6
 _args:
    .space   4+16*4+256   # argc, 16 arguments, 256 bytes payload
+
+   .sdata
+   .global __dso_handle
+   .weak __dso_handle
+__dso_handle:
+   .long 0
```
rickgaiser: crt0.o should actually be defined in newlib itself.
Is this for c++ destructors?

tools/bin2s/src/bin2s.c
```diff
@@ -64,6 +68,8 @@ int main(int argc, char *argv[])
 
 	fprintf(dest, ".balign 16\n\n");
 	fprintf(dest, ".globl %s\n",argv[3]);
+	fprintf(dest, ".type %s,@object\n",argv[3]);
+	fprintf(dest, ".size %s,%d\n",argv[3],fd_size);
```
rickgaiser: Compatible with old toolchain as well?

uyjulian: Yes, old toolchain also generates it.

ee/startup/src/crt0.s
uyjulian: Probably. I was getting linking error, so I decided to search newlib code for the symbol. It was defined that way so I copied it.

common/include/limits.h
uyjulian: My patch of newlib doesn't change `__LONG_MAX__`.

PR #108, opened 2020-03-14
Move libc to newlib

rickgaiser:
These PR series move libc from ps2sdk to newlib:

**1**: File and directory functions (both fio* and fileXio*) are now handled by newlib, tranparently. With the following functions supported:
-- close/open
-- read/write
-- lseek
-- ioctl
-- unlink (remove)
-- link (rename)
-- mkdir/rmdir
-- stat
-- opendir/closedir
-- readdir
-- rewinddir

Existing code must be modified to use these standard posix function names instead of calling fio* fileXio functions directly, becouse of an incompatibility between newlib and the ps2 file handing. For this reason existing code will provide an error when including ps2 headers that should not be used:

> #error "Using fio/fileXio functions directly in the newlib port will lead to problems."
> #error "Use posix function calls instead."

For some applications however, using the fio*/fileXio* functions directly may still be needed (for instance when mounting partitions). By defining 'NEWLIB_PORT_AWARE' before including these headers this is still possible.
As far as I know the only incompatibility it with opening a file. The flags, like O_RDONLY, are incompatible and must be changed to **FIO**_O_RDONLY.

**2**: time function is implemented, returning the UTC time. Using the stat function the times from files can also be returned in UTC format.

**3**: nanosleep is implemented. It sleeps using SetAlarm. Then busy waits the last nanoseconds, resulting in both a lot of free cpu time, and a very accurate sleep time.

**4**: The linkfile has been moved to the binutils port, and crt0.s has been moved to newlib port.

**5**: The newlib port is upgraded from 1.10 to 1.14. This adds  (amongs other things) wchar support, needed by some applications like uLE. Upgrading to newer versions resulted in a lot of compiler issues. Perhaps newlib can be upgraded to 3.x when the compiler gets updated.

The end result is that porting new applications to the ps2 will become a lot more easy. Or sometimes they will just run.

Special thanks to @fjtrujy, the creator of the PS2 RetroArch port, for making this possible.

---

This PR series consists of ps2toolchain, ps2sdk, ps2sdk-ports and gsKit to get a working newlib toolchain. This is the ps2sdk PR.
The commits are separated by topic as much as possible to make reviewing and discussing them easy. Most of them are small, but there's 1 big commit: "Move ee/libc to newlib".


fjtrujy:
Amazing, finally the duplication of the libc libraries is going to end.
Thanks @rickgaiser

PR #112, opened 2020-04-25
Include the access function in the ps2sdkapi

fjtrujy:
## Description

This PR is just for including the missing `access` function in the `ps2sdkapi.c`

Additional info:
https://linux.die.net/man/2/access

Thanks


rickgaiser:
Thanks. This looks good, but I think we need to update travis-ci.

AKuHAK:
@rickgaiser regarding CI you can also update DockerCI, if you have enough rights for it of course

PR #115, opened 2020-05-08
small update to adpenc

KrahJohlito:
This is just to write the fileSize of the input file to the output header... useful to determine duration of the compressed adp later on.

### DIFF
--- a/tools/adpenc/src/main.c
+++ b/tools/adpenc/src/main.c
@@ -4,7 +4,7 @@
 #include <ctype.h>
 #include <math.h>
 
-#define VERSION	"1.1"
+#define VERSION	"1.2"
 
 extern int adpcm_encode(FILE* fp, FILE* sad, int offset, int sample_len, int flag_loop);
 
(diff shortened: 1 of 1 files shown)


rickgaiser:
The play duration of an adpcm file is what we're interested in, not the filesize of the original file that was used to generate the adpcm file.

I would prefer this value to be in units of samples. It's the most accurate and usefull unit. Becouse we also record the sample frequency so we can calculate the time in milliseconds if we want to.

A 32bit samples field can hold a maximum of
(2^32) / 48000 / 3600 = 24.8 hours

I think you can use the 'sample_len' field that's already calculated as:
```
		if(fread( &sample_len, 4, 1, fp )==1){
			sample_len /= (channels*2);
		}
```

What do you think?

KrahJohlito:
What calc would we use to determine duration? nSamples / sample rate * channels? I’d like to test before committing but not sure of the formula

KrahJohlito:
adp-test
Left: OPL calculating duration in milliseconds
Right: Actual duration

rickgaiser:
Looks great.

Nice clean 3 line change, squash merging. Thanks.

PR #116, opened 2020-05-11
Update for BDM drivers

rickgaiser:
I've had these improvements laying around for some time now:
- Add MINI_DRIVER: add's a 2nd build option for the block devices. This removes all debugging and makes them small enough to be used as OPL in-game drivers.
- IEEE1394: small improvements to compatibility, but compatibility is still poor.


fjtrujy:
Thanks

PR #117, opened 2020-05-11
Newlib updates and gcc 10 compatibility

rickgaiser:
2 fixes for compiling ps2sdk on gcc 10.

Also libmf removed completely. Just like libc, it's part of newlib and should have been removed with the newlib update. Application note: inlude `-lm`, not `-lmf`

12 additions and **18,144** deletions


fjtrujy:
Nice! More generic a smaller code in our side

Velocet:
Amazing times! Thank you so much! And thanks to everyone working on PS2SDK!

PR #122, opened 2020-05-18
Artifacts and fixes to action

AKuHAK:
Removed -lmf
Added ps2sdk sample compilation stage
ps2sdk samples now available as artifacts

`github.repository_owner` replaced to `ps2dev` cause this environment doesn't work in forked projects, but this Testing Workflow can be useful for forks as well


fjtrujy:
Thanks @AKuHAK !!!

rickgaiser:
Hi @AKuHAK thank you for these additions. They look good!

PR #123, opened 2020-05-18
Put shorter SHA for example binaries

fjtrujy:


AKuHAK:
So difficult way to proceed shorter sha

fjtrujy:
> So difficult way to proceed shorter sha

The only way so far :(

rickgaiser:
The short SHA's are great for artifacts/downloads. It's indeed a little complicated to get them. If there's a simpler solution in the future we can change it then.

PR #124, opened 2020-05-29
Fixed job server

astrelsky:
I also had to run dos2unix on build-exports.sh but no matter what I did git always ignored the line endings.

Note that `make install` must still be run without jobs. Some of the prerequisite rules clean up without running `make clean` and I couldn't figure out the correct order to run them to make it work with a job server.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,9 +28,9 @@ subdir_release = $(patsubst %,release-%,$(SUBDIRS))
 subdirs: dummy $(subdir_list)
 
 $(subdir_list): dummy
-	$(MAKEREC) $(patsubst all-%,%,$@)
+	+$(MAKEREC) $(patsubst all-%,%,$@)
 $(subdir_clean): dummy
-	$(MAKEREC) $(patsubst clean-%,%,$@) clean
+	+$(MAKEREC) $(patsubst clean-%,%,$@) clean
 $(subdir_release): dummy
 	$(MAKEREC) $(patsubst release-%,%,$@) release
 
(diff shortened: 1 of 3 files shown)


fjtrujy:
Hello thanks to making the PR,
But I don't really understand where and how are you suffering these issues?
On the other hand, I don't understand if the `+` that you're adding could have additional issues.

As you can see we have CI/CD integrated into the repo, and over there we're compiling and generating the samples without any issue.

Could you create a separate branch trying to replicate your issues in the CI/CD? What OS are you using?

Thanks again

astrelsky:
> Hello thanks to making the PR,
> But I don't really understand where and how are you suffering these issues?
> On the other hand, I don't understand if the `+` that you're adding could have additional issues.
> 
> As you can see we have CI/CD integrated into the repo, and over there we're compiling and generating the samples without any issue.
> 
> Could you create a separate branch trying to replicate your issues in the CI/CD? What OS are you using?
> 
> Thanks again

I can make a copy of the output later today. This effectively removes the ‘warning: jobserver unavailable: using -j1. Add `+' to parent make rule.’ warning. You can read more information about it [here](https://www.gnu.org/software/make/manual/html_node/Error-Messages.html) but basically without the `+` they are effectively running with `-j1` instead of the specified number of jobs.

AKuHAK:
Did you check cross-dependencies before? Does each job workflow independent from another? @fjtrujy this PR for making multi-core compilation efficient, some kind of new feature, not error fixing.

fjtrujy:
Ok, so is already working with `-jX`???
@AKuHAK  If this is true, can we modify the CI/CD before merge it to assure that is working using multi-cores?

Thanks

astrelsky:
I have a change coming to fix running `make -jX install` without running building first.

astrelsky:
That should do it. There is a dependency on liberl which is why it was moved to be first. It's not the best way to handle it though.

PR #125, opened 2020-06-05
Add CDFS driver

fjtrujy:
## Description
This PR is to include the `CDFS` driver in the main `PS2SDK`

I have tested this driver in `RetroArch` and in a dummy project and in both worked like a charm.
The driver it-self has been created taking as a reference to the implementation that `uLaunchELF` did in the past. Now it is adjusted to the `newlib` era, making the use of it easier than ever.

Most probably the driver is not perfect and will have some issues and/or missing functionality, but so far is enough like to be included in the `PS2SDK` because it won't have any conflict with the existing code.

Here you have a dummy example about how to use it:
```
extern unsigned char cdfs_irx;
extern unsigned int size_cdfs_irx;

int main()
{      
	int fd;
	DIR *dirp;
	struct dirent *direntp;
	struct stat buf;

	while(!SifIopReset(NULL, 0)){};
	while(!SifIopSync()){};
	SifInitRpc(0);
	sbv_patch_enable_lmb();

	/* CDFS */
	int loaded = SifExecModuleBuffer(&cdfs_irx, size_cdfs_irx, 0, NULL, NULL);
	printf("IRX has been loaded %i\n", loaded);

	// USE FROM HERE NORMAL I/O FUNCTION AS:

	printf("*****\n\nTrying mkdir\n *****\n\n");
	mkdir("cdfs:test", 0755);

	printf("*****\n\nTrying open\n *****\n\n");
	fd = open("cdfs:README.TXT", O_RDONLY);

	printf("*****\n\nTrying close\n *****\n\n");
	close(fd);

	printf("*****\n\nTrying open\n *****\n\n");
	fd = open("cdfs:BADAPPLE4.ZIP", O_RDONLY);

	printf("*****\n\nTrying close\n *****\n\n");
	close(fd);

	printf("*****\n\nTrying open\n *****\n\n");
	fd = open("cdfs:PS2IDENT", O_RDONLY);

	printf("*****\n\nTrying close\n *****\n\n");
	close(fd);

	printf("*****\n\nTrying opendir\n *****\n\n");
	dirp = opendir("cdfs:");
	if (dirp == NULL) {
		printf("Error dir can not be opened\n");
	}

	while ((direntp = readdir(dirp)) != NULL) {
		printf("Reading dir!\n");
	}

	printf("*****\n\nTrying closedir\n *****\n\n");
	closedir(dirp);

	printf("*****\n\nTrying getStat\n *****\n\n");
	stat("cdfs:/GB", &buf);

	printf("Is dir %i\n", S_ISDIR(buf.st_mode));

	while (1) {}

	return 0;
}
```

The makefile is like any project that uses `IRX` files.

Thanks!


rickgaiser:
Great work @fjtrujy !

So previously we needed to load CDVDMAN and CDVDFSV. Then we could use the CD/DVD from the EE with a special API, like sceCdInit / sceCdReadDVDV / sceCdSync.

With this new driver we can load CDVDMAN and the new CDFS. Then use normal libc open/close/read/write. It's strange that sony didn't make it like this in the first place (with CDVDFSV).

I'll try to do a code review and a test if I get some time. I've got 1 comment so far:
All CD/DVD related IRX's seem to be prefixed with "CDVD", indicating it's for both CD and DVD. You've also named the folder "cdvd", but the library is "CDFS". Shouldn't this be "CDVDFS"?

sp193:
You can access files via the FILEIO functions if CDVDMAN. It was directory functions that it lacked. If the default version of CDVDMAN does not support the FILEIO functions, it was added later on then.

fjtrujy:
Hello,
Yes @sp193 is right, I forgot to mention in the PR, this is also adding new file system support with the directory functions available.

@rickgaiser I'm so bag giving names, I decided to put `cdfs_` in the functions because the driver looking for `cdfs` path, but I'm totally open to rename it.

Thanks.

rickgaiser:
@sp193 I've searched for examples of this but couldn't find any. There's some references to using fileio for CDVD in the official SDK, but I could not get anything working (using rom0:CDVDMAN + rom0:CDVDFSV). Do you know of a (preferably working) example somewhere?

sp193:
Unless you manipulated the boot process, CDVDMAN+CDVDFSV are loaded with every boot and you should not be loading them manually.
You need to access files via cdrom0. CDVDMAN only supports ISO9660 level 1, so all filenames must be in 8+3 format and in uppercase. Path segments must be separated with the backslash, unlike most other device drivers. Filenames must have the version 1 suffix (;1).

For example, to load SYSTEM.CNF, you could do something like:
``int fd = open("cdrom0:\\SYSTEM.CNF;1", O_RDONLY);``

I'm pretty sure I used to FILEIO methods in my [OSD init example](https://sites.google.com/view/ysai187/home/projects/initializing-the-ps2psx?authuser=0#h.p_9CPpcpE6hwxj). We've been using this method in FMCB. In fact, I don't really recall any project using the libcdvd APIs..

AKuHAK:
There was also licensing problems. Hiryu, who developed libcdvd library with extended iso support does not allow to include his libraries into mainstream ps2sdk. This is why all developers grab his library into his source. For example my fceumm-ps2 project has his library. You can find his readme here https://github.com/AKuHAK/Fceumm-PS2/blob/master/libcdvd/docs/libcdvd_ref.pdf

The same story about his libhdd library, the only difference, he allowed to use his library, but does not allow to use pdf manual.
I cant find any reference for this story in 2020, but I remember that this was true

fjtrujy:
Hello @AKuHAK ,
I think that you're partially wrong.

You're right I forgot to include the `lincese.txt`, (I already did it) but If you read carefully is the only requirement to distribute and/or modify this library
Is to include the license itself.

https://github.com/AKuHAK/Fceumm-PS2/blob/master/libcdvd/license.txt

As you can see here.

Thanks

iop/cdvd/cdfs/src/cdfs_iop.c
```diff
@@ -0,0 +1,1263 @@
...
+        return TRUE;
+
+    memset(buf, 0, (sectors << 11));
+
+    return FALSE;  // error
+}
+
+// This is the getdir for use by the EE RPC client
+// It DMA's entries to the specified buffer in EE memory
+int cdfs_getDir(const char *pathname, const char *extensions, enum CDFS_getMode getMode, struct TocEntry tocEntry[], unsigned int req_entries) {
```
rickgaiser, line 985: It looks like this function transfers to the EE memory, but it's used from IOP. Is this correct?

iop/cdvd/cdfs/src/cdfs_iop.c
```diff
@@ -0,0 +1,1263 @@
...
+                    dmaStruct.src = &localTocEntry;
+                    dmaStruct.dest = &tocEntry[matched_entries];
+                    tocEntry[matched_entries] = localTocEntry;
+                    dmaStruct.size = sizeof(struct TocEntry);
+                    dmaStruct.attr = 0;
+
+                    // Do the DMA transfer
+                    CpuSuspendIntr(&intStatus);
+
+                    dmaID = sceSifSetDma(&dmaStruct, 1);
```
rickgaiser, line 1093: DMA to EE?

iop/cdvd/cdfs/src/cdfs_iop.c
```diff
@@ -0,0 +1,1263 @@
...
+#endif
+                            strcpy(localTocEntry.filename, "..");
+                        }
+                    }
+
+                    // DMA localTocEntry to the address specified by tocEntry[matched_entries]
+
+                    // setup the dma struct
+                    dmaStruct.src = &localTocEntry;
+                    dmaStruct.dest = &tocEntry[matched_entries];
```
rickgaiser, line 1085: This points to a struct on the IOP, right?

iop/cdvd/cdfs/src/cdfs_iop.c
```diff
@@ -0,0 +1,1263 @@
...
+                    dmaID = sceSifSetDma(&dmaStruct, 1);
+
+                    CpuResumeIntr(intStatus);
+
+                    matched_entries++;
+                } else  // it must be a file
+                {
+#ifdef DEBUG
+                    printf("We found a file, but we dont want files (at least not yet)\n\n");
+#endif
```
rickgaiser, line 1102: These 3 lines are in many places. The code would be more readable when using for instance DPRINTF, defined once at the top of the file.

iop/cdvd/cdfs/src/imports.lst
```diff
@@ -0,0 +1,64 @@
...
+I_sceCdDiskReady
+I_sceCdGetDiskType
+I_sceCdTrayReq
+I_sceCdSearchFile
+cdvdman_IMPORTS_end
+
+intrman_IMPORTS_start
+I_CpuSuspendIntr
+I_CpuResumeIntr
+intrman_IMPORTS_end
```
rickgaiser, line 16: I don't think these would be nessecary

iop/cdvd/cdfs/src/imports.lst
```diff
@@ -0,0 +1,64 @@
...
+I_sceSifInitRpc
+I_sceSifSetRpcQueue
+I_sceSifRegisterRpc
+I_sceSifRpcLoop
+sifcmd_IMPORTS_end
+
+sifman_IMPORTS_start
+I_sceSifSetDma
+I_sceSifDmaStat
+sifman_IMPORTS_end
```
rickgaiser, line 47: Same here, all related to communication to the EE ?

iop/cdvd/cdfs/src/main.c
```diff
@@ -0,0 +1,469 @@
...
+#include <iox_stat.h>
+#include <sysclib.h>
+
+#include "cdfs_iop.h"
+
+// 16 sectors worth of toc entry
+#define MAX_FILES_PER_FOLDER 256
+#define MAX_FILES_OPENED 16
+#define MAX_FOLDERS_OPENED 16
+#define MAX_BITS_READ 16384
```
rickgaiser, line 12: Bytes probably?

iop/cdvd/cdfs/src/main.c
```diff
@@ -0,0 +1,469 @@
...
+
+/***********************************************
+*                                              *
+*             DRIVER FUNCTIONS                 *
+*                                              *
+***********************************************/
+
+static int fio_init(iop_device_t *driver)
+{
+    printf("CDFS Filesystem v2\n");
```
rickgaiser, line 52: This is the driver description. It would be nice if it matches the description passed to AddDrv below. Perhaps put the description in a define at the top, just like the #define UNIT_NAME.

iop/cdvd/cdfs/src/main.c
```diff
@@ -0,0 +1,469 @@
...
+    static iop_device_t fio_driver;
+
+    // Prepare cache and read mode
+    cdfs_prepare();
+
+    // setup the fio_driver structure
+    fio_driver.name = UNIT_NAME;
+    fio_driver.type = IOP_DT_FS;
+    fio_driver.version = 2;
+    fio_driver.desc = "CDFS Filedriver";
```
rickgaiser, line 462: Different description?

iop/cdvd/cdfs/src/cdfs_iop.c
fjtrujy: Yeah, you were right, it was using EE memory. Now I have updated the PR and remove this logic

iop/cdvd/cdfs/src/cdfs_iop.c
fjtrujy: Yeah, you were right, it was using EE memory. Now I have updated the PR and remove this logic

iop/cdvd/cdfs/src/cdfs_iop.c
fjtrujy: Done!

iop/cdvd/cdfs/src/imports.lst
fjtrujy: I have reviewed all the imports and I just have now the used ones

iop/cdvd/cdfs/src/imports.lst
fjtrujy: Removed!

iop/cdvd/cdfs/src/main.c
fjtrujy: Done!

iop/cdvd/cdfs/src/main.c
fjtrujy: Done!

fjtrujy:
Merging after your approval.
Thanks @rickgaiser for your review.

iop/cdvd/cdfs/src/cdfs_iop.c
```diff
@@ -0,0 +1,1263 @@
...
+static sceCdRMode cdReadMode;
+
+/***********************************************
+*                                              *
+*            PRIVATE FUNCTIONS                 *
+*                                              *
+***********************************************/
+
+// Used in findfile
+static int strcasecmp(const char *s1, const char *s2) {
```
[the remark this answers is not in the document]
rickgaiser: On the EE we do, but on the IOP we do not have newlib.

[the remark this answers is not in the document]
AKuHAK: Oh, thank you for information.

PR #126, opened 2020-06-09
Patch 3

AKuHAK:


ee/input/src/input.c
```diff
@@ -91,7 +91,7 @@ void pad_close(pad_t *pad)
 		free(pad->actuator);
 	}
 
-	padPortClose(pad->port,pad->slot);
+	while (padPortClose(pad->port,pad->slot) < 0);
```
rickgaiser, line 94: Why are you waiting forever for this function to return ">= 0":
https://github.com/ps2dev/ps2sdk/blob/78ba532388da5de97f8cb681814a12e0586a735d/ee/rpc/pad/src/libpad.c#L448-L467
I'm not sure but I think this can cause an application to freeze under some circumstances. Shouldn't instead SifCallRpc be called in a loop with a timeout?

AKuHAK: probably you are right, I will close this PR

PR #129, opened 2020-06-20
Create a library to make easier the Loading of ELF files

fjtrujy:
## Description

This PR is for making it easier to load ELFs from other `ELFs`. 
The main concept is inspired by the logic that `uLE` is following for launching `ELFs`, but obviously is improved.

Now, whatever ELF app could load another ELF.
The only steps that need to do is:
Add in the make file the `-lelf-loader` to the linked libraries.
Call `sbv_patch_disable_prefix_check()`
Call `LoadELFFromFile(newELFPath, argc, argv)`

I did test through a dummy test app and in `RetroArch` as well.

Thanks.

### DIFF
--- a/ee/elf-loader/include/elf-loader.h
+++ b/ee/elf-loader/include/elf-loader.h
@@ -22,7 +22,7 @@
 extern "C" {
 #endif
 
-int LoadELFFromFile(const char *filename);
+int LoadELFFromFile(const char *filename, int argc, char *argv[]);
 
 #ifdef __cplusplus
 }
--- a/ee/elf-loader/include/elf-loader.h
+++ b/ee/elf-loader/include/elf-loader.h
@@ -22,6 +22,7 @@
 extern "C" {
 #endif
 
+// Before call this method be sure that you have previously called sbv_patch_disable_prefix_check();
 int LoadELFFromFile(const char *filename, int argc, char *argv[]);
 
 #ifdef __cplusplus
--- a/ee/elf-loader/include/elf-loader.h
+++ b/ee/elf-loader/include/elf-loader.h
@@ -1,0 +1,31 @@
+/*
+# _____     ___ ____     ___ ____
+#  ____|   |    ____|   |        | |____|
+# |     ___|   |____ ___|    ____| |    \    PS2DEV Open Source Project.
+#-----------------------------------------------------------------------
+# (c) 20020 Francisco Javier Trujillo Mata <fjtrujy@gmail.com>
+# Licenced under Academic Free License version 2.0
+# Review ps2sdk README & LICENSE files for further details.
+*/
+
+/**
+ * @file
+ * ELF Loader functions.
+ */
+
+#ifndef __ELFLOADER_H__
+#define __ELFLOADER_H__
+
+#include <tamtypes.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int LoadELFFromFile(const char *filename);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __ELFLOADER_H__ */
--- a/ee/elf-loader/include/elf-loader.h
+++ b/ee/elf-loader/include/elf-loader.h
@@ -3,7 +3,7 @@
 #  ____|   |    ____|   |        | |____|
 # |     ___|   |____ ___|    ____| |    \    PS2DEV Open Source Project.
 #-----------------------------------------------------------------------
-# (c) 20020 Francisco Javier Trujillo Mata <fjtrujy@gmail.com>
+# (c) 2020 Francisco Javier Trujillo Mata <fjtrujy@gmail.com>
 # Licenced under Academic Free License version 2.0
 # Review ps2sdk README & LICENSE files for further details.
 */
@@ -15,8 +15,6 @@
 
 #ifndef __ELFLOADER_H__
 #define __ELFLOADER_H__
-
-#include <tamtypes.h>
 
 #ifdef __cplusplus
 extern "C" {
--- a/ee/elf-loader/src/elf.c
+++ b/ee/elf-loader/src/elf.c
@@ -14,8 +14,7 @@
 
 #include "elf.h"
 
-#define MAX_PATH 1025
-
+// Loader ELF variables
 extern u8 loader_elf[];
 extern int size_loader_elf;
 
@@ -23,16 +22,7 @@ extern int size_loader_elf;
 #define ELF_MAGIC 0x464c457f
 #define ELF_PT_LOAD 1
 
-
-//------------------------------
-//End of func:  int checkELFheader(const char *path)
-//--------------------------------------------------------------
-// RunLoaderElf loads LOADER.ELF from program memory and passes
-// args of selected ELF and partition to it
-// Modified version of loader from Independence
-//	(C) 2003 Marcus R. Brown <mrbrown@0xd6.org>
-//------------------------------
-void RunLoaderElf(char *filename, char *party)
+void LoadELFFromFile(char *filename)
 {
 	u8 *boot_elf;
 	elf_header_t *eh;
--- a/ee/elf-loader/src/elf.c
+++ b/ee/elf-loader/src/elf.c
@@ -11,6 +11,9 @@
 #include <string.h>
 #include <sifrpc.h>
 #include <kernel.h>
+#include <sys/stat.h>
+#include <stdbool.h>
+#include <malloc.h>
 
 #include "elf.h"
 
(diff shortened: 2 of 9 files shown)


ee/elf-loader/src/elf.c
```diff
@@ -0,0 +1,83 @@
...
+	elf_header_t *eh;
+	elf_pheader_t *eph;
+	void *pdata;
+	int i;
+	int new_argc = argc + 1;
+	char **new_argv = malloc(new_argc);
+	
+	// We need to check that the ELF file before continue
+	if (!file_exists(filename)) {
+		return -1; // ELF file doesn't exists
```
Darkshadow2, line 47: If the file doesn't exist, you're leaking the memory just allocated for `new_argv`

sp193:
For knowledge, since you seem to be using some parameters from LaunchELF's old ELF loader: the official Sony ELF loaders load at either 0x00082000 or 0x00084000, depending on the SDK.

At 0x00080000 lives the idle thread module (EENULL). It also contains the original userland dispatcher for the kernel's alarm functions.
The alarm function patch, which contains its own replacement for the userland dispatcher, loads this replacement to 0x00082000. Software compiled with such a SDK (i.e. HDD Browser) will have an ELF loader that begins from 0x00084000 instead of the classic 0x00082000.

Our SDK now has this patch, so how low this can actually go, depends on whether the patch is included or not.

The memory wipe begins from the end of the ELF loader, until the end of memory. The official loaders have the location of the stack defined before the end of the program, allowing it to be excluded from the  memory wipe. The sizes of the heap and stack are also predefined.

ee/elf-loader/src/elf.c
fjtrujy: Thanks!

fjtrujy:
Thanks @sp193, for sharing the info.
Sorry if I did something wrong, I just wanted to make it easier and generic the loading of ELF from any single PS2 App, and that is why I took as reference what currently uLE does.

But with your previous response, I'm not pretty sure about what you recommend me to do.
Could you kindly explain it for dummies (as me)?

Thanks again.

ee/elf-loader/src/loader/Rules.make
```diff
@@ -0,0 +1,58 @@
...
+	$(EE_CXX) $(EE_CXXFLAGS) $(EE_INCS) -c $< -o $@
+
+%.o: %.S
+	$(EE_CC) $(EE_CFLAGS) $(EE_INCS) -c $< -o $@
+
+%.o: %.s
+	$(EE_AS) $(EE_ASFLAGS) $< -o $@
+
+$(EE_BIN): $(EE_OBJS)
+	$(EE_CC) $(EE_CFLAGS) -o $(EE_BIN) $(EE_OBJS) $(EE_LDFLAGS) $(EE_LIBS)
```
rickgaiser, line 51: The loader should be as small and simple as possible. I think you're including newlib/libc. How big is the loader? And can it be smaller when not using libc (adding -nostdlib)?

sp193:
It's not a strict must-have since most homebrew software will boot software that run within the normal memory ranges (evident by how LaunchELF has worked for years). I preferred to follow whatever Sony did, as it tends to cover everything that could be done by their licensees.
However, using LaunchELF's method does leave gaps:
- Memory between the end of the EE ELF loader and 0x00100000 is uninitialized.
- Using 0x000A0000 would prevent software that loads there, from being loaded.

By the way: the ELF loader core itself could be possibly built with libkernel-nopatch. Then you won't have all patches included into the ELF loader itself. Particularly the unrelated patches like TLB, alarm patch and thread patch.
If ExecPS2() requires a patch, the parent program (which loads the ELF loader) could contain it (by linking with the regular libkernel).

Another thing Sony did, was to store the ELF loader as a static binary file. It resulted in more savings, from what I do remember - since you have no need for code to support the ELF format and there's no overhead from the other parts of the ELF too.

ee/elf-loader/src/elf.c
```diff
@@ -0,0 +1,85 @@
...
+		if (eph[i].type != ELF_PT_LOAD)
+			continue;
+
+		pdata = (void *)(boot_elf + eph[i].offset);
+		memcpy(eph[i].vaddr, pdata, eph[i].filesz);
+
+		if (eph[i].memsz > eph[i].filesz)
+			memset(eph[i].vaddr + eph[i].filesz, 0,
+			       eph[i].memsz - eph[i].filesz);
+	}
```
rickgaiser, line 77: ELF loading requires loading 2 ELF's:
- First: APP1 loads the LOADER elf, at a special memory addres that does not conflict with loaded apps
- Second LOADER loads the APP2 elf

This selected code is custom code for loading the LOADER elf. Why is this code so very different from the code below, for loading the APP2 elf? Shouldn't they be the same, or even better, use the same library?

ee/elf-loader/src/loader/loader.c
```diff
@@ -0,0 +1,73 @@
...
+		return -EINVAL;
+	}
+
+	// Initialize
+	SifInitRpc(0);
+	wipeUserMem();
+
+	//Writeback data cache before loading ELF.
+	FlushCache(0);
+	ret = SifLoadElf(argv[0], &elfdata);
```
rickgaiser, line 55: This is the code for loading the APP2 elf. Using a function instead of custom code. Also note the special wipeUserMem fuction, shouldn't the sections already be wiped when loading the ELF, as the cusom loading code above does?

ee/elf-loader/src/elf.c
```diff
@@ -0,0 +1,85 @@
...
+	if (!file_exists(filename)) {
+		return -1; // ELF file doesn't exists
+	}
+	// ELF Exists
+	new_argv = malloc(new_argc);
+
+	new_argv[0] = (char *)filename;
+    for (i = 0; i < argc; i++) {
+		new_argv[i + 1] = argv[i];
+	}
```
rickgaiser, line 55: I'm sorry, but there's so many things wrong here:

1. Assuming "argc=1" and therefore "new_argc=2", you're allocating 2 bytes instead of 8 "malloc(new_argc*sizeof(char*));"
2. malloc also requires free, or you'll have a "memory leak". I can immagine in this specific case that it is intentional, but then you should comment why and how.
3. You're copying the POINTERS, not the DATA. This may seem to work fine in some situations, but will most likely fail in others. The pointers point to data from the OLD application. That data will be overwritten by the NEW application that is being loaded. To solve this you must copy this data to the loader's heap, assuming the heap is also below 0x100000, and not being overwritten by the new application.
4. And lastly, tabs or spaces, not both.

rickgaiser:
> For knowledge, since you seem to be using some parameters from LaunchELF's old ELF loader: the official Sony ELF loaders load at either 0x00082000 or 0x00084000, depending on the SDK.

What would be the best for ps2sdk in this case? Re-use the exising loader addresses (0x00082000 or 0x00084000), or not using any of the addresses used by sony? Like 0x90000 as uLE and this loader now does?

> The memory wipe begins from the end of the ELF loader, until the end of memory. The official loaders have the location of the stack defined before the end of the program, allowing it to be excluded from the memory wipe. The sizes of the heap and stack are also predefined.

This memory wipe is done by the ExecPS2 function? If so, then do we still need the wipeUserMem function in the loader?

> Another thing Sony did, was to store the ELF loader as a static binary file. It resulted in more savings, from what I do remember - since you have no need for code to support the ELF format and there's no overhead from the other parts of the ELF too.

Like for example these: https://github.com/ps2dev/ps2sdk/tree/master/ee/kernel/src/eenull ?
Would it be possible to also create a compatible loader at 0x00082000, 0x00084000 or both? Perhaps add it to libkernel? Just as eenull? How would we make compatible linkfile and functions?

sp193:
I think 0x00084000 is best, since we should not assume that the parent program does not use the alarm patch. The current address also works but its position is just not as low as it could possibly get.

Memory wipe is done by LoadExecPS2(), but not ExecPS2(). LoadExecPS2() wipes memory, before loading EELOAD (the ELF loader). ExecPS2() is used as part of this process

If you create a loader that could be loaded from any address, it would not really be of much benefit since most software will load to the standard memory ranges. Sony moved it once, due to the introduction of the alarm patch.
Maybe it would work better for OPL's usecases.

ee/elf-loader/include/elf-loader.h
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * @file
+ * ELF Loader functions.
+ */
+
+#ifndef __ELFLOADER_H__
+#define __ELFLOADER_H__
+
+#include <tamtypes.h>
```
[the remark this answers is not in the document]
fjtrujy: Done!

ee/elf-loader/src/loader/Rules.make
fjtrujy: Yes you're totally right, I think that the loader it-self could be done in less than 5kb, however, it will require to have custom `linkfile`, and I don't feel confident creating `custom linkfiles`.

I think this should be tried in a second phase.

ee/elf-loader/src/elf.c
fjtrujy: I think that initial reason why it was done in this way in the `uLE` was to reduce the size of the `loader`, instead of create a custom linkfile to reduce the executable size, they tried to remove the sections that are not needed, reducing as well the amount of used memory.

ee/elf-loader/src/loader/loader.c
fjtrujy: Not sure if I understood your question here...
I think that `SIfLoadElf` just wipe the sections where `ELF` is going to be loaded, but I suppose that is safer if the whole user-space memory is clean before executing the APP2

ee/elf-loader/src/elf.c
fjtrujy: Hi there!
1 and 2, I have changed back to static allocation.
3. Yes, you're right I'm copying just the pointers and not the content because I have seen how the `ExecPS2` function it-self creates new copies internally.
How I checked was printing the address memory of the `argvs parameters` in the `elf.c -> LoadELFFromFile` function and they were totally different than the one in the `loader.c -> main` function.
4. Thanks fixed!

ee/elf-loader/src/loader/loader.c
sp193: Sony also does this memory wipe. It's good for ensuring consistent behaviour. krHACKen observed that SifLoadElf doesn't zero the unloaded sections.
This memory wipe would also be the one to erase the memory used by the program's stack.

PR #130, opened 2020-06-23
CDFS: DVD video support

uyjulian:
Fixes support for DVD video discs.

### DIFF
--- a/iop/cdvd/cdfs/src/cdfs_iop.c
+++ b/iop/cdvd/cdfs/src/cdfs_iop.c
@@ -814,6 +814,7 @@ int cdfs_findfile(const char *fname, struct TocEntry *tocEntry) {
 
 int cdfs_readSect(u32 lsn, u32 sectors, u8 *buf) {
     u32 i;
+    u32 consecutive_sectors = (sectors > 2) ? (sectors * 2048) / 2064 : 0;
     int retry;
     int result = 0;
     cdReadMode.trycount = 32;
@@ -831,7 +832,24 @@ int cdfs_readSect(u32 lsn, u32 sectors, u8 *buf) {
 
         if (sceCdGetDiskType() == SCECdDVDV)
         {
-            for (i = 0; i < sectors; i += 1)
+            if (consecutive_sectors > 0)
+            {
+                result = !sceCdReadDVDV(lsn, consecutive_sectors, buf, &cdReadMode);
+                if (result == 0)
+                {
+                    sceCdSync(0);
+                    result = sceCdGetError();
+                }
+                if (result != 0)
+                {
+                    continue;
+                }
+            }
+            for (i = 0; i < consecutive_sectors; i += 1)
+            {
+                memmove(buf + (2048 * i), buf + (2064 * i) + 12, 2048);
+            }
+            for (i = consecutive_sectors; i < sectors; i += 1)
             {
                 result = !sceCdReadDVDV(lsn + i, 1, dvdvBuffer, &cdReadMode);
                 if (result == 0)
--- a/iop/cdvd/cdfs/src/cdfs_iop.c
+++ b/iop/cdvd/cdfs/src/cdfs_iop.c
@@ -132,6 +132,7 @@ enum PathMatch {
 static struct CacheInfoDir cacheInfoDir;
 static struct CDVolDesc cdVolDesc;
 static sceCdRMode cdReadMode;
+static u8 dvdvBuffer[2064];
 
 /***********************************************
 *                                              *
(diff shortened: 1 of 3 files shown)


fjtrujy:
Thanks @uyjulian for updating the cdfs support.
Why aren't you including the `AllowDVDV` as well in the PR?
I have been taking a look to the code of `AllowDVDV` the the `irx` generated is so small, maybe we should attach inside of the `cdfs` driver.

What do you think @rickgaiser ?
Thanks

rickgaiser:
Agreed, AllowDVDV is small and probably only usefull in combination with cdfs.
Likewise, this PR is only usefull in combination with AllowDVDV.

Is there a reason to keep it separated?

TnA-Plastic:
I suppose he left it out, because it is up to the program to "decide l" how to access DVD-based discs.

fjtrujy:
@TnA-Plastic you're right, let the developer choose, but offer the possibility in the SDK, right now with this PR we are doing changes in the SDK than are "not functional" at all unless you include `AllowDVDV` methods.

Thanks

TnA-Plastic:
Yeah, well it was a change/PR to the disc-driver and FS-access and AllowDVDV is an own function/module.

uyjulian:
I integrated AllowDVDV.

sp193:
AllowDVDV is used to unlock the DVD video playback function of the DVD drive. It's unlocked until something locks it. In the official flow, it probably gets unlocked when the DVD player is loaded - which is why no official software needed a hack like that.
FMCB will enable the DVD drive. So the only usecase for this - would be if you used a modchip or some disc (e.g. Swapmagic), to boot your software. (But if you're writing homebrew software and not using the DVD player with DVD video discs, why use discs?)

uyjulian:
An alternate entrypoint is used when FMCB is not used.

AKuHAK:
> I integrated AllowDVDV into the CDFS driver.

Is it possible to comment out what each AllowDVD hex pattern means? For now, it is difficult to understand how does it work. Such documentation isn't needed for projects like wLe but for ps2sdk it is better to keep things documented.

uyjulian:
> Is it possible to comment out what each AllowDVD hex pattern means?

Sure.

rickgaiser:
Unlike official sony software, homebrew does not know where it's loaded from and what modules are loaded. It's nice that FMCB unlocks the DVD, but it's probably also useless. An application that needs to use the DVD cannot assume it's being loaded from FMCB. So the application still needs to unlock the DVD to be sure it can be used.
The homebrew application should work not only from FMCB, but also from disc/fortuna/modchip/etc...

@sp193 you agree?

TnA-Plastic:
At least for wLE as THE filemanager and app/ELF-Launcher, it is a "must have" IMO.

It's great to have the feature itself (implemented in the SDK), in any way!

sp193:
It's up to you guys. Since the commit was already made and it isn't a change with detrimental effects.

As fjtrujy mentioned, the problem is with the purpose of this PR. I feel that it'll change what this module was meant for. I saw this module as originally meant for supporting new filesystems, not a means to support this hack. I thought that it would make more sense to use USB devices for homebrew, since they're easier to work with and are common. Unless one would like to build a DVD video player.
But in either case, it doesn't help if DVD video playback isn't enabled.

This commit seemed to be for a trick like ESR - the disc is a burned disc that appears to the MECHACON as a DVD Video Disc, but is not used for DVD Video playback.
Accesses to such a disc, must be done through sceCdReadDVDV instead of sceCdRead() - so even with AllowDVDV applied, one cannot use sceCdRead() on a burned disc. The AllowDVDV module unlocks DVD Video Disc reading - not reading of data DVD discs.

uyjulian:
The following is the main inspiration for this pull request: https://github.com/CTurt/FreeDVDBoot

TnA-Plastic:
The news will be on psx-place.com later on today.
I'm quite busy right now, but I prepared some portions already.

VersatileNinja:
It's already making the rounds today, saw it first on reddit. Can't wait to see how this evolves.

uyjulian:
Since FMCB and the recently released FreeDVDBoot allow DVD Video discs to be used, I removed the integrated AllowDVDV.

sp193:
Since that exploit is for the DVD player, I presume the DVD player must be first loaded. That process itself would also unlock DVD video playback.

rickgaiser:
So this PR without AllowDVDV is now good? It can be pulled?

uyjulian:
I've tested the DVD video disc functionality using uLaunchELF, so I can say that this PR works fine for me.  

In the future, when CDVDMAN is reimplemented, I plan to remove this, so this is a workaround for now.

PR #131, opened 2020-06-28
ps2_clock() needs an u64 datatype

rickgaiser:
Since the newlib port the ps2_clock function uses the clock_t datatype as defined in newlib. Currently newlib has it defined as 64bit, but in different versions this can also be defined as 32bit, causing ps2_clock to then fail.

This PR creates a new ps2_clock_t type that is always 64bit.


fjtrujy:
Nice!!!

PR #135, opened 2020-07-12
Use pwd in filename passed to `open()`.

### LINKED ISSUE #134: chdir() is broken
When the change was made for newlib to replace the existing libc implementation, it looks like the logic to use the current working directory set by `chdir()` has been lost. Previously if you called `fopen("abc", "r")` it would automatically set this to an absolute path based in wherever the current working directory is (which is the directory where the ELF was launched from by default, thanks to `_ps2sdk_args_parse()`). Now this function will try to open `abc` as if it were an absolute path, which isn't correct.

Observe the old `fopen()` function [here](https://github.com/ps2dev/ps2sdk/blob/ea69e7cf8e4cea516b18dbdee13f698b8f336257/ee/libc/src/stdio.c#L393) where it uses `__direct_pwd` to create an absolute path if a relative path is provided. `__direct_pwd` still exists in [ps2sdkapi.c](https://github.com/ps2dev/ps2sdk/blob/master/ee/libc/src/ps2sdkapi.c) but it doesn't seem to be used anywhere else.

I suspect the path-building logic will need to be copied from the old `fopen()` function into a new function, then `ps2sdkapi.c` and `fileXio_rpc.c` will need to be updated to call this function before calling `fioOpen()` and `fileXioOpen()`, respectively.

root670:
Fixes #134. Majority of this code is copied from [fopen()](https://github.com/ps2dev/ps2sdk/blob/ea69e7cf8e4cea516b18dbdee13f698b8f336257/ee/libc/src/stdio.c#L347) prior to the switch to using newlib for libc.

This should work with fileio and fileXio.

### DIFF
--- a/ee/libc/src/ps2sdkapi.c
+++ b/ee/libc/src/ps2sdkapi.c
@@ -232,6 +232,71 @@ void compile_time_check() {
 	ct_assert(sizeof(uint64_t)==8);
 }
 
+/* Normalize a pathname by removing . and .. components, duplicated /, etc. */
+static char* normalize_path(const char *path_name)
+{
+	int i, j;
+	int first, next;
+	static char out[255];
+
+	/* First copy the path into our temp buffer */
+	strcpy(out, path_name);
+	/* Then append "/" to make the rest easier */
+	strcat(out,"/");
+
+	/* Convert "//" to "/" */
+	for(i=0; out[i+1]; i++) {
+		if(out[i]=='/' && out[i+1]=='/') {
+			for(j=i+1; out[j]; j++)
+					out[j] = out[j+1];
+			i--;
+		;}
+	}
+
+	/* Convert "/./" to "/" */
+	for(i=0; out[i] && out[i+1] && out[i+2]; i++) {
+		if(out[i]=='/' && out[i+1]=='.' && out[i+2]=='/') {
+			for(j=i+1; out[j]; j++)
+					out[j] = out[j+2];
+			i--;
+		}
+	}
+
+	/* Convert "/path/../" to "/" until we can't anymore.  Also convert leading
+	 * "/../" to "/" */
+	first = next = 0;
+	while(1) {
+		/* If a "../" follows, remove it and the parent */
+		if(out[next+1] && out[next+1]=='.' &&
+			out[next+2] && out[next+2]=='.' &&
+			out[next+3] && out[next+3]=='/') {
+			for(j=0; out[first+j+1]; j++)
+				out[first+j+1] = out[next+j+4];
+			first = next = 0;
+			continue;
+		}
+
+		/* Find next slash */
+		first = next;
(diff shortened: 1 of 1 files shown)


rickgaiser:
Hi, and thank you for contributing!

However I don't think this is the correct way of handling the "current working directory" / CWD. Changing the CWD should be performed by the posix function chdir:
https://linux.die.net/man/2/chdir

chdir is already in newlib (I think) and should map to fioChdir and fileXioChdir just like the other functions in ps2sdkapi.c. This will then set the CWD on the IOP side.

For reference, this issue has been described also here:
https://gitlab.com/ps2max/ps2sdk/-/issues/5

root670:
You're right, and I had thought of doing it like that. Are you sure that a `fioChdir()` function exists?

rickgaiser:
You're right, fioChdir does not exist. It's part of the extended ops:
https://github.com/ps2dev/ps2sdk/blob/master/iop/kernel/include/iomanX.h#L80

sp193:
I think this is one thing that does not map nicely to the standard POSIX functions. Usually, you can pass a relative path to the POSIX function and it will resolve to an absolute path. However, we must always specify a device with the Sony functions - but to do that, you must already know what device to use.

root670:
This needs to be done on the EE, otherwise loading iomanX would be required.

rickgaiser:
@root670 I think you're right. If fio does not support this function then it's probably best to do it on the EE side, like you already did.

The code assumes `__direct_pwd` contains the CWD (PWD = CWD ???). `__direct_pwd` is already set by the posix `chdir` function so that's good:
https://github.com/ps2dev/ps2sdk/blob/62733149a8912f412e292b5e8a77452b4b703b7b/ee/libc/src/ps2sdkapi.c#L339-L342
The CWD should include the device name. So for instance "host:sub1/sub2" or "mass:sub1/sub2".

Then when an application is loaded, it's path should be set using `chdir`. But it seems difficult to know from where the ELF file is loaded:
https://github.com/ps2dev/ps2sdk/blob/34b6875d632cf171239624c7acff1452f92792be/ee/libc/src/init.c#L37-L60
Do we know what loaders work with this code? ps2client? uLE? Others?

@sp193 what do you think? Pull this PR? Compared to the old code only `fopen` worked. Now both `fopen` and `open` should work, in the same buggy way as before ;).

sp193:
Nearly all software should have worked with this, since the first argument to any C/C++ program is the path to the program. It's also a behaviour offered by the LoadExecPS2() syscall.
There's an exceptional case for Naplink, perhaps because it called ExecPS2() without any arguments. I think that's wrong, but this sort of software has its own usecases.

Some Sony software (e.g. EELOAD) will have some special path used instead. So the first argument might or might not be a valid path, which the software developer must factor in. However, I don't remember any case where ExecPS2() was called with zero arguments.

PR #137, opened 2020-07-21
Fix printf specifier for _tzname

uyjulian:
Before this change, the string in _ps2sdk_tzname could be named "GMT+-x" for negative timezone offsets


fjtrujy:
Thanks @uyjulian !

rickgaiser:
x should be in hours, not seconds. GMT should be UTC. Right?

uyjulian:
Yes, x should be in hours.  
I've never seen UTC specified with an offset, while I've seen GMT specified with offset many times.

PR #140, opened 2020-08-02
libmc: replace busyloop checking with semaphore checking

uyjulian:
Busyloop checking is replaced with semaphore checking, so other threads can be run while waiting for RPC.

### DIFF
--- a/ee/rpc/memorycard/src/libmc.c
+++ b/ee/rpc/memorycard/src/libmc.c
@@ -156,6 +156,25 @@ static unsigned int g_currentCmd = 0;
 /** specifies whether using MCSERV or XMCSERV modules */
 static int g_mcType = MC_TYPE_MC;
 
+/** semaphore that is signaled when operation is complete */
+static int g_completionSema = -1;
+
+static void mcSignalCompletionSema(void* arg)
+{
+	iSignalSema(g_completionSema);
+}
+
+static int _lock_sema_id = -1;
+static inline int _lock(void)
+{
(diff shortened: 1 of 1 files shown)


rickgaiser:
https://github.com/ps2dev/ps2sdk/blob/1a4587d387022135fbb7eba242c0b10b5b499b8e/ee/kernel/src/sifrpc.c#L179-L201

I like what you're trying to do, but there is no busy waiting loop inside SifCallRpc. Actually, SifCallRpc already uses a semaphore for waiting.

uyjulian:
The busyloop is here: https://github.com/ps2dev/ps2sdk/pull/140/commits/05b422524c62a08e52adb13f921bc92e1c687502#diff-c65198628f56f148984131b4835e6f4cL878

sp193:
As with #141, this change will cause (sce)McSync() to become no longer thread safe and the RPC status becomes inconsistent. If multiple threads call this function, only one will get the signal that the RPC function completed.
The EE kernel does not check on semaphore limits, so I suppose the mechanism might eventually stop working correctly if g_completionSema is allowed to be signalled more than once.

uyjulian:
I copied the design from filexio, so filexio may also have the issue.

uyjulian:
I'll close this, as it would be better to use the filexio functions in some way.

sp193:
The original design is not strictly a less-efficient method; given the design of the EE and IOP kernels, performing this synchronization at the end of your processing or in a thread with a low priority, would result in no loss of useful CPU cycles.

PR #141, opened 2020-08-02
libcdvd: replace busyloop checking with semaphore checking

uyjulian:
Busyloop checking is replaced with semaphore checking, so other threads can be run while waiting for RPC.  

This code is untested


sp193:
I can think of one weakness: the sync function becomes no longer thread safe. You want something like an event flag, which is not implemented within the EE kernel.

uyjulian:
I'll close this, as it would be better to use the filexio functions in some way.

PR #143, opened 2020-08-20
EE SIFRPC: Fix bug in SifRegisterRpc

PSI-Rockin:
This PR fixes a bug in SifRegisterRpc, where if the queue already had registered servers, the function would incorrectly iterate through the next field of the server linked list instead of the link field to find a free space for the server. With the old code, if an RPC request arrived for a queue with multiple registered servers, the SIFRPC protocol would execute not only the correct server, but all servers registered after it as well.

Practically speaking, this bug has little to no influence on current homebrew, as the vast majority of programs do not register any servers, let alone multiple ones to the same queue. However, the official SDK correctly handles this, so it is important that ps2sdk does as well.

### DIFF
--- a/ee/kernel/src/sifrpc.c
+++ b/ee/kernel/src/sifrpc.c
@@ -375,7 +375,7 @@ static void _request_call(SifRpcCallPkt_t *request, void *data)
     (void)data;
 
     if (base->start)
-        base->end->link = server;
+        base->end->next = server;
     else
         base->start = server;
 
--- a/ee/kernel/src/sifrpc.c
+++ b/ee/kernel/src/sifrpc.c
@@ -492,11 +492,11 @@ SifRegisterRpc(SifRpcServerData_t *sd,
     if (!(server = qd->link)) {
         qd->link = sd;
     } else {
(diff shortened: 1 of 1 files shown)


rickgaiser:
After this change the `next` field is only written at initialization with NULL. It is still used in `SifGetNextRequest`, but now always contains NULL. I've searched for `sd->next` and `server->next`. This cannot be correct, but perhaps I'm missing something?

PSI-Rockin:
It looks like you're correct. `_request_call` currently sets the `link` field of the server list to another server if there is already a pending RPC request. This seems to be a bug regardless of my commit... on the EE, the official SDK sets the `next` field instead. This would allow `sceSifGetNextRequest` to process multiple requests correctly, from what I understand.

sp193:
If you're certain that Sony fixed this bug later on or if our logic was just copied incorrectly, then I think it should be corrected.
As you have noted, very few homebrew software use the EE SIFRPC functionality for making IOP to EE calls.

uyjulian:
I've validated that this logic is correct in sifrpc by reverse engineering. I'll go ahead and merge

PR #152, opened 2020-11-24
Doxygen: Add markups for libpacket2

h4570:
Hi.

Added some markups for doxygen, so now, libpacket2 documentation will be visible in github pages :)

Thanks


rickgaiser:
Nice, thanks!

PR #155, opened 2020-11-30
Fixes for gcc11 compatibility, and changes to packet2

rickgaiser:
The vu1 sample now also works using gcc11!
Thanks to @astrelsky for the DVP changes.

For those interested:
- ps2stuff now also compiles using gcc11 (works, but only some simple classes tested)
- ps2gl now also compiles using gcc11 (but none of the samples work)


h4570:
Wow.

I'm glad to hear it!

Cant wait to switch into GCC11.

astrelsky:
Do you have any idea where the samples may be going wrong?

rickgaiser:
I don't know yet where ps2stuff/ps2gl are going wrong. But perhaps the problem is C++:
- new/delete work as expected?
- alignment of new/delete?
- constructors/destructors being called?
- constructors/destructors being called for global objects?
I'm trying to modify @h4570 vu1 example to use ps2stuff (not ps2gl), but I'm already seeing strange behaviour I cannot yet exlain.

ee/packet2/include/packet2_vif.h
```diff
@@ -50,10 +51,10 @@ extern "C"
     {
         assert(packet2->vif_code_opened_at == NULL); // All previous UNPACK/DIRECT are closed.
         packet2->vif_code_opened_at = (vif_code_t *)packet2->next;
-        *((u32 *)packet2->next)++ =
+        packet2_add_u32(packet2,
```
h4570, line 54: I'm must check it, but now ...**open_unpack()**, ..**open_direct()**, ..**_padX()** and a lot of other ..**\_vif()\_**.. methods, will advance **vif_added_bytes**, which should not happen.

What do you think. Is it a good idea, to add parameter like **u8 increaseVif** in **add_u32()**, **add_X()** methods?

rickgaiser: `*((u32 *)packet2->next)++` is no longer valid code for gcc11 (the rvalue @astrelsky also mentioned).

As for design, I think packet2.h should have ALL functions for adding data to the packet. So all oprations on `packet->base` should happen there.
packet2_chain.h cannot use `packet2->base` directly, but must use the functions provided by packet2.h.
packet2_vif.h can use the functions provided by packet2.h and packet2_chain.h.
packet2_util.h can only combine functions from the other classes

I think you're right about vif_added_bytes not being correct anymore :(, but the solution you're proposing will add more vif related code into packet2. The problem is that the "packet2" class should not even know what a 'vif' is. All vif related code and counting should be done by the "packet2_vif" sub-class.

Instead of counting the bytes, we can also have all `vif_open_*` store a pointer the tag. Then on `vif_close_*` calculate the bytes added? That way we can remove all counting from packet2, making it more simple.

h4570: Ah I see..
https://gitlab.com/ps2max/ps2stuff/-/blob/master/src/packet.cpp#L188
`vif_added_bytes` will not be necessary anymore. 
Should I/you do it there?

rickgaiser: Would you like to do it? I'm a little short on time at the moment.

Can you cherry-pick 73bbfd2 and fix it?

Then I'll merge in the other commits.

h4570: Yes, sure.

fjtrujy:
Thanks guys!
You're making it possible, we're closer to use GCC11!

ee/packet2/include/packet2_vif.h
astrelsky: > `*((u32 *)packet2->next)++` is no longer valid code for gcc11 (the rvalue @astrelsky also mentioned).

It is undefined behavior and just wasn't really enforced in earlier versions of gcc. The reason it is ub is because in c/c++ a cast creates a temporary. Writing a value to a temporary (ie an rvalue) doesn't make sense which is why the emitted behavior is undefined.

I think a clean way of handling this could be to use a union so that you can specify which type of pointer you want to use.

```c
typedef union _packet_buf_t {
    u8 *u8;
    u16 *u16;
    u32 *u32;
    u64 *u64;
    u128 *u128;
} packet_buf_t;
```

astrelsky:
> I don't know yet where ps2stuff/ps2gl are going wrong. But perhaps the problem is C++:
> 
> * new/delete work as expected?
> * alignment of new/delete?
> * constructors/destructors being called?
> * constructors/destructors being called for global objects?
>   I'm trying to modify @h4570 vu1 example to use ps2stuff (not ps2gl), but I'm already seeing strange behavior I cannot yet explain.

I think the C++ stuff can be ruled out since iostreams are functioning, specifically `std::cout`. I know from reverse engineering that `std::cout` relies on all of the above (except the alignment) for proper functionality.

I can double check everything except the new/delete alignment via ghidra for sanity. Iirc new/delete just delegate to malloc/free so if whatever checks are in place for that pass it should confirm that the new/delete alignment is correct.

ee/packet2/include/packet2_vif.h
h4570: Here it is guys.

@rickgaiser, https://github.com/h4570/ps2sdk/commit/634c7a3cc32e480936489d0947ec921abec3b095

+ removed `vif_added_bytes`
+ refactored `packet2_*_close_unpack_*`

rickgaiser:
@h4570 can you rebase and send a PR? Then I can pull-merge your changes.

PR #158, opened 2020-12-30
Improve as_reg_compat.h and gcc10 compatibility

rickgaiser:
It can be difficult to find out the register names used by ABI EABI64 and N32. The file as_reg_compat.h should now show crearly the differences, and can also be used to force a .S file to use a specific ABI.

Also all SifBindRpc loops have been made equal, working for both the old and new toolchain. The delay loop with sinple integer counters where optimized by gcc10.

### DIFF
--- a/common/include/as_reg_compat.h
+++ b/common/include/as_reg_compat.h
@@ -10,45 +10,80 @@
 
 /**
  * @file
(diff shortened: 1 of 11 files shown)


fjtrujy:
Thanks @rickgaiser !
Amazing work

PR #160, opened 2021-01-04
GCC10 required updates

rickgaiser:
This PR will update ps2sdk to be compatible with the new GCC10 toolchain for the EE. The current CI (using old toolchain) will fail, becouse not all of these changes are compatible with the old toolchain.

Mostly these are just compatibility fixes, but there are a few things that have been disabled:
- erl-loader: This has never been used as far as I know, and it's been broken for a long time
- mpeg: no-one will probably notice, but I think this should someday be made working with the new toolchain. I will create a github issue about this.
- sio_printf: a function that has been broken since the newlib port. For this I will also create a github issue.

By now the most important ps2 projects (ps2link, RA, OPL, uLE) are compatible with the gcc10 toolchain, so I think it's time to move towards using gcc10 as the default compiler for the EE on the PS2.

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -20,8 +20,8 @@ jobs:
     - name: Compile project
       run: |
         make clean all install
-        ln -sf "$PS2SDK/ee/lib/libps2sdkc.a" "$PS2DEV/ee/ee/lib/libps2sdkc.a"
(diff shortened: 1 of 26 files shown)


fjtrujy:
Moving forward! Thanks @rickgaiser !!!

terremoth:
I consider you both Gods

PR #164, opened 2021-01-22
Update sifcmd.c

psxdev:
Fixed support for user land ee sifcmd registering , it was lost i don't know why.

### DIFF
--- a/ee/kernel/src/sifcmd.c
+++ b/ee/kernel/src/sifcmd.c
@@ -23,6 +23,8 @@
 #define CMD_PACKET_MAX		128
 #define CMD_PACKET_DATA_MAX 	112
 #define SYS_CMD_HANDLER_MAX	32
+#define USR_CMD_HANDLER_MAX	32
+
 
 /* EE DMAC registers.  */
 #define DMAC_COMM_STAT	0x1000e010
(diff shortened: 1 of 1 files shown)


sp193:
You were supposed to call SifSetCmdBuffer(SifCmdHandlerData_t) first.

rickgaiser:
Cross referencing `usr_cmd_handlers` or `nr_usr_handlers` would have been enough to see these values already get used, and shows how a user application can/should use them:
https://github.com/ps2dev/ps2sdk/blob/e8e9edb7309478d71be12bc44f2e087f3d1735a2/ee/kernel/src/sifcmd.c#L305-L314

Please leave changes like these open a little longer next time so others can comment before they are committed. 1 or 2 days would be enough I think.

I think we should revert this commit.

fjtrujy:
Ok @psxdev can you try the solution proposed over here, if this is working, I will revert the change.

Thanks & sorry

sp193:
If I remember right, we originally had a static buffer of a fixed number of entries, hence the OP's question. I probably changed it to fit the official behaviour because it was undocumented; since the handler buffer size is fixed, one could only use SIFCMD IDs up to the last index, but yet this number is not documented.

psxdev:
you can revert the change however if SifAddCmdHandler is called before set the user buffer it will crash without message

sp193:
That's because the documentation wasn't followed. At least it's possible to get the exception trapped. If we want to make this idiot-proof in the sense that it won't crash, then SifAddCmdHandler should probably be made to do nothing - but then it'll become a silent fail. If we printed an error, then this becomes coupled with some library.

PR #165, opened 2021-02-10
Some submissions

AKuHAK:
I created a bunch of commits with different purposes and combined them in one PR. If it is difficult to review files at once, you can check separate commits.
What have been done:
- returned `regress` sample, a nice feature for the testing host: filesystem. Was removed in 00b4130 commit with `prof` sample. `prof` sample wasn't returned cause I don't know what it is used for, and the `-pg` flag isn't working in the newest toolchain.
- poweroff library and sample documented, some disabled samples enabled
- librm library rewritten and documented. Added IR codes for power off/ power on. Can be useful for making debug cycle easier. If you have some programmable IR sender (for example Broadlink RM), you can program IR codes for turning on and off. This feature will work on all models with the built-in receiver (>=SCPH-5xxxx). More info on this project: https://github.com/telyn/ps2runner
- Github Pages, generated with Doxygen was improved, config file updated, added CSS style from @BoGanon repo
- added clang-format (for v12 version) and EditorConfig. I understand that each developer prefers his own coding style (indentation, braces position, comment style), but it makes code hard for understanding. I rewrote the @uyjulian template to feet ps2sdk standards, updated it to version clang-12. I understand that this can break something so I applied clang-format only in projects which I touched and tested. It will be useful to run a clang-format when some dev touches some library. Keep in mind that I used `clang-format-12` which is not in stable releases yet.
- for the same purposes I added EditorConfig. Install Editorconfig plugin in your favorite code editor (I am using Notepad++ for example), and it will fix small errors in indentation and formatting. It is not so advanced as `.clang-format` but still can help a lot.

I planned to fix more stuff, but here we go.


ee/libc/samples/stdio_tests.c
```diff
@@ -0,0 +1,472 @@
...
+static inline int __udelay(unsigned int usecs)
+{
+
+    register unsigned int loops_total = 0;
+    register unsigned int loops_end   = usecs * 148;
+
+    if (usecs > loops_end)
+        return -1;
+
+    __asm__ volatile(".set\tnoreorder\n\t"
```
fjtrujy, line 366: I think that we should use here the delay functions created in `newlib`

AKuHAK: Totally removed asm code as `sleep()` function is doing the same.

AKuHAK:
Btw you can check how Doxygen documentation will look now:
https://akuhak.github.io/ps2sdk/libpwroff_8h.html#ac8d5763cdfcb68960a5fc19ccf2d0fcf

fjtrujy:
Thanks @AKuHAK , merging it!

PR #166, opened 2021-02-12
iop: Add stdint.h

mlaux:
I noticed that we don't have an implementation of the C99 fixed-size integer types. I needed a few of them so I figured I'd add the rest and contribute them all.


uyjulian:
Did you mean to put that in the root?

mlaux:
Oops, messed up copying from an installed ps2sdk to a fresh checkout. Fixed!

rickgaiser:
Thank you, this is real helpfull when porting libraries (like libsmb2), merging.

PR #169, opened 2021-04-03
Update ps2ip.h

dirsors:
This fixes the following compiler warning:
"/usr/local/ps2dev/ps2sdk/ee/include/ps2ip.h:49: warning: `struct timeval' declared inside parameter list
/usr/local/ps2dev/ps2sdk/ee/include/ps2ip.h:49: warning: its scope is only this definition or declaration, which is probably not what you want"


KrahJohlito:
Could be wrong but I don’t think this is correct, wouldn’t the solution just be to #include <time.h>

dirsors:
> Could be wrong but I don’t think this is correct, wouldn’t the solution just be to #include <time.h>

You are right.  #include <time.h> also fix this warning.

fjtrujy:
Thanks!

PR #170, opened 2021-04-30
Multiple small updates/improvements

rickgaiser:
I'm cleaning up all the changes/improvements I've made to my local branches. Please see all separate commits for more information.

One thing to note is that I've reverted #164


fjtrujy:
Thanks for the improvements @rickgaiser !

PR #171, opened 2021-05-10
Some changes to reduce binary sizes

fjtrujy:
These PRs make a collection of changes with the objective of reducing the binary sizes:

1. Separate functions in `ps2sdklibc` library, generating a `.o` object per function. This helps linker to take just the needed code, this strategy was done in most of the libraries in the `ps2sdk`.
2. Reduce `zmax-pag-size`. After upgrading to newer GCC, this value was increased automatically to `64KB`, however, this is not a requirement in the `PS2`, so we can reduce this value to a minimum, in this case, `128 bytes`.

Thanks


rickgaiser:
What is the difference in size before and after this change of a simple hello world ELF?

Thank you for fixing this, the change looks good.

PR #176, opened 2021-05-25
Remove not needed functions from ps2sdklibc because now they are part of newlib

fjtrujy:
## Description 
This PR does some improvements:
- Remove some declared functions from `/src/ee/libc` because now they are included inside of `newlib`
- Remove the weak reference to time, because now `libcdvd` is one of the standard libraries.
- Fixes wrong return code in the I/O function, now we're setting the `errno` making the toolchain more POSIX and standard.
- Update readme explaining there the standard libraries used.
- Extend the regression test to check return value when error opening files.

This PR should be merged together with https://github.com/ps2dev/ps2toolchain-ee/pull/13

Once this PR is accepted and before get merged, we need to update GCC and newlib branch references.


rickgaiser:
Lol, all small comments I made before this was a PR where also visible here. It's a great improvement. Making better use of newlib, and having less custom code. Thanks.

Issue #93, opened 2019-01-20
Fixing -Wunused-value


AKuHAK:
Modern toolchain produces 3 warnings regarding -Wunused-value:
```cpp
src/ps2dev9.c: In function 'dev9_smap_read_phy':
/work/common/include/smapregs.h:87:67: warning: value computed is not used [-Wunused-value]
 #define SMAP_EMAC3_GET(offset)  ((SMAP_EMAC3_REG((offset)) << 16) | \
                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
      (SMAP_EMAC3_REG((offset)+2)))
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                 
/work/common/include/smapregs.h:102:3: note: in expansion of macro 'SMAP_EMAC3_GET'
   SMAP_EMAC3_GET(offset)
   ^~~~~~~~~~~~~~
```
and
```cpp
/work/common/include/smapregs.h:90:81: warning: value computed is not used [-Wunused-value]
 ine SMAP_EMAC3_GET32(offset) (((SMAP_EMAC3_REG32((offset)) >> 16) & 0xffff) | \
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
  ((SMAP_EMAC3_REG32((offset)) & 0xffff) << 16))
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                              
/work/common/include/smapregs.h:106:3: note: in expansion of macro 'SMAP_EMAC3_GET32'
   SMAP_EMAC3_GET32(offset)
   ^~~~~~~~~~~~~~~~
```
and
```cpp
src/smap.c:972:2: note: in expansion of macro 'mSMAP_EMAC3_SET32'
  SMAP_EMAC3_SET32(SMAP_R_EMAC3_INTER_FRAME_GAP, 4);
  ^~~~~~~~~~~~~~~~
/work/common/include/smapregs.h:90:81: warning: value computed is not used [-Wunused-value]
 ine SMAP_EMAC3_GET32(offset) (((SMAP_EMAC3_REG32((offset)) >> 16) & 0xffff) | \
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
[...]

Issue #90, opened 2019-01-19
Fixing -Wmisleading-indentation


AKuHAK:
Just reopen #31 cause I fixed in #88  only part of warnings. There is still some of them:
```cpp
src/callstack.c: In function 'ps2GetStackTrace':
src/callstack.c:200:14: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
         else if (inst == RETURN)
              ^~
src/callstack.c:202:11: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
           ra++;
           ^~
src/callstack.c:205:7: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
       if (sp_adjust == 0 && (found_const_upper || found_const_lower))
       ^~
src/callstack.c:207:9: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
         rc->raOffset = ra_offset;
         ^~
src/stdtrf.c: In function 'stdtrf':
src/stdtrf.c:155:1: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
 if( t < 0 )
 ^~
src/stdtrf.c:158:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  p = 0.5 + 0.5 * p;
  ^
```
Im not completely sure if this is real missing brackets or just indentation problem. For example first warning can bi fixed in such way:
```c
      while ((!found_ra_offset || !found_sp_adjust) && ra < ra_limit)
      {
        inst = *ra;
        /* look for the offset of the PC in the stack frame */
        if ((inst & RESTORE_RETURNVAL_MASK) == RESTORE_RETURNVAL)
        {
          ra_offset = inst & ~RESTORE_RETURNVAL_MASK;
          found_ra_offset = TRUE;
        }
        else if ((inst & RESTORE_RETURNVAL_MASK) == RESTORE_RETURNVAL2)
        {
            ra_offset = inst & ~RESTORE_RETURNVAL_MASK;
            found_ra_offset = TRUE;
        }
        else if ((inst & RESTORE_RETURNVAL_MASK) == RESTORE_RETURNVAL3)
        {
            ra_offset = inst & ~RESTORE_RETURNVAL_MASK;
            found_ra_offset = TRUE;
        }
        else if ((inst & ADJUST_STACKP_C_MASK) == ADJUST_STACKP_C)
        {
[...]

PR #177, opened 2021-06-08
IOP fixes for new GCC version

uyjulian:
Various fixes for various warnings.  

Also, a method to disable usage of GP-relative accesses (and thus disable the save/restore code) was added. For a smooth transition, this will remain enabled for GCC 3.2.3, while it will be properly disabled in GCC 11.

### DIFF
--- a/common/include/hdd-ioctl.h
+++ b/common/include/hdd-ioctl.h
@@ -122,7 +122,7 @@ typedef struct
 {
 	u32 lba;
 	u32 size;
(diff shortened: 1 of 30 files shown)


AKuHAK:
Does this solve https://github.com/ps2dev/ps2sdk/issues/93 and https://github.com/ps2dev/ps2sdk/issues/90 ?

uyjulian:
Yes.

PR #183, opened 2021-06-27
Add backwards compatibility for older MCSERV and PADMAN RPC clients 

uyjulian:
Add backwards compatibility for older `MCSERV` and `PADMAN` RPC clients.

This allows both `libpad` and `libpadx` to be used on the same `PADMAN` module.
This allows `libmc` in both `MC_TYPE_MC` and `MC_TYPE_XMC` modes to be used on the same `MCSERV` module.

This is mainly for compatibility with older software.

### DIFF
--- a/iop/memorycard/mcserv/src/mcserv.c
+++ b/iop/memorycard/mcserv/src/mcserv.c
@@ -20,47 +20,89 @@ IRX_ID(MODNAME, 2, 8);
 	#define DPRINTF(args...)	printf(args)
 #endif
 
-static const u8 XMCSERV_RpcCmd[2][16] =
-{	// libmc rpc cmd values for XMCMAN/XMCSERV
-	{	0xFE,	// CMD_INIT
-		0x01,	// CMD_GETINFO
-		0x02,	// CMD_OPEN
-		0x03,	// CMD_CLOSE
-		0x04,	// CMD_SEEK
-		0x05,	// CMD_READ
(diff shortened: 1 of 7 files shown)


fjtrujy:
I'm fine with the changes, but cannot we do it in a cleaner way?

I saw a lot of `if/else`, so for instance instead of having 2 variables for
```c
static const u8 XMCSERV_RpcCmd[2][17]
static const u8 MCSERV_RpcCmd[2][17] 
```

Can we create something as:
```c
static const u8 MCSERVs_RpcCmd[2][2][17]
```
where the first position of the array is the choice between `MCSERV` and `XMCSERV`?

Then we have a lot of `if (pad_ee_addr).... else....`, I would like to remove all this kind of conditions.

Again this is just a suggestion.

Thanks

uyjulian:
I made some of the suggested changes.

For the `pad_ee_addr` condition, it is needed because the pad structure layout and size are different between RPC clients.

fjtrujy:
Ok, for me is fine the changes.

fjtrujy:
Hello,
@uyjulian now we have integrated Codacy for the Static Analysis in the repos, if you see your PR is failing.
https://app.codacy.com/gh/ps2dev/ps2sdk/pullRequest?prid=7671393

Take a look if you don't mind
Thanks

uyjulian:
I resolved the Codacy issue.

PR #187, opened 2021-06-30
Additional CDVD related definitions

uyjulian:
Add additional CDVD related definitions.

RPC related definitions will be in a separate PR (since I'm still planning a way to generate the RPC bindings)


fjtrujy:
Thanks @uyjulian !!

PR #189, opened 2021-07-03
MX4SIO Work in progress

rickgaiser:
I'm providing the current MX4SIO block device driver "as-is" in this PR. Having the driver in ps2dev/ps2sdk allows it to be included in the CI builds of OPL. Since it's now possible to add it as an optional driver it should be fun to play with, but not very usefull to most in it's current state.

Things left to do:

1. DMA write support, similar to DMA read support. Currently writing seems to work but extremely slow.
2. Better hooking/hacking into sio2man. Becouse PADMAN, MCEMU and MX4SIO all emulate SIO2 by hooking/hacking into sio2man. All 3 use hackish methods. Using different combinations can give different results. I think the emulation of SIO2 should become a separate module. Where the best of all 3 hacks are combined into a SIO2 emulator.

Big problem left to fix:
We are using the serial ports in a way they where not made for. This results in so many problems that I'm not sure we can reliably use this interface without an MCU that knows (or works with) our crippled SPI bus. Problems:
- Most cards do not work
- Card detection and recovery is problematic.
- Small code changes seem to cause the data transmission to become unstable. I have not figured out what is the cause of this. However with CRC16 now added, most errors get detected (at the cost of some speed, 1700 -> 1400KB/s). Sometimes reading the sector again is possible, but sometimes the card becomes unresponsive. And again, error recovery is problematic. Ejecting and reinserting the card is mostly the best way to recover from an error.

I will not continue working on this driver, so I'm hoping it's somewhat usefull in it's current state, and that others can continue the work.


fjtrujy:
Thanks @rickgaiser ! 
Merging

PR #191, opened 2021-07-05
Buildsystem improvements

uyjulian:
Improvements to buildsystem.

Warning/optimization flags are changeable, optimization flags are used at link time (in case you want to use LTO), EE libraries are generated with fat LTO allowing usage in LTO and non-LTO configurations, and unneeded echo statements are removed.


fjtrujy:
Nice to have the `LTO` stuff, I did some testing with it a time ago, and honestly I didn't face too much improvement in terms of performance but is true that compilation times are longer.

Thanks

PR #192, opened 2021-07-07
More IOP definition fixes

uyjulian:
More IOP definition fixes.

### DIFF
--- a/iop/kernel/include/excepman.h
+++ b/iop/kernel/include/excepman.h
@@ -53,8 +53,14 @@
 /** Hardware DeBug(aka "Hardware Breakpoint") */
 #define IOP_EXCEPTION_HDB 15
 
-// does it really return void ?
-typedef void (*exception_handler_t)(void);
+typedef struct _exception_handler_struct_t
+{
+	void* next;
+	int info;
+	u32 funccode[];
+} exception_handler_struct_t;
+
+typedef exception_handler_struct_t* exception_handler_t;
 
 void* GetExHandlersTable();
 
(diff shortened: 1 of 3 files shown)


iop/kernel/include/excepman.h
```diff
@@ -53,8 +53,14 @@
 /** Hardware DeBug(aka "Hardware Breakpoint") */
 #define IOP_EXCEPTION_HDB 15
 
-// does it really return void ?
-typedef void (*exception_handler_t)(void);
+typedef struct _exception_handler_struct_t
```
fjtrujy, line 56: Could we have here alignment problems here?

And a different question is the `_exception_handler_struct_t` needed?

uyjulian: The IOP alignment is 4 bytes so there shouldn't be any alignment problems.

The `_exception_handler_struct_t` is needed so that it can be used as a pointer type. The pointer type will then be used as `exception_handler_t` which should avoid changing the API.

fjtrujy: What you use for creating the pointer is `exception_handler_struct_t` I'm asking about `_ exception_handler_struct_t`, I see this approach in the whole PS2SDK repo, but I think isn't useful at all

uyjulian: There is the structure name `_exception_handler_struct_t`, and there is the typedef name `exception_handler_struct_t`.

If there isn't `_exception_handler_struct_t`, it will create an anonymous structure, which is not valid C11 when used in typedefs.

fjtrujy: I see.
Thanks

PR #198, opened 2021-07-14
Clang format

AKuHAK:


AKuHAK:
The main purpose of this commit - future sync with opl. OPL and ps2sdk share some parts and it will be better to keep it synced at least in formatting.

PR #201, opened 2021-07-20
Add DVRP driver

uyjulian:
Add DVRP driver for DESR / PSX DVR.


AKuHAK:
Wow. Does that mean that we will see `xfrom0:` and `dvr_hdd0:` / `dvr_pfs?:` support in the ps2sdk?

uyjulian:
`xfrom0` is related to flash device, not DVRP.  
Interfaces for `dvr_hdd0` and `dvr_pfs0` will be implemented in a separate PR.

sp193:
Why is this "pvrdrv" instead of "dvrdrv"?

uyjulian:
It is the name in the IRX import table.

PR #204, opened 2021-07-22
libmpeg related fixes

uyjulian:
Libmpeg now builds on new toolchain, but it is unverified.


AKuHAK:
There was some strange Codacy warnings, btw

PR #211, opened 2021-08-08
Improve elf-loader

fjtrujy:
## Description

This PR makes some "improvements" for having a safer scenario when we are going to load another `elf` file.
The PR does:
- Clean `bram` (BIOS RAM) space, before the tiny loader is loaded.
- I have copied some code from `OPL` and `wLE`, now the `IOP` is reset and turned to cleaned state before `ExecPS2` is executed
-  Finally, in order to have a better `cwd` value when launching from `HDD`, I have exposed a new function where you can also specify the partition where the `elf` is.

Additional info:
I have implemented all these changes because I was suffering freezes when launching `RA` from `HDD`. 
How `RA` works, every time a rom is loaded, it `re-load` the whole `elf` (they do this, for avoiding possible memory leaks that belong to the specific core).
So the scenario that was failing, was to "execute the loader" from an elf file that is inside of the `HDD`. 

However, all the changes above didn't make the execution work, the freezes were still happening, then I faced that main problem was that I was forgetting to `deinit` audio and controllers before calling this `elf-loader` in my own app (`RA`).

With this info, I just to make emphasize, that the current `elf-loader` doesn't work under all the circumstances, you as a developer, somehow need to prepare things in your program before calling `ExecPS2`.

Thanks, @uyjulian for helping me debugging this.

Pinging, @rickgaiser @sp193 @uyjulian @parrado

### DIFF
--- a/ee/elf-loader/src/loader/loader.c
+++ b/ee/elf-loader/src/loader/loader.c
@@ -72,18 +72,19 @@ int main(int argc, char *argv[])
 	elfdata.epc = 0;
 
 	GS_BGCOLOUR = WHITE_BG;
-	// arg[0]=path to ELF
-	// arg[1] partition if exists otherwise is ""
+	// arg[0] partition if exists, otherwise is ""
+	// arg[1]=path to ELF
 	if (argc < 2) {  
 		GS_BGCOLOUR = RED_BG;
 		return -EINVAL;
 	}
 
 	char *new_argv[argc - 1];
-	int fullPath_length = 1 + strlen(argv[1]) + strlen(argv[0]);
+	int fullPath_length = 1 + strlen(argv[0]) + strlen(argv[1]);
 	char fullPath[fullPath_length];
-	strcpy(fullPath, argv[1]);
-	strcat(fullPath, argv[0]);
+	strcpy(fullPath, argv[0]);
+	strcat(fullPath, argv[1]);
+	// final new_argv[0] is partition + path to elf
 	new_argv[0] = fullPath;
 	for (i = 2; i < argc; i++) {
 		new_argv[i - 1] = argv[i];
@@ -99,7 +100,7 @@ int main(int argc, char *argv[])
 	FlushCache(0);
 	GS_BGCOLOUR = GREEN_BG;
 	SifLoadFileInit();
-	ret = SifLoadElf(argv[0], &elfdata);
+	ret = SifLoadElf(argv[1], &elfdata);
 	SifLoadFileExit();
 	GS_BGCOLOUR = BLUE_BG;
 	if (ret == 0 && elfdata.epc != 0) {
--- a/ee/elf-loader/src/loader/loader.c
+++ b/ee/elf-loader/src/loader/loader.c
@@ -8,8 +8,12 @@
 # Review ps2sdk README & LICENSE files for further details.
 */
 
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
 #include <kernel.h>
 #include <loadfile.h>
+#include <iopcontrol.h>
 #include <sifrpc.h>
 #include <errno.h>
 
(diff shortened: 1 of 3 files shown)


ee/elf-loader/src/loader/loader.c
```diff
@@ -57,17 +62,33 @@ static void wipeUserMem(void)
...
+	if (argc < 2) {  
 		GS_BGCOLOUR = RED_BG;
 		return -EINVAL;
 	}
+
+	char *new_argv[argc - 1];
+	int fullPath_length = 1 + strlen(argv[1]) + strlen(argv[0]);
+	char fullPath[fullPath_length];
+	strcpy(fullPath, argv[1]);
+	strcat(fullPath, argv[0]);
```
KrahJohlito, line 86: can you add +1 to len and tack a ':' on to the end of argv[1]? or is it better to just include it in partition name when calling?

sp193:
If you consider what Sony documented, there are some things that must be done before LoadExecPS2() or ExecPS2() must be done. There is perhaps no easy way out, as only the developer knows what is necessary to be cleaned up.

Which is something like:
- Delete all semaphores
- Delete all threads
- Disable all interrupt handlers.
- Disable all DMAC interrupt handlers.
- Stop all DMA transfers.
- Deinit SIFRPC (as the SIF involves DMA & interrupts)

This list may have mostly come about because:
- ExecPS2() in the first CEX consoles did nearly no cleanup, in the sense it did not delete any active threads or remove any interrupt handlers. While there is a patch (which we also apply), the recommendation to do some proper deinitialization first still seemed to exist.
- Not stopping DMA transfers (e.g. by PADMAN) may result in memory corruption if the IOP overwrites something important in the new program. 

As with LoadExecPS2(), we're involving an IOP reboot. The recommend actions to take before an IOP reboot are similar, whereby we're supposed to stop all DMA & RPCs.
Not stopping DMA before the IOP reboot may result in the DMAC getting stuck, which may also result in the SIF becoming inoperative and a hard reset would be the only way out.

ee/elf-loader/src/loader/loader.c
fjtrujy: Have in mind that the concatenation of parameters, always happen, so when you load an `elf` that doesn't contain partition it is just `"" + mass:/your.elf`, however, if I add the `:` then it will break the path of the parameter.

I think that is "developer responsibility" to call the function putting in the partition parameter, something finished with the `:`

fjtrujy:
> If you consider what Sony documented, there are some things that must be done before LoadExecPS2() or ExecPS2() must be done. There is perhaps no easy way out, as only the developer knows what is necessary to be cleaned up.
> 
> Which is something like:
> 
> * Delete all semaphores
> * Delete all threads
> * Disable all interrupt handlers.
> * Disable all DMAC interrupt handlers.
> * Stop all DMA transfers.
> * Deinit SIFRPC (as the SIF involves DMA & interrupts)
> 
> This list may have mostly come about because:
> 
> * ExecPS2() in the first CEX consoles did nearly no cleanup, in the sense it did not delete any active threads or remove any interrupt handlers. While there is a patch (which we also apply), the recommendation to do some proper deinitialization first still seemed to exist.
> * Not stopping DMA transfers (e.g. by PADMAN) may result in memory corruption if the IOP overwrites something important in the new program.
> 
> As with LoadExecPS2(), we're involving an IOP reboot. The recommend actions to take before an IOP reboot are similar, whereby we're supposed to stop all DMA & RPCs.
> Not stopping DMA before the IOP reboot may result in the DMAC getting stuck, which may also result in the SIF becoming inoperative and a hard reset would be the only way out.

Thank @sp193 it was really useful info

ee/elf-loader/src/loader/loader.c
KrahJohlito: Oh ok fair enough, didn’t think of that

PR #212, opened 2021-08-09
Add dvr, dvrav, dvripl, dvrmisc modules

uyjulian:
Adds the rest of the DVRP related modules.


AKuHAK:
Is it possible to describe (in few words) what all these modules are used for?

uyjulian:
It's interfaces to the DVRP using structures.

I plan to add documentation of what each devctl does once I get around to checking the EE side of things. This will be in a separate PR

PR #216, opened 2021-08-10
atad: use 48-bit LBA commands on DVRP only when required

uyjulian:
Use 48-bit LBA commands on DVRP only when required.


rickgaiser:
Is this a workaround for the limitations of the DVR hdd? Can the entire hdd be used with this?

uyjulian:
The rest of the HDD can be used only for **read** access.

PR #222, opened 2021-08-12
Detect DESR machines and expand RAM

fjtrujy:
# Description
This PR does 2 things:
- Detect if we are running in a DESR machine
- In case we are, then automatically prepare the app to use the available RAM of DESRV machines

I have extracted from `osd_config` the part related to `ROM0` to a new file called `rom0_info`. Then during TLB init, I'm checking if it is a DESR machine.

I have created an additional sample, which is basically a `mem_test`, it checks the available RAM, and performs write/read to check that the whole memory is working as expected.

I would like if it is possible to be reviewed by @sp193, @rickgaiser, @uyjulian and @AKuHAK .

Thanks


rickgaiser:
The TLB is currently initialized with `InitTLB`. It will automatically configure 32MB or 64MB depending on the result of `GetMemorySize`. But `GetMemorySize` will only return the correct value, if `SetMemoryMode` has been called first. So I think the proper patch would be to do `SetMemoryMode(!IsDESRMachine());` right before `InitTLB` is called. And not add another `_InitTLB` in another place.

But this leads me to something strange: `InitTLB` is called in `TerminateLibrary`. `TerminateLibrary` is called right before we exit or start another elf. But I was expecting `InitTLB` to be called during initialization (like from crt0.o somewhere). It is as if TLB initialization is the job of the loader, and not of the elf? Does this make sense?

Anyway, I don't think we should initialize the TLB, and then initialize it in another place again. We should keep it in 1 place.

fjtrujy:
> The TLB is currently initialized with `InitTLB`. It will automatically configure 32MB or 64MB depending on the result of `GetMemorySize`. But `GetMemorySize` will only return the correct value, if `SetMemoryMode` has been called first. So I think the proper patch would be to do `SetMemoryMode(!IsDESRMachine());` right before `InitTLB` is called. And not add another `_InitTLB` in another place.
> 
> But this leads me to something strange: `InitTLB` is called in `TerminateLibrary`. `TerminateLibrary` is called right before we exit or start another elf. But I was expecting `InitTLB` to be called during initialization (like from crt0.o somewhere). It is as if TLB initialization is the job of the loader, and not of the elf? Does this make sense?
> 
> Anyway, I don't think we should initialize the TLB, and then initialize it in another place again. We should keep it in 1 place.

Hello @rickgaiser,
It was a really good investigation.
And you're totally right, it should be just in one place.
I saw what you said like the `InitTLB` is done before the new `elf` is loaded, so it looks to be "loader responsibility". I don't really like either this approach it sounds weird for me, I would prefer to put the at the beginning of every single execution.

Maybe @sp193 could bring some light to the discussion.

Thanks

fjtrujy:
After do, additional tests, I couldn't use the `speed registers` from `EE`, I don't really know why....

So I went back to the original approach.
I'm going to update PR description

AKuHAK:
@sp193 BTW how to setup TOOL in 128Mb mode? `SetMemoryMode(2)`?

rickgaiser:
Here's what I think should be the way forward in this TLB mess we're currently in:

**Loaders**
I think it's the responsibility of the loaders to create a consistent environment for the application/game to start. Our loaders are also loaded by other loaders (like fmcb->ps2link or fmcb->uLE), so in theory a simple loader does not have to touch the TLB. If the loader wants to configure the TLB again, then it should be set to 32MB, for compatibility.

**Applications**
If an application wants to use non-standard ps2 capabilities, then the application will be responsible for initializing these extra's, and also for reverting these changes on exit (or when loading another aplication).

**ps2sdk**
As to how ps2sdk should handle this I see 2 options:

For ps2sdk applications, simple option:

- on init: do not touch the TLB
- on exit: do not touch the TLB
- on loading another application: do not touch the TLB


For ps2sdk applications, 64MB option:

- on init: change TLB to 64MB if possible
- on exit: return TLB to 32MB for compatibility
- on loading another application: return TLB to 32MB for compatibility

See also the discussion in #228

PR #228, opened 2021-08-24
don't use fio functions in TerminateLibrary

KrahJohlito:
When TerminateLibrary is called there is a good chance all IO services will already be shutdown so only check for DESR at init

### DIFF
--- a/ee/kernel/src/initsys.c
+++ b/ee/kernel/src/initsys.c
@@ -32,7 +32,7 @@ void _InitSys(void)
 void TerminateLibrary(void)
 {
 #ifndef KERNEL_NO_PATCHES
-	InitTLB();
+	_InitTLB();
 #endif
 }
 #endif


AKuHAK:
or maybe doesnt use InitTlb in processes termination at all? This change again will lead to the fact that all external projects which call InitTlb should be fixed. Or just call something like InitTLBMinimal on Terminate?

AKuHAK:
> I agree with @AKuHAK, in my opinion it doesn't have any sense to do the InitTLB when going out, when we are doing already during the init of the apps

I am not quite sure if we can remove it totally from terminating. Will it be safe? It will be useful to set up memory mode to standard, for example if we just are exitng to ps2 OSD, for example

KrahJohlito:
That what’s what I tried first and it didn’t work for me when testing booting apps or games in OPL

AKuHAK:
@KrahJohlito maybe can call _InitTlb instead of InitTlb?

KrahJohlito:
> @KrahJohlito maybe can call _InitTlb instead of InitTlb?

Lol idk it’s up to you guys, I don’t know the difference between _initTLB and InitTLB32MB other than _initTLB is always available on every kernel version as I gathered from the comments.. still don’t get why it needs to be done at terminate though and what it’s doing, that doing it at init of new app isn’t enough but it seems to need it.. I can change it later if you want

rickgaiser:
I think we should close this PR untill a proper solution (discussion in #222) has been found. This PR is creating more mess on top of an already messy situation. If a quick solution is desired then it would be better to revert the comits that created these problems.

KrahJohlito:
well I'm happy to close it that's fine but yes a quick solution is desired

PR #230, opened 2021-08-24
Added htonll/ntohll functions

AKuHAK:


AKuHAK:
New functions are use by OPL NBD server

PR #258, opened 2022-05-23
BDM: add support for Removable media

AKuHAK:
This pr for fixing some issues with USB without MBR scheme.

### DIFF
--- a/iop/fs/bdm/src/part_driver.c
+++ b/iop/fs/bdm/src/part_driver.c
@@ -44,7 +44,7 @@ static struct block_device g_part_bd[MAX_PARTITIONS];
 static struct file_system g_part_fs;
 
 //---------------------------------------------------------------------------
-static int part_getPartitionRecord(struct block_device* bd, part_raw_record* raw, part_record* rec)
+static int part_isPartitionRecordInvalid(struct block_device* bd, part_raw_record* raw, part_record* rec)
 {
     M_DEBUG("%s\n", __func__);
 
@@ -83,7 +83,7 @@ static int part_getPartitionTable(struct block_device* bd, part_table* part)
     if (sbuf[0x1FE] == 0x55 && sbuf[0x1FF] == 0xAA) {
         for (i = 0; i < 4; i++) {
             part_raw = (part_raw_record*)(sbuf + 0x01BE + (i * 16));
-            if (part_getPartitionRecord(bd, part_raw, &part->record[i]) != 0)
+            if (part_isPartitionRecordInvalid(bd, part_raw, &part->record[i]) != 0)
                 return 0; //Invalid record encountered, so the table is probably invalid.
         }
         return 4;
--- a/iop/fs/bdmfs_vfat/src/fat_driver.c
+++ b/iop/fs/bdmfs_vfat/src/fat_driver.c
@@ -995,8 +995,9 @@ int fat_mount(struct block_device* bd)
     // Filter for supported partition IDs:
     // - 0x0b = FAT32 with CHS addressing
     // - 0x0c = FAT32 with LBA addressing
-    if (bd->parId != 0x0b && bd->parId != 0x0c)
-        return -1;
+    // AKuHAK: dont filter partition, fat_getPartitionBootSector() will care about that, this is filtering too much
+    // if (bd->parId != 0x0b && bd->parId != 0x0c)
+    //     return -1;
 
     for (i = 0; i < NUM_DRIVES && fatd == NULL; ++i) {
         if (g_fatd[i] == NULL) {


iop/fs/bdm/src/part_driver.c
```diff
@@ -83,8 +87,25 @@ static int part_getPartitionTable(struct block_device* bd, part_table* part)
     if (sbuf[0x1FE] == 0x55 && sbuf[0x1FF] == 0xAA) {
         for (i = 0; i < 4; i++) {
             part_raw = (part_raw_record*)(sbuf + 0x01BE + (i * 16));
-            if (part_getPartitionRecord(bd, part_raw, &part->record[i]) != 0)
-                return 0; //Invalid record encountered, so the table is probably invalid.
```
AKuHAK, line 87: @rickgaiser I am not sure but should there be == 0 ? I guess that you meant that partition fails if the function returns zero, but it looks like you did it opposite. Can it be a typo? Or maybe I misunderstood something.

iop/fs/bdm/src/part_driver.c
```diff
@@ -129,7 +150,6 @@ int part_connect(struct block_device* bd)
     M_DEBUG("%s\n", __func__);
 
     // Filter for supported partition IDs:
-    // - First sector of disk can be MBR partition
     // - 0x05 = extended MBR partition with CHS addressing
```
rickgaiser, line 153: Is this not true?

iop/fs/bdm/src/part_driver.c
```diff
@@ -234,6 +254,7 @@ void part_init()
     }
 
     g_part_fs.priv          = NULL;
+    // TODO: switch from hardcoded string to variable, at least "MBR" and "Removable"
     g_part_fs.name          = "MBR";
```
rickgaiser, line 258: No, this is the "MBR" file system driver. There is no need for a "Removable" driver, becouse it's already the entire block device.

iop/fs/bdmfs_vfat/src/fat_driver.c
```diff
@@ -995,8 +995,9 @@ int fat_mount(struct block_device* bd)
     // Filter for supported partition IDs:
     // - 0x0b = FAT32 with CHS addressing
     // - 0x0c = FAT32 with LBA addressing
-    if (bd->parId != 0x0b && bd->parId != 0x0c)
-        return -1;
+    // AKuHAK: dont filter partition, fat_getPartitionBootSector() will care about that, this is filtering too much
+    // if (bd->parId != 0x0b && bd->parId != 0x0c)
+    //     return -1;
```
rickgaiser, line 1000: You have not mentioned which problem you are trying to solve in this PR, but could it be this was the only change needed?

If there is no MBR partition found, then the FAT32 file system can mount the entire drive (usb0p0). I see now that this filter would then fail, becouse the parId is always 0 without an MBR.

The filter can be changed to something like this:
```
    // Filter for supported partition IDs:
    // - 0x00 = Possibly a removable drive
    // - 0x0b = FAT32 with CHS addressing
    // - 0x0c = FAT32 with LBA addressing
    if (bd->parId != 0x00 && bd->parId != 0x0b && bd->parId != 0x0c)
        return -1;
```

But it can also be completely removed as you proposed.
Also take a look here at all possible partition types:
https://en.wikipedia.org/wiki/Partition_type

iop/fs/bdm/src/part_driver.c
```diff
@@ -83,8 +87,25 @@ static int part_getPartitionTable(struct block_device* bd, part_table* part)
...
+                     FAT driver will ignore those values anyway */
+                    part->record[i].sid = 0x0C;
+                    part->record[i].start = 0;
+                    // part->record[i].count = sbuf[0x20] + (sbuf[0x21] << 8) + (sbuf[0x22] << 16) + (sbuf[0x23] << 24);
+                    part->record[i].count = 0xFFFFFFFF; // maybe just assign hardcoded maximum value; FAT driver will use its own probing anyway
+                }
+                // if we have magic 0x55 0xAA, partition records cannot fail, we will have at least one removable storage
+                // that will be expanded into mass0p0 for example
+                return (i + 1);
+            }
```
rickgaiser, line 108: If there is no partition table, then the entire block device can be mounted to the FAT driver directly, without an intermediate partition driver.

If there is a partition table then:
- usb0p0 = created by usbmass_bd driver, mounted to MBR file system
- usb0p1 = created by MBR file system, mounted to FAT32 file system

If there is NO partition table then:
- usb0p0 = created by usbmass_bd driver, mounted to FAT32 file system

This change should be removed.

rickgaiser (changes requested):
Great work on finding this bug!

iop/fs/bdm/src/part_driver.c
AKuHAK: Ah thanks I just didnt found that if there is no partition table - then it will fallback to Removable device. If it is true, then this is not needed.

iop/fs/bdm/src/part_driver.c
AKuHAK: Yes, this also was for Removable vs MBR, if there is fallback to REmovabvle this is not needed

AKuHAK:
> You have not mentioned which problem you are trying to solve in this PR, but could it be this was the only change needed?

If there is no MBR partition found, then the FAT32 file system can mount the entire drive (usb0p0). I see now that this filter would then fail, becouse the parId is always 0 without an MBR.

The filter can be changed to something like this:
```
    // Filter for supported partition IDs:
    // - 0x00 = Possibly a removable drive
    // - 0x0b = FAT32 with CHS addressing
    // - 0x0c = FAT32 with LBA addressing
    if (bd->parId != 0x00 && bd->parId != 0x0b && bd->parId != 0x0c)
        return -1;
```

But it can also be completely removed as you proposed.
Also take a look here at all possible partition types:
https://en.wikipedia.org/wiki/Partition_type 


I received reports that usb drives without MBR partitioning is not working. I didnt test this yet, just looked through code.
Probably yes, this is the stopper. There are plenty other FAT system SIDs, not only 0xC/0xB. Various FAT16, and also SID=0 for Removable media. Also FAT driver supports extended partition as well, but this filter exclude those as well.

iop/fs/bdm/src/part_driver.c
AKuHAK: > If `bd->sectorCount` is zero

huh, I mean that `bd->sectorCount` isnt initialized anywhere. What is assigned there before partition manager starts probing everything?

iop/fs/bdm/src/part_driver.c
AKuHAK: yes, I understand that - I just didn't found that switch. I thought that it is doing things in such a way:
check if usb0p0 is MBR
fail

iop/fs/bdm/src/part_driver.c
rickgaiser: The block device drivers create first block device (p0), being the entire drive:

https://github.com/ps2dev/ps2sdk/blob/898cefeb3b2d7cc56bd59a4a92c3cbbd37dd21cd/iop/usb/usbmass_bd/src/scsi.c#L168

https://github.com/ps2dev/ps2sdk/blob/898cefeb3b2d7cc56bd59a4a92c3cbbd37dd21cd/iop/iLink/IEEE1394_bd/src/scsi.c#L168

https://github.com/ps2dev/ps2sdk/blob/898cefeb3b2d7cc56bd59a4a92c3cbbd37dd21cd/iop/sio/mx4sio_bd/src/mx4sio.c#L845

The partition drivers then create new block devices on top of the entire drive block device, like p1, p2, p3.

AKuHAK: Oh, thanks!

iop/fs/bdmfs_vfat/src/fat_driver.c
AKuHAK: FAT16 and Extended also is supported. `fat_getPartitionBootSector` will proceed with necessary checks if this partition is really from the FAT family, so we don't need to care about that.

rickgaiser (approved):

rickgaiser:
If you can confirm both GPT formatted and Removable media formatted drives are working with the current code, then I'll merge the PR.

AKuHAK:
@rickgaiser received the report from @GDX-X that the USB without the partition table started to work again. My USB with MBR partitioned scheme also works fine, I think it is safe to merge this PR.

rickgaiser:
Thanks @AKuHAK ! Merging.

GDX-X:
I confirm works well.

PR #259, opened 2022-05-27
Clang-format some essential parts

AKuHAK:
This was mainly done for sync with other ps2homebrew projects. Currently, there are many code duplications that are hard to track when formatting is so different. For example, cdvdman module in opl. So in future modules can be synced and cleaned.


rickgaiser:
This PR creates a huge change to many files, making it impossible to do git merge/rebase/cherry-pick between versions before and after this change. Plus many more files will change in the future when run through clang-format.

However I do like having a more standard coding style.

@fjtrujy what do you think? Merge?

PR #261, opened 2022-05-31
More IOP definitions

uyjulian:
Adds more definitions from modload, and adds definitions for romdrv.


rickgaiser:
Thanks, merging.

PR #262, opened 2022-05-31
Rename _Exit to KExit to avoid conflicts with _Exit from POSIX

fjtrujy:
## Description
We should have a different name for the function [`_Exit`](https://github.com/ps2dev/ps2sdk/blob/master/ee/kernel/include/kernel.h#L343)

We probably put in the past that name because we are wrapping the syscall function by patching some kernel functions.

Then our custom `Exit` function, after doing our things, calls [`_Exit`](https://github.com/ps2dev/ps2sdk/blob/master/ee/kernel/src/exit.c#L64-L70)

However, `_Exit` (together with `_exit` and others) is a "defined [POSIX](https://man7.org/linux/man-pages/man2/exit.2.html) function" 

So, what's the problem then?
Some libraries/apps/developers could call directly to `_Exit` thinking that it is going to be safe, but it isn't, because we expect `Exit` to called instead of `_Exit`.

This is precisely an example is what I have been suffering when porting `SDL2`.
https://github.com/libsdl-org/SDL/blob/main/src/SDL.c#L107

So in order to avoid this issue, I have renamed the `_Exit` function to `KExit`

Thanks

### DIFF
--- a/ee/kernel/Makefile
+++ b/ee/kernel/Makefile
@@ -20,7 +20,7 @@ ifeq ($(KERNEL_NO_PATCHES),1)
 	ISUSPEND_THREAD_SYSCALL = iSuspendThread.o
 	TLB_SYSCALLS =
 else
-	EXEC_SYSCALLS = _Exit.o _LoadExecPS2.o _ExecPS2.o
+	EXEC_SYSCALLS = KExit.o _LoadExecPS2.o _ExecPS2.o
 	EXECOSD_SYSCALL = _ExecOSD.o
 	ALARM_SYSCALLS = _SetAlarm.o SetAlarm.o _ReleaseAlarm.o ReleaseAlarm.o
 	ALARM_INTR_SYSCALLS = _iSetAlarm.o iSetAlarm.o _iReleaseAlarm.o iReleaseAlarm.o
--- a/ee/kernel/include/kernel.h
+++ b/ee/kernel/include/kernel.h
@@ -340,7 +340,7 @@ void iInvalidDCache(void *start, void *end);
 /* System call prototypes */
 void ResetEE(u32 init_bitfield);
 void SetGsCrt(s16 interlace, s16 pal_ntsc, s16 field);
-void _Exit(s32 exit_code) __attribute__((noreturn));
+void KExit(s32 exit_code) __attribute__((noreturn));
 void _LoadExecPS2(const char *filename, s32 num_args, char *args[]) __attribute__((noreturn));
 s32 _ExecPS2(void *entry, void *gp, int num_args, char *args[]);
 void RFU009(u32 arg0, u32 arg1);
--- a/ee/kernel/include/syscallnr.h
+++ b/ee/kernel/include/syscallnr.h
@@ -8,7 +8,7 @@
 
 #define __NR_ResetEE                  1
 #define __NR_SetGsCrt                 2
-#define __NR__Exit                    4
+#define __NR_KExit                    4
 #define __NR_ResumeIntrDispatch       5 // Arbitrarily named
 #define __NR__LoadExecPS2             6
 #define __NR__ExecPS2                 7
--- a/ee/kernel/src/exit.c
+++ b/ee/kernel/src/exit.c
@@ -65,7 +65,7 @@ const char *SetArg(const char *filename, int argc, char *argv[])
 void Exit(s32 exit_code)
 {
     TerminateLibrary();
-    _Exit(exit_code);
+    KExit(exit_code);
 }
 #endif
 
--- a/ee/kernel/src/kernel.S
+++ b/ee/kernel/src/kernel.S
@@ -100,7 +100,7 @@ SYSCALL(SetGsCrt)
 #ifdef KERNEL_NO_PATCHES
 
 #ifdef F_Exit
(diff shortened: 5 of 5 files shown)


uyjulian (approved):
lgtm

fjtrujy:
Additional info:
`newlib` already has a `_Exit` implementation, and we were already using it, however `newlib` by default creates weak references.
https://github.com/ps2dev/newlib/blob/master/newlib/libc/stdlib/_Exit.c

rickgaiser (changes requested):
I understand there's a mess with many exit functions with almost the same name, but I do not see how this change makes the situation any better, besides fixing 1 issue with SDL2.

Current exit functions:
- `_Exit` is the PS2 bios exit function
- `Exit` is the ps2sdk wrapper that calls `TerminateLibrary` first, then `_Exit`
- `exit` is the posix exit function - defined in ps2sdk/ee/libc that calls `_exit`
- `_exit` is the crt0.o exit function that calls `_ps2sdk_libc_deinit` then `Exit`
also in newlib:
- `_Exit` that calls `_exit` <- conflicting with ps2sdk
- `exit` that calls `_exit` <- conflicting with ps2sdk

Functionality:
- when main returns: `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `_Exit `
- when `exit` called: `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `_Exit `

To me this looks like 1 'hook' or 'wrapper' too many. It should be more like:
- `_exit` -> `_ps2sdk_deinit` -> `Exit`

With `_exit` being the crt0 exit (also when main exits), `_ps2sdk_deinit` doing all that's needed to deinit ps2sdk, and fignally `Exit` to the ps2 bios.

> Some libraries/apps/developers could call directly to `_Exit` thinking that it is going to be safe, but it isn't, because we expect `Exit` to called instead of `_Exit`.

These libraries and apps should call the posix `exit` function, instead of the ps2 specific `_Exit`/`_exit`/`Exit`/`KExit`/... function. And we must make sure `exit` works as expected.

**edit**: added newlib functions to the list

fjtrujy:
> Functionality:
> 
> * when main returns: `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `_Exit `
> * when `exit` called: `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `_Exit`

I think that you are partially wrong.

As I said before `_Exit` is a public POSIX function. This `_Exit` function is offered by `newlib` as we saw it [here](https://github.com/ps2dev/newlib/blob/master/newlib/libc/stdlib/_Exit.c).

So these is the list of public POSIX functions:
* [`noreturn void exit(int status);`](https://man7.org/linux/man-pages/man3/exit.3.html) 
* [`noreturn void _exit(int status);`](https://man7.org/linux/man-pages/man2/exit.2.html)
* [`noreturn void _Exit(int status);`](https://man7.org/linux/man-pages/man2/exit.2.html)

How these are the public functions, we can suppose that developers and/or libraries can use them.
In total we have the next scenarios: 

* when main returns: `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `KExit `
* when `exit` called: `exit` -> `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `KExit`
* when `_exit` called: `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `KExit`
* when `_Exit` called: `_Exit` -> `_exit` -> `_ps2sdk_libc_deinit` -> `Exit` -> `TerminateLibrary` -> `KExit`

Conclusion:
- We must rename our `_Exit` to something else, `KExit` for me is a good name, you can suggest a better one, but for sure we can't keep `_Exit` as it is, because it could make libraries to don't work properly.
- We need to remove the custom `exit` implementation from [PS2SDK](https://github.com/ps2dev/ps2sdk/blob/11500fe1dcec03592e71406cbd9d34b39aa52589/ee/libc/src/terminate.c#L24-L29) because we have it already defined in [newlib](https://github.com/ps2dev/newlib/blob/master/newlib/libc/stdlib/exit.c)

fjtrujy:
Additionally `abort` function can be also removed from `PS2SDK` because it is included in `newlib` as well https://github.com/ps2dev/newlib/blob/master/newlib/libc/stdlib/abort.c

So I'm going to remove the whole `abort.c` file

rickgaiser:
> So I'm going to remove the whole terminate.c file

Nice cleanup!

Ok,I don't want to make this a big deal, so lets rename _Exit to KExit as suggested.

Btw, can this also be removed?:
https://github.com/ps2dev/ps2sdk/blob/a322f4867e0ee0a29310b56f793067d584d8fb46/ee/kernel/include/kernel.h#L509-L512

Perhaps also this?:
https://github.com/ps2dev/ps2sdk/blob/a322f4867e0ee0a29310b56f793067d584d8fb46/ee/kernel/include/kernel.h#L514-L518

rickgaiser (approved):

fjtrujy:
Yes these lines can be removed as well, I will remove them.

PR #263, opened 2022-05-31
Fatfs

JoseAaronLopezGarcia:


fjtrujy:
Please provide a full description about what is actually adding this PR

rickgaiser (changes requested):
Code looks good to me, but some small points:
- chmod -x the files
- FF_FS_TINY should probably be 1 (the default) since this driver is not meant for ingame (how much kb is the difference?)

USBMASS_IOCTL_GET_CLUSTER
returns 0 on exFat ?

USBMASS_IOCTL_NUM_FRAGS
returns 0 (tested with exFat), so this can't be correct

USBMASS_IOCTL_CHECK_CHAIN
is based on the same `n_frag`, so here must also be a problem. Change the compare to `n_frag==1` and you'll see the problem.

USBMASS_IOCTL_NEXT_CLUSTER
Does not work with the cluster provided by USBMASS_IOCTL_GET_CLUSTER

Perhaps for now remove all added ioctl's (added for defragmentation support) so we can get a working replacement for the old vfat driver. Then later try adding the defragmentation support in a separate PR.

AKuHAK:
> I don't think we're in a rush to merge this

@JoseAaronLopezGarcia 
So what about removing fragmentation support as for now?

JoseAaronLopezGarcia:
> > I don't think we're in a rush to merge this
> 
> @JoseAaronLopezGarcia So what about removing fragmentation support as for now?

DO I just comment the code or remove it entirely?

JoseAaronLopezGarcia:
Changed the title to WiP as I've been requested by homebrew devs to backport fatfs to the older (pre-BDM) driver for some older apps that use it.

rickgaiser:
If there's apps using the pre-bdm driver (usbhdfsd?) then these apps can be upgraded simply by replacing usbhdfsd, with bdm+usbmass_bd+<filesystemofchoise>. Its like changing 3 lines of code, and the rest is compatible.

Creating a new driver with **copies** of both the FatFs driver AND the usbmass driver is a bad solution.

JoseAaronLopezGarcia:
> If there's apps using the pre-bdm driver (usbhdfsd?) then these apps can be upgraded simply by replacing usbhdfsd, with bdm+usbmass_bd+. Its like changing 3 lines of code, and the rest is compatible.
> 
> Creating a new driver with **copies** of both the FatFs driver AND the usbmass driver is a bad solution.

Alright then I'll just make the old driver with fatfs as a different project. I agree homebrews should change to the new BDM drivers.

rickgaiser:
@JoseAaronLopezGarcia since I'm now making and testing the fragmentation support I think it's best to do that in a separate PR. Becouse the fragmentation support is a new feature that can lead to all sorts of new discussions, plus how do I get my changes in your (this) PR?

Adding FatFs is a great feature of it's own. If you remove the extra ioctl commands then I can squash-merge this PR. Do you also happen to have a PR ready for OPL once this gets merged?

common/include/usbhdfsd-common.h
```diff
@@ -25,7 +25,15 @@
...
-#define USBMASS_IOCTL_CHECK_CHAIN    0x0004
+#define USBMASS_IOCTL_CHECK_CHAIN	 0x0004
+/** Return number of fragments **/
+#define USBMASS_IOCTL_NUM_FRAGS      0x0005
+/** Return first cluster **/
+#define USBMASS_IOCTL_DATA_BASE      0x0006
+/** Return next cluster **/
+#define USBMASS_IOCTL_NEXT_CLUSTER   0x0007
+/** Return cluster size **/
+#define USBMASS_IOCTL_SIZE_CLUSTER   0x0008
```
rickgaiser, line 36: Undo this change.

rickgaiser (changes requested):

common/include/usbhdfsd-common.h
JoseAaronLopezGarcia: Done

rickgaiser (approved):

PR #264, opened 2022-06-04
Kernel lib changes

uyjulian:
More symbols and changes to make it more inline with official software.


sp193:
Did you really find an SDK that checks whether timer 3 is running, before applying the alarm patch?

fjtrujy:
I think that we can split it in 2 different PRs.

uyjulian:
> Did you really find an SDK that checks whether timer 3 is running, before it applies the alarm patch?

Yes, in 3.0.3

PR #265, opened 2022-06-04
[Don't merge] Add timer related functionality

uyjulian:
**Untested** by myself.  

Requested by fjtrujy

### DIFF
--- a/ee/kernel/src/timer.c
+++ b/ee/kernel/src/timer.c
@@ -13,7 +13,1179 @@
  * Some routines to do some timer work
  */
 
-#include <tamtypes.h>
+#include <kernel.h>
+#include <timer.h>
+#include <string.h>
+#include <stdint.h>
+
+#ifdef F_timer_data
+timer_ee_global_struct g_Timer;
+timer_ee_entry_struct g_CounterBuf[128];
+alarm_ee_entry_struct g_AlarmBuf[64];
+alarm_ee_entry_struct *g_pFreeAlarm = NULL;
+#else
+extern timer_ee_global_struct g_Timer;
+extern timer_ee_entry_struct g_CounterBuf[128];
+extern alarm_ee_entry_struct g_AlarmBuf[64];
+extern alarm_ee_entry_struct *g_pFreeAlarm;
+#endif
+
+#ifdef F_SetT2
+s32 SetT2(volatile void *a1, u32 a2)
+{
+	u32 v2;
+	u32 v3;
+	u32 v5;
+
+	__asm__ __volatile__ ("mfc0\t%0, $12" : "=r" (v3));
+	v3 = v3 & 0x10000;
+	if (v3 != 0)
+	{
+		do
+		{
+			DI();
+			EE_SYNCP();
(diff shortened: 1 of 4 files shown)


fjtrujy:
I have tested so far `DelayThread` and it is not working.
Let me try with `SetTimerAlarm`

ee/kernel/src/timer.c
```diff
@@ -13,7 +13,1179 @@
...
+#endif
+
+#ifdef F_InitTimer
+s32 InitTimer(s32 a1)
+{
+	s32 v6;
+	u32 oldintr;
+	u32 v8;
+
+	if (g_Timer.intc_handler >= 0)
```
fjtrujy, line 93: I think that this comparison should be `if (g_Timer.intc_handler > 0)` otherwise the timer won't do the init properly because by default the `g_Timer` is empty

fjtrujy: Or the initial value of `g_Timer` is wrong defined

davidgfnet:
I'd be against merging this for several reasons:
 - It is a C version of some ASM code, as variable names suggest (and inline asm)
 - Variables that came from .data section are not... initialized properly, for instance g_Timer
 - It obviously doesn't work (since initial state is bad)
 - Is this really needed?
 - Other reasons probably copyright related :)

davidgfnet:
Don't get me wrong this is very useful, but perhaps could do better in a place like ps2tek.
I'm working on a similar thing for the SDK and i'm checking on this to have an idea on how some stuff works.
Thanks!

AKuHAK:
@davidgfnet do you plan to contribute into ps2sdk?

uyjulian:
> I'd be against merging this for several reasons:
> 
> * It is a C version of some ASM code, as variable names suggest (and inline asm)
Most of the code in ps2sdk is already reverse engineered from asm, so I don't see a problem with this.
> * Variables that came from .data section are not... initialized properly, for instance g_Timer
I plan to fix it later
> * It obviously doesn't work (since initial state is bad)
I plan to fix it later
> * Is this really needed?
Yes, mainly for `DelayThread`, a more accurate `nanosleep` implementation that doesn't rely on busyloops, and also for a future implementation of `libipu` which depends on this.
> * Other reasons probably copyright related :)
If this code has a problem, then all of the RE'd code in ps2sdk already has a problem.

davidgfnet:
Oh I see, thanks for your comments! And sorry, re-reading my comments I sounded very harsh, didnt inted to!
Yeah indeed, was already working on something similar (since fjtrujy is working on a new SDL2 port and we came across the same sleep() issue you guys mention). I created PR #266 which might require some review and fixing, since I'm not an expert in how the SDK works (and the library organization).
It implements a similar mechanism using the TIM2 timer. For clock calculations it uses ps2_clock, which uses TIM0/1 clock as a stable time source.
Let me know what you think!
Also, while doing this I found a small (but potentially important) issue with ps2_clock, I will send a fix for it.

fjtrujy:
@uyjulian I think that we can close already this PR, thank you very much because it was really useful for have it as reference

uyjulian:
Yes

PR #266, opened 2022-06-05
Add a timer-based alarm system based on TIM0/1

davidgfnet:
Using TIM2 we can schedule events in the future. An event queue is used
for the purpose of book keeping, holding the events in chronological
order. A user callback is provided (running as a IRQ handler).

This can be used for accurate thread sleeping (by using BIOS sempahores)


ee/kernel/include/timer_alarm.h
```diff
@@ -0,0 +1,61 @@
...
+// Sleeping function
+void ThreadWaitClock(u64 clock_cycles);
+
+// Conversion functions, from time to bus cycles
+static inline u64 USecToCycles(u64 usec) {
+    return (147456 * usec) >> 10;  // approximate
+}
+
+static inline u64 MSecToCycles(u64 msec) {
+    return 147456 * msec;
```
fjtrujy, line 53: there are constans already defined for these values `kBUSCLK` and family in `ee/kernel/include/timer.h`

ee/kernel/src/initsys.c
```diff
@@ -14,10 +14,12 @@
 
 #include "kernel.h"
 #include "string.h"
+#include <timer_alarm.h>
 
 #ifdef F__InitSys
 void _InitSys(void)
 {
+    InitTimerAlarm();
```
fjtrujy, line 22: I think that following the other implementations it should be inside of the `#ifndef KERNEL_NO_PATCHES` what do you think @rickgaiser ?

ee/kernel/src/initsys.c
```diff
@@ -30,6 +32,7 @@ void _InitSys(void)
 #ifdef F_TerminateLibrary
 void TerminateLibrary(void)
 {
+    DeinitTimerAlarm();
```
fjtrujy, line 35: same here?

fjtrujy:
@davidgfnet thanks you very much for this PR it is going to be really useful for fixing the `nanosleep` implementation which is not working fine at all.
This PR will invalidate the one created yesterday by @uyjulian #265 

Additionally, thank you again for creating the samples, it makes the testing phase easier.

ee/kernel/src/initsys.c
davidgfnet: IIUC the KERNEL_NO_PATCHES define protects some functions that "patch" the BIOS (it literally writes over the BIOS memory patching some bugs in known code).
However I'm not doing any sort of patching.
Perhaps is not InitSys the right place to initialize an SDK lib? (In any case the ifdef doesnt apply here I think).

ee/kernel/include/timer_alarm.h
davidgfnet: Indeed!

ee/kernel/src/initsys.c
fjtrujy: Ok, got it, I think it is the proper place where it is then

ee/kernel/src/timer_alarm.c
```diff
@@ -0,0 +1,264 @@
...
+}
+
+static void _wake_sema_cb(struct timer_alarm_t *alarm, void *userptr) {
+    (void)alarm;
+
+    s32 sema_id = *(s32*)userptr;
+    iSignalSema(sema_id);
+}
+
+void ThreadWaitClock(u64 clock_cycles) {
```
fjtrujy, line 250: how does it behave with `0` as input?

davidgfnet: For consistency we guarantee that all alarms will trigger. This is important if there are many alarms close in time, we do not want to "miss" one of them. Thats why in the code, when we calculate the timer cycles we do this:

```
s64 wait_time = (s64)alarm->scheduled_time - (s64)(ps2_clock() * 256);
u64 alarm_clocks = wait_time <= 0 ? 1 : wait_time;```

Which means: if the alarm is in the past or present (negative delay or zero), we schedule 1 cycle, so that the timer will be triggered immediately. Again, for consistency. With zero of very small values the alarm triggers as soon as the StartTimerAlarm exits since the timer runs out almost immediately.

Comment (fjtrujy) [reply ee/kernel/src/timer_alarm.c:250]: That's a really good approach.

**Comment by fjtrujy** [comment]
@rickgaiser should we update the `nanosleep` implementation and use use something like this instead?
```c
static inline u64 timespec_to_nanosecs(const struct timespec *pts) {
    return (pts->tv_sec * 1000000000ULL) + pts->tv_nsec;
}

int nanosleep(const struct timespec *req, struct timespec *rem)
{
    u64 nano_secs, cycles;

    nano_secs = timespec_to_nanosecs(req);
    cycles = NSecToCycles(nano_secs);
    ThreadWaitClock(cycles);

    if (rem != NULL) {
        rem->tv_sec = 0;
        rem->tv_nsec = 0;
    }

	return 0;
}
```

Should we do it in this PR? or do you prefer to create a separate one for it?

ee/kernel/src/timer_alarm.c
```diff
@@ -0,0 +1,264 @@
...
+/** Same as StartTimerAlarm but for IRQ handlers.
+ *
+ */
+void iStartTimerAlarm(struct timer_alarm_t *alarm) {
+    // Bail if the alarm is already set!
+    if (alarm->scheduled_time)
+        return;
+
+    // Calculate clock scheduling time
+    u64 sched_time = ps2_clock() * 256 + alarm->timer_cycles;
```
fjtrujy, line 176: `ps2_clock` belong to `ee/libc/src/time.c` which is initialised with `_ps2sdk_time_init`, here we can have something wrong if `_ps2sdk_time_init` is not called (it is currently a `weak function`).
Maybe we should move that implementation to `kernel` and do the init always?
Pinging @rickgaiser

fjtrujy:
Overall the PR is almost ready to be merged. 
For me, it is just pending IMO to move `ee/libc/src/time.c` to `ee/kernel/src/time.c` and do the init always in the `_InitSys` method.

However, this change may require some changes in other libraries/ports, even like this I prefer to move it because we will have a cleaner implementation.

Once again, thanks @davidgfnet !

ee/kernel/src/initsys.c
uyjulian: It should be indeed inside `#ifndef KERNEL_NO_PATCHES`, for those wanting to reduce code size (e.g. a memory card boot program).  
And if the user wants to initialize it manually using `libkernel-nopatch`, then they can do so, and the code will get linked. Otherwise it won't get linked and will not increase the code size.  
When using `libkernel`, it will be linked always.

davidgfnet:
Yeah I agree. The time.c just takes ~300 bytes, less if you do not use it (and only initialize it). I think it's worth merging it with the other kernel bits, so that every library can rely on ps2_clock.
Things like ee/network/tcpip/src/sys_arch.c could benefit from using ps2_clock and the functionality added in this PR, since they seem to still use some hsync based interrupt mechanism that depends on things like screen resolution, which is definitely sub-optimal.

fjtrujy:
@davidgfnet I think that we should extend these PRs including the next things if possible:
1. Move `ee/libc/src/time.c` and join it with `ee/kernel/src/time.c`
2. Try to fix the `ps2_clock` implementation #267 
3. Use your new functionality to implement [`nanosleep`](https://github.com/ps2dev/ps2sdk/pull/266#issuecomment-1146776197)
4.  Add some tests example for testing `nanosleep` and `ps2_clock`

fjtrujy:
@davidgfnet you have forgotten to add the tests to the `samples/Makefile_sample` file

davidgfnet:
This needs some rework, since TIM2 is not user-usable (without some workarounds).
Let's merge PR #274 first and I will update this then.

fjtrujy:
I have just tested all the examples in real hardware and PCSX2:

1. `test_alarm`
```
Hello there, sqrt(2) second tick!
Hello there, 2 second tick!
Hello there, sqrt(2) second tick!
Hello there, 2 second tick!
Hello there, sqrt(2) second tick!
Hello there, sqrt(2) second tick!
Hello there, 2 second tick!
Hello there, sqrt(2) second tick!
Hello there, 2 second tick!
Hello there, sqrt(2) second tick!
Hello there, sqrt(2) second tick!
Hello there, 2 second tick!
Hello there, sqrt(2) second tick!
Hello there, 2 second tick!
Hello there, sqrt(2) second tick!
Hello there, 2 second tick!
Hello there, sqrt(2) second tick!
```

2. `timer_alarm_sleep`
```
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
Hello there, 350ms tick!
```

3. `timer_alarm_stop`
```
2s alarm triggered!
2s alarm triggered!
5s alarm triggered!
5s alarm triggered!
5s alarm triggered!
5s alarm triggered!
5s alarm triggered!
```


Thanks @davidgfnet , super ready to be merged

PR #269, opened 2022-06-07
Add specific Kernel function for getting SystemTime

### LINKED ISSUE #267: `ps2_clock` is not always increasing
Having the following test:

```c
int main()
{
   uint64_t index;
   ps2_clock_t current, previous;
   previous = ps2_clock();
   index = 0;
   while (1)
   {
      current = ps2_clock();
      if (current < previous) {
         printf("Something is wrong with ps2_clock, index: %llu, previous: %llu, current:%llu\n", index, previous, current);
      }
      
      index++;
      previous = current;
   }
   return 0;
}
```

It produces the following output:
```
Something is wrong with ps2_clock, index: 24394150, previous: 983039, current:917504
Something is wrong with ps2_clock, index: 27645769, previous: 1114111, current:1048576
Something is wrong with ps2_clock, index: 30897299, previous: 1245183, current:1179648
Something is wrong with ps2_clock, index: 35775553, previous: 1441791, current:1376256
Something is wrong with ps2_clock, index: 40653964, previous: 1638399, current:1572864
Something is wrong with ps2_clock, index: 45532208, previous: 1835007, current:1769472
Something is wrong with ps2_clock, index: 76438234, previous: 3080191, current:3014656
Something is wrong with ps2_clock, index: 86196797, previous: 3473407, current:3407872
```

I don't know exactly where the problem is, probably @davidgfnet has more knowledge about it.

fjtrujy:
## Description
This PR tries to make easier the usage of the `cpu ticks` function,

Currently, it was done through the function `ps2_clock` which was in the `ee/src/libc`. This function has an issue as we can see here #267 . 
So following Sony's recommendation we have moved this into the `kernel` which is where it should be. So we have created now functions for getting the `cpu_ticks`, they are:
- `GetTimerSystemTime`
- `iGetTimerSystemTime`

In order to keep retro-compatibility we have also kept the current `ps2_clock`

Finally, I have added a new test where we can check if it is working as expected.

Test execution output 
```
Starting ps2_clock example!
Diff:0
Diff:73
Diff:68
Diff:67
Diff:68
Diff:74
Diff:67
Diff:68
Diff:68
Diff:67
Diff:74
Diff:67
.........
```

Closes #267


ee/kernel/src/timer.c
```diff
@@ -13,7 +13,147 @@
...
+   return -1;
+}
+
+/** Init the timer
+ *
+ *
+ * Start the interruption handler for getting the system time
+ * This function should be called at beginning of your program
+ */
+void InitTimer(void)
```
rickgaiser, line 60: The naming of this function pair (InitTimer / EndTimer) is misleading.

What about StartTimerSystemTime and StopTimerSystemTime (as in @uyjulian PR) ?

ee/kernel/src/timer.c
```diff
@@ -13,7 +13,147 @@
...
+{
+    u32 oldintr;
+    
+    *T_MODE = 0x0000; // Disable T_MODE
+
+    oldintr = DIntr();
+   if (__time_intr_overflow_id == -1)
```
rickgaiser, line 67: indentation :)

ee/kernel/src/timer.c
```diff
@@ -13,7 +13,147 @@
...
+ */
+u64 iGetTimerSystemTime(void) {
+    u64 t;
+    u32 oldintr = DIntr();
+    t = GetTimerSystemTime();
+    if (oldintr)
+        EIntr();
+
+    return t;
+}
```
rickgaiser, line 155: This function **should not** enable/disable interrupts. Becouse it should only be called from an interrrupt disabled state.

Also, this function should return BUSCLK resolution: `return (t<<8);`

ee/kernel/src/timer.c
```diff
@@ -13,7 +13,147 @@
...
+   // Tn_COUNT is 16 bit precision. Therefore, each __time_intr_overflow_count is 65536 ticks
+   t = *T_COUNT + (__time_intr_overflow_count << 16);
+
+   // check if the overflow was produced but the interruption callback isn't be executed yet
+   if (((*T_MODE) & (1 << 11)) != 0) {
+      t += (1 << 16);
+   }
+
+   return t;
+}
```
rickgaiser, line 134: This function **should** disable/enable interrupts, becouse it can be called from an interrupt enabled state.

Also, this function should return BUSCLK resolution: `return (t<<8);`

ee/libc/include/ps2sdkapi.h
```diff
@@ -39,7 +40,10 @@ extern int (*_ps2sdk_closedir)(DIR *dir);
 #define PS2_CLOCKS_PER_MSEC (PS2_CLOCKS_PER_SEC / 1000) // 576
 
 typedef uint64_t ps2_clock_t;
-ps2_clock_t ps2_clock(void);
+static inline ps2_clock_t ps2_clock(void) {
+    // DEPRECATED VERSION USE INSTEAD GetTimerSystemTime
+    return (ps2_clock_t)GetTimerSystemTime();
+}
```
rickgaiser, line 46: This function should return BUSCLK/256 resolution: `return (ps2_clock_t)GetTimerSystemTime()>>8;`

rickgaiser (changes requested):
With a few small changes I think this PR can be more easily merged than the PR from @uyjulian . Mostly becouse it's code we have written ourselves and can read and understand better. But at the same time I think we can learn a lot from the code in the @uyjulian PR.

ee/kernel/src/timer.c
fjtrujy: Yes! Doing it.

ee/kernel/src/timer.c
fjtrujy: Fixed!

ee/kernel/src/timer.c
fjtrujy: Yes I missed up between iGetTimerSystemTime and GetTimerSystemTime

ee/libc/include/ps2sdkapi.h
fjtrujy: Yes I missed up between iGetTimerSystemTime and GetTimerSystemTime

fjtrujy:
@rickgaiser just one thing, I know that you are suggesting to use the timer with the maximum accuracy, and then in `ps2_clock` do `GetTimerSystemTime() >> 8`, however, I don't if it makes sense at all.

Pros:
- We will have a more accurate system timer

Cons:
- It will overflow x8 times faster, wasting useless CPU cycles.

IMO I think that the current timer accuracy is good enough, I prefer to keep it as it was (it is just my personal opinion)

ee/kernel/src/timer.c
fjtrujy: Yes I missed up between iGetTimerSystemTime and GetTimerSystemTime

uyjulian:
Yeah I think the current timer accuracy is good enough also

uyjulian (approved):

rickgaiser:
GetTimerSystemTime is a function that should always return a value in BUSCLK units. No matter what resolution the internal timer is running on. So if we have the timer running at BUSCLK/16, we multiply the value by 16 before returning, and if we have the timer running at BUSCLK/256, we multiply the value by 256 before returning. Just like the reverse engineered code does.

> It will overflow x8 times faster, wasting useless CPU cycles.

256x faster actually, but only if we use that resolution. And for now, by default, we're not going to use that resolution. But to keep the interface of the function the same as the sony function I think we should return in BUSCLK units.

The advantage of this standardized interface is that when someone needs a higher resolution timer, and changes the code, the interface of the function will stay the same, and no other code or library needs to be changed.

fjtrujy:
> GetTimerSystemTime is a function that should always return a value in BUSCLK units. No matter what resolution the internal timer is running on. So if we have the timer running at BUSCLK/16, we multiply the value by 16 before returning, and if we have the timer running at BUSCLK/256, we multiply the value by 256 before returning. Just like the reverse engineered code does.
> 
> > It will overflow x8 times faster, wasting useless CPU cycles.
> 
> 256x faster actually, but only if we use that resolution. And for now, by default, we're not going to use that resolution. But to keep the interface of the function the same as the sony function I think we should return in BUSCLK units.
> 
> The advantage of this standardized interface is that when someone needs a higher resolution timer, and changes the code, the interface of the function will stay the same, and no other code or library needs to be changed.

Ok, now I understand it, and it makes sense,
I have applied the changes

rickgaiser (approved):

PR #270, opened 2022-06-08
Add (de)fragmentation support to fat32/exFat

rickgaiser:
This PR adds a new ioctl2 command: `USBMASS_IOCTL_GET_FRAGLIST`. This lets you request the number of fragments, and a list of the location and sizes of up to 10 fragments.

With this fragment list, the file system driver is no longer needed to read a file with up to 10 fragments.
To process this fragment list, a new library `libbdm` has been added.

Also in this PR all compile warnings of `bdmfs_fatfs` have been fixed.

### DIFF
--- a/common/include/usbhdfsd-common.h
+++ b/common/include/usbhdfsd-common.h
@@ -15,6 +15,20 @@
 #ifndef __USBHDFSD_COMMON_H__
 #define __USBHDFSD_COMMON_H__
 
+#include <tamtypes.h>
+
+#define BD_MAX_FRAGMENTS 10
+
+typedef struct bd_fragment {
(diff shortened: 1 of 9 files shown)


fjtrujy (approved):
Super!

AKuHAK:
@rickgaiser just to be sure, if file has 11 fragments, it will fail, or it will read only 10 fragments? Cause this potentially can read to the corruption, if we read the file partially.

rickgaiser:
If a file has more than 10 fragments, for instance 20, then this function will return the value 20, and a list with the first 10 fragments.

It's up to the application to decide what to do with this information. In the OPL demo I made I check for:

```
if (fragcount > 10) {
  popup("file too fragmented");
  return -1;
}
```

PR #274, opened 2022-06-09
Move system clock to TIM2

davidgfnet:
TIM2 is not used and, in theory, reserved for the kernel (but it never uses it). Since we only use it to measure the time, it's easy to configure it (using kmode enter/exit to get around permissions) and use it without much overhead (reads are allowed), while at the same time protecting it from the users.

This will free timer 0 and 1 to be used by homebrew users, and potentially by the timer library in PR #266

### DIFF
--- a/ee/kernel/include/timer.h
+++ b/ee/kernel/include/timer.h
@@ -39,6 +39,11 @@
 #define T3_MODE  ((volatile unsigned int *)0x10001810)
 #define T3_COMP  ((volatile unsigned int *)0x10001820)
 
+// Pointers to the kernel segment (uncached)
+#define K_T2_COUNT  ((volatile unsigned int *)0xB0001000)
+#define K_T2_MODE   ((volatile unsigned int *)0xB0001010)
+#define K_T2_COMP   ((volatile unsigned int *)0xB0001020)
+
 #define Tn_MODE(CLKS, GATE, GATS, GATM, ZRET, CUE, CMPE, OVFE, EQUF, OVFF) \
     (u32)((u32)(CLKS) | ((u32)(GATE) << 2) |                               \
           ((u32)(GATS) << 3) | ((u32)(GATM) << 4) |                        \
--- a/ee/kernel/src/timer.c
+++ b/ee/kernel/src/timer.c
@@ -16,17 +16,12 @@
 #include <timer.h>
 #include <kernel.h>
 
-#ifdef TIME_USE_T0
-#define INTC_TIM       INTC_TIM0
-#define T_COUNT        T0_COUNT
-#define T_MODE         T0_MODE
-#define T_COMP         T0_COMP
-#else
-#define INTC_TIM       INTC_TIM1
-#define T_COUNT        T1_COUNT
-#define T_MODE         T1_MODE
-#define T_COMP         T1_COMP
-#endif
+#define INTC_TIM       INTC_TIM2
+#define T_COUNT_RD     T2_COUNT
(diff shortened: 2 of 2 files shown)


sp193:
Only TIM3 is reserved for the kernel. I suppose you referred to the Sony SDK documentation? If so, then if we went back far enough to look at the documentation for earlier SDKs, I recall it only reserved TIM3. If I am not wrong, TIM2 became reserved after they added these SDK functions.

fjtrujy:
> Only TIM3 is used for the kernel's alarm functionality. I suppose you referred to the Sony SDK documentation? If so, then if we went back far enough to look at the documentation for earlier SDKs, I recall it only reserved TIM3. If I am not wrong, TIM2 became reserved after they added these SDK functions. It has been years since I looked at it, however.

That's totally true, however, the reason why he did this is that setting the TIM2 produces a TLB exception.
In #266 if you see we were using `TIM2` for our porpuses, it looks to work fine in `PCSX2` but it crashes in real hardware producing a TLB exception

sp193:
I was referring to his comment that TIM2 is unused by the kernel despite being documented to.

The code accesses the TIM2 registers within KSEG, which cannot be accessed via user mode. I am not sure why the change involves a change to the memory segment too, but I know too little now to comment on it.

fjtrujy:
> I was referring to his comment that TIM2 is unused by the kernel despite being documented to.
> 
> The code accesses the TIM2 registers within KSEG, which cannot be accessed via user mode. I am not sure why the change involves a change to the memory segment too, but I know too little now to comment on it.

Thanks for the quick reply 
Do you have any concern? Or should we merge it?

sp193:
Would #266 conflict with this? If so, is there a reason to use TIM2, when this code is specific to our SDK and could block us from using it for a copy of Sony's functionality?

davidgfnet:
Hey sp193!
Yeah exactly what you guys already discussed. Since TIM2 is not used by the kernel, and only by some later versions of the official SDK, we can safely use it.
TIM2 and 3 are write protected by the TLB (so a user cannot easily reprogram them), but still allows users to read them, making it a perfect candidate for clock (since most operations are reads, except the IRQ handler that needs to cleanup the flag).
And yeah, once we merge this, I will update #266 to use TIM0 or 1. That way we still have one free timer for other purposes.
Does that sound sensible? Thanks for checking this out BTW!

sp193:
Is #266 based on the Sony library? If so, did it use TIM2? If that was the case, then shouldn't it continue using TIM2 instead, so that we will not reinvent the wheel? By doing so, TIM2 can be known for one SDK purpose, rather than serving two different purposes - depending on the SDK which uses it.

fjtrujy:
> We could indeed go that route. However I figured that TIM2 was a better fit for a clock, while TIM0/1 are perhaps better for timers. Mainly cause reprogramming them is slightly slower and requires some odd kernel privileges.
> Since nobody has ported the TIM2 alarm system until now I assume it doesn't matter that much right? I assume the reason why Sony went for TIM2 was that it was an unused and reserved timer that was just free for them to use (since the library was added at a much later version of the SDK).
> Also we get some nice advantages: if a user reprograms TIM0/1 for their own use, they will break the time functions (including (nano)sleep and other library functions like in SDL), which is suboptimal. If they know how to reprogram the TIM2, then it means they know what they're doing :)

On top of that I just want to add that we will give the user the possibility to disable the initialization of the timers. This will be done in a separate PR that most probably I will do today.

PR #275, opened 2022-06-10
Clean around `libkernel` and `libkernel-nopatch`

fjtrujy:
## Description

This PR simplifies the "logic" around the 2 different flavors we have for the `kernel library`.

Currently, the project was generating 2 different libraries:
1. `libkernel.a`
2. `libkernel-nopatch.a`

The original purpose of the `libkernel-nopatch` was to avoid the `patched functionality` that we do during the initialization process of the binary. However, step by step in a wrong way, the "Patched concept" has been mixed with "light kernel version", avoiding some functionality (that isn't patching the kernel) for the `libkernel-nopatch.a`.

Having the 2 different libraries make the compilation process more difficult. Now we're using [standard libraries](https://github.com/ps2dev/gcc/blob/ee-v11.2.0/gcc/config/mips/ps2sdk.h#L2-L9), so if we want to use the `libkernel-nopatch.a` forced us to do the `-nodefaultlibs`

I have changed totally the approach used for leading with the normal, no-patch, and light conditions, just generating a single `libkernel.a`

As I said before there are functions set during the initialization process of every single binary, implying unfortunately having a bigger size binary.

Let's first take a look at the current `_InitSys` method:

```c
void _InitSys(void)
{
    StartTimerSystemTime();
    InitAlarm();
    InitThread();
    InitExecPS2();
    InitTLBFunctions();
}
```

The nature of the functions called can be split into 2 different sections:

1. Patched functionality
```c
    InitAlarm();
    InitThread();
    InitExecPS2();
    InitTLBFunctions();
```

2. Additional timer utils
```
StartTimerSystemTime();
```

## NO PATCHED BINARIES AND/OR TINY BINARIES

On the `kernel.h` file at the very end, we have a section where we defined some `macros` for making it super easy the disable these functions.

Let's put an example of how to disable the patched functionality:

```c
.....
#include <kernel.h>
#include <stdio.h>

// With this line we disabled the patched functionalities
DISABLE_PATCHED_FUNCTIONS();

int main(int argc, char **argv)
{
    printf("Hello world!\n");
    return 0;
}
```

Let's put another example where we disable the extra timer functionality:

```c
.....
#include <kernel.h>
#include <stdio.h>

// With this line we disabled the extra functionalities for timers
DISABLE_EXTRA_TIMERS_FUNCTIONS();

int main(int argc, char **argv)
{
    printf("Hello world!\n");
    return 0;
}
```

Therefore we can combine both:

```c
.....
#include <kernel.h>
#include <stdio.h>

// With this line we disabled the patched functionalities
DISABLE_PATCHED_FUNCTIONS();
// With this line we disabled the extra functionalities for timers
DISABLE_EXTRA_TIMERS_FUNCTIONS();

int main(int argc, char **argv)
{
    printf("Hello world!\n");
    return 0;
}
```

## HOW IS THIS WORKING?

It honestly is quite easy. 
All these functions that we may want to disable are defined as `__attribute__((weak))`, so it means that if a "strong" implementation is defined, they will override.
For this reason, the macros defined in the `kernel.h` are helpers that either does dummy implementations or bypass the patched wrappers.


fjtrujy:
Could be nice if @sp193 can take a look, because If I'm not mistaken he was the one that created the `libkernel-nopatch.a` library

rickgaiser (approved):
I like this new approach.

Please note that every application using libkernel-nopatch will break with this change. So ps2link and opl's eecore for instance.

Will this also affect the size of these ELF's?

fjtrujy:
> I like this new approach.
> 
> Please note that every application using libkernel-nopatch will break with this change. So ps2link and opl's eecore for instance.
> 
> Will this also affect the size of these ELF's?

Well the size of binary could be improved if you disable some of the functions you don’t need.

And you’re right binaries using `no-patch` kernel now need to be updated.

uyjulian:
Maybe keep the `nopatch` variant during the transition period, then applications can be updated as needed

uyjulian (approved):
lgtm

sp193:
I don't know much about this technique as it is specific to GCC. I created two versions of libkernel because I could not think of a better way. You should do testing to ensure that it works as intended: the symbols are available in both modes and the unwanted functions do not appear where unexpected.

The TLB patches add TLB functions. So I think you should treat the new alarm functionality the same as how you would treat the TLB functions. These were added by Sony later on, to their SDK.

fjtrujy:
> I don't know much about this technique as it is specific to GCC. I created two versions of libkernel because I could not think of a better way. You should do testing to ensure that it works as intended: the symbols are available in both modes and the unwanted functions do not appear where unexpected.
> 
> The TLB patches add TLB functions. So I think you should treat the new alarm functionality the same as how you would treat the TLB functions. These were added by Sony later on, to their SDK.

Thanks for you reply, yes this technique works fine this is something that newlib does quite often, and also is something that we have done in PSP already.

Anyway I checked the assembly code generated and the results were the expected ones.
Cheers

PR #276, opened 2022-06-10
Improve the nanosleep functionality

fjtrujy:
Now that we have real timers in the `kernel` and we can do `ThreadWaitClock` let's update our tricky `nanosleep` implementation.

### DIFF
--- a/ee/kernel/include/timer_alarm.h
+++ b/ee/kernel/include/timer_alarm.h
@@ -65,12 +65,13 @@ static inline u64 Sec2TimerBusClock(u64 sec) {
 }
 
 static inline void TimerBusClock2USec (u64 ulClock, u32 *pSec, u32 *pUsec) {
-    *pSec = ulClock/kBUSCLK;
-    *pUsec = ulClock - (*pSec * kBUSCLK);
+    u64 sec = ulClock/kBUSCLK;
+    *pSec = (u32)sec;
+    *pUsec = ulClock - (sec * kBUSCLK);
 }
 
 static inline u64 TimerUSec2BusClock (u32 sec, u32 usec) {
-    return USec2TimerBusClock(sec * 1000000 + usec);
+    return Sec2TimerBusClock(sec) + USec2TimerBusClock(usec);
 }
 
 #endif
--- a/ee/kernel/include/timer_alarm.h
+++ b/ee/kernel/include/timer_alarm.h
@@ -46,22 +46,31 @@ void StopTimerAlarm(struct timer_alarm_t *alarm);
 void ThreadWaitClock(u64 clock_cycles);
 
 // Conversion functions, from time to bus cycles
-static inline u64 NSecToCycles(u64 usec) {
+static inline u64 NSec2TimerBusClock(u64 usec) {
     // Approximate, avoid 64 bit division
     return ((((kBUSCLK / 1000) * 65536ULL) / 1000000) * usec) >> 16;
 }
 
-static inline u64 USecToCycles(u64 usec) {
+static inline u64 USec2TimerBusClock(u64 usec) {
     // Approximate, avoid 64 bit division
     return ((((kBUSCLK / 1000) * 1024) / 1000) * usec) >> 10;
 }
 
-static inline u64 MSecToCycles(u64 msec) {
+static inline u64 MSec2TimerBusClock(u64 msec) {
     return (kBUSCLK / 1000) * msec;
 }
 
-static inline u64 SecToCycles(u64 sec) {
+static inline u64 Sec2TimerBusClock(u64 sec) {
     return kBUSCLK * sec;
+}
+
+static inline void TimerBusClock2USec (u64 ulClock, u32 *pSec, u32 *pUsec) {
+    *pSec = ulClock/kBUSCLK;
+    *pUsec = ulClock - (*pSec * kBUSCLK);
+}
+
+static inline u64 TimerUSec2BusClock (u32 sec, u32 usec) {
+    return USec2TimerBusClock(sec * 1000000 + usec);
 }
 
 #endif
--- a/ee/kernel/src/timer_alarm.c
+++ b/ee/kernel/src/timer_alarm.c
@@ -91,7 +91,7 @@ static int timOverflow(int ca) {
 
     // Check if we actually triggered an alarm
     if (event_queue) {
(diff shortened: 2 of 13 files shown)


ee/kernel/src/timer_alarm.c
```diff
@@ -89,7 +89,7 @@ static int timOverflow(int ca) {
 
     // Check if we actually triggered an alarm
     if (event_queue) {
-        if (ps2_clock() > (event_queue->scheduled_time >> 8)) {
+        if (GetTimerSystemTime() > (event_queue->scheduled_time)) {
```
rickgaiser, line 92: Have all references to the legacy ps2_clock been replaced by the new GetTimerSystemTime?

ee/libc/src/sleep.c
```diff
@@ -14,83 +14,23 @@
...
-            // Correct the HSyncs for next time
-            iPS2ClockPerHSync += (clock_real_end - clock_end) / (hsync_delay-1);
-        }
-    }
-
-    // Delay whatever time is left
-    while(clock_end > ps2_clock())
-        ;
+    nano_secs = timespec_to_nanosecs(req);
+    cycles = NSecToCycles(nano_secs);
```
rickgaiser, line 32: Have the official TimerUSec2BusClock / TimerBusClock2USec functions also been implemented?

rickgaiser (approved):
It's smaller and it works better, win-win :).

ee/libc/src/sleep.c
davidgfnet: Gave them a different name. We can rename them if you prefer.
Also the cycles to time functions are not implemented. I didnt need to use them, but perhaps they do have a use for some users?

ee/kernel/src/timer_alarm.c
fjtrujy: Now, yes (at least in `ps2sdk`)

ee/libc/src/sleep.c
fjtrujy: I have added a new commit where I have implemented them, and also I have changed the name of the other to be "closer" to the PS2 official ones

AKuHAK:
Samples failing

fjtrujy:
> Samples failing

Fixed

fjtrujy:
@rickgaiser I have created a specific `nanosleep` test.
I have fixed as well the `_times` function, if we take a look to the [`newlib`](https://github.com/ps2dev/newlib/blob/ee-v4.1.0/newlib/libc/time/clock.c#L59-L60) implementation we have:

```c
clock_t 
clock ()
{
  struct tms tim_s;
  clock_t res;

  if ((res = (clock_t) _times_r (_REENT, &tim_s)) != (clock_t) -1)
    res = (clock_t) (tim_s.tms_utime + tim_s.tms_stime +
		     tim_s.tms_cutime + tim_s.tms_cstime);

  return res;
}
```

Our implementation was:

```c
clock_t _times(struct tms *buffer) {
	clock_t clk = GetTimerSystemTime() / (kBUSCLK / CLOCKS_PER_SEC);

	if (buffer != NULL) {
		buffer->tms_utime  = clk;
		buffer->tms_stime  = 0;
		buffer->tms_cutime = clk;
		buffer->tms_cstime = 0;
	}

	return clk;
}
```

So if you see, we were adding `clk` twice.

Another thing, this time to @davidgfnet .
The `nanosleep` test, in theory, should wait a total of 55 seconds, however, it is waiting 55,004 seconds.

```
Starting NANOSLEEP TESTS...
Waiting 1 second...
Waiting 2 second...
Waiting 3 second...
Waiting 4 second...
Waiting 5 second...
Waiting 6 second...
Waiting 7 second...
Waiting 8 second...
Waiting 9 second...
Waiting 10 second...
Checking if we have waited 55 seconds...
We have waited: 55004 milliseconds
Nanoseconds looks to works properly
```

I suppose that this variation is for the operation that happens in the middle, like the `printf`s, because if we disable these printfs we have:

```
Starting NANOSLEEP TESTS...
Checking if we have waited 55 seconds...
We have waited: 55000 milliseconds
Nanosecond looks to work properly
```

Thanks

ee/kernel/include/timer_alarm.h
```diff
@@ -46,23 +46,32 @@ void StopTimerAlarm(struct timer_alarm_t *alarm);
...
+static inline u64 Sec2TimerBusClock(u64 sec) {
     return kBUSCLK * sec;
 }
 
+static inline void TimerBusClock2USec (u64 ulClock, u32 *pSec, u32 *pUsec) {
+    *pSec = ulClock/kBUSCLK;
+    *pUsec = ulClock - (*pSec * kBUSCLK);
+}
+
+static inline u64 TimerUSec2BusClock (u32 sec, u32 usec) {
```
davidgfnet, line 72: This one could overflow with any time bigger than 70 minutes or so. I think it's ok really, but if we think it's not we can simply do:
```
static inline u64 TimerUSec2BusClock (u32 sec, u32 usec) {
    return Sec2TimerBusClock(sec) + USec2TimerBusClock(usec);
}
```

ee/kernel/include/timer_alarm.h
```diff
@@ -46,23 +46,32 @@ void StopTimerAlarm(struct timer_alarm_t *alarm);
...
 }
 
-static inline u64 SecToCycles(u64 sec) {
+static inline u64 Sec2TimerBusClock(u64 sec) {
     return kBUSCLK * sec;
 }
 
+static inline void TimerBusClock2USec (u64 ulClock, u32 *pSec, u32 *pUsec) {
+    *pSec = ulClock/kBUSCLK;
+    *pUsec = ulClock - (*pSec * kBUSCLK);
```
davidgfnet, line 69: I'd write it like:
```
u64 sec = ulClock/kBUSCLK;
*pSec = (u32)sec;
*pUsec = ulClock - (sec * kBUSCLK);
```
Since I suspect the psec*BUS will overflow the 32 bit int (that's only 30 seconds).

ee/kernel/include/timer_alarm.h
fjtrujy: Yeah, thanks

ee/kernel/include/timer_alarm.h
fjtrujy: Thanks, done!

PR #278, opened 2022-06-12
Add UDNL

uyjulian:
The initial source file is unchanged from OPL.  

This is mainly to allow facilitating sharing code from UDNL with the `romdrv` and `loadcore` implementations in a future PR.


iop/system/udnl-t300/src/IOPRP_img.c
```diff
@@ -0,0 +1,8 @@
+#ifndef __IOPRP_img__
+#define __IOPRP_img__
+
+unsigned int size_IOPRP_img = 0;
+unsigned char IOPRP_img[] __attribute__((aligned(16))) = {
+};
+
+#endif
```
rickgaiser, line 8: In OPL this file is generated from IOPRP.img using extra makefile rules?

iop/system/udnl/src/IOPRP_img.c
```diff
@@ -0,0 +1,8 @@
+#ifndef __IOPRP_img__
+#define __IOPRP_img__
+
+unsigned int size_IOPRP_img = 0;
+unsigned char IOPRP_img[] __attribute__((aligned(16))) = {
+};
+
+#endif
```
rickgaiser, line 8: In OPL this file is generated from IOPRP.img using extra makefile rules?

rickgaiser (commented):
Both modules look largely the same. Can these 2 sources be merged into 1? Then using a define we can build the 1 version or the other version. Much like we build libkernel and libkernel-nopatch?
This should reduce the number of UDNL modules we need to support from 2 to 1 (if possible).

These modules are mostly copied from OPL. Is it the intention to remove them there?

iop/system/udnl-t300/src/IOPRP_img.c
uyjulian: The functionality was made optional, so this file was removed.

uyjulian:
I merged the 2 modules into 1. I also made changes to remove the empty generated IOPRP\_img file.

The intention is to remove the modules from OPL and use the modules in ps2sdk instead.

uyjulian:
Sure, you can merge now

PR #283, opened 2022-06-15
Fix nanosleep implementation

fjtrujy:


uyjulian (approved):
lgtm

1000 nanoseconds = 1 microsecond

PR #288, opened 2022-06-19
IOP sysclib implementation

uyjulian:
Add sysclib implementation and the buildsystem support needed for intermediate directories.  

There are 3 different configurations built:  
1. sysclib -> The recommended version. Has the accelerated assembly versions of functions.  
2. sysclib-nano -> For memory savings. Smaller than the original SCE version.  
3. sysclib-full -> Based on sysclib, but its format series of functions supports long long and printf. Around double the size.


rickgaiser:
Hello @uyjulian what a lot of work!

So after this PR we have 4 libc libraries to choose from. Can you explain in what scenario's the default libc is not good enough, and we would want to use a different libc?

The 4 versions, correct me if I'm wrong:
- sysclib bios, what we're always used
- sysclib ps2sdk, the new default? same size binary? how well is it tested?
- sysclib-nano ps2sdk, when we want to save space, so this is for OPL only I think? How much does it save, and is the interface still compatible with the bios sysclib?
- sysclib-full, the new default? Most of the time we have plenty of space in the IOP (only OPL ingame is different)

I think we can pull this, becouse it's just extra libraries to choose from (will break nothing). Adding a readme would be nice so developers have some information on what version they can choose, and why. Without a readme these 4 versions will lead to many questions in the future.

sp193:
Normally, the IOP modules can be replaced and are replaced during boot by games. It is why our ROM is not really a BIOS, but contains modules that support the game.

Our Homebrew SDK was just shaped to run with the modules from ROM, thus being based on and equivalent to the SDKs (~1.3.4) used for the early OSD. The early OSD had no IOP reboot, perhaps because there were no updates to apply. The launch titles had an update for SDK 1.4, which replaced only minimal modules.

It looks like uyjulian wishes to add open-sourced copies of these modules, with some customizations.

To make such a module (that is already part of the boot ROM's IOPBTCONF) useful, it has to be built into an IOPRP image and the IOP should be reboot with it, at the application's startup. Otherwise, it would be just useful as a source code reference or for use in advanced projects that can use such a module.

uyjulian:
I have made some improvements to the readme so it describes the module better.

PR #290, opened 2022-06-20
modload: make name of export 15 more accurately describe the function

uyjulian:


rickgaiser:
CheckDevicePathValid -> IsIllegalBootDevice ?

What's a **Boot** Device? The reason for this change is becouse the return value is the opposite of what would be expected? What is the return value?

CheckDevicePathValid: 0=invalid, 1=valid ?
IsIllegalBootDevice: 0=legal, 1=illegal ?

uyjulian:
A illegal device (one which gets rejected by this function) is one which name starts with the following:
* mc
* hd
* net
* dev

The return result is 1->illegal, 0->legal

PR #293, opened 2022-06-21
usbd: add MINI_DRIVER version that uses 40+ KB less IOP RAM

rickgaiser:
For OPL's ingame it is important to use as little RAM as possible to preserve game compatibility. This PR adds the creation of a usbd_mini.irx (from the same sources, no copies), that uses over 40KiB less RAM.

Inspired by the ingame usbd module from @sp193 's ps2esdl.


fjtrujy:
LOL, so easy to have a stackoverflow xD

PR #295, opened 2022-06-27
Merge smap from ps2eth

uyjulian:
Merge the `smap` sources from the `ps2eth` repository to reduce code duplication.  

Mainly for backwards compatibility.

### DIFF
--- a/iop/network/smap-ps2ip/Makefile
+++ b/iop/network/smap-ps2ip/Makefile
@@ -6,12 +6,12 @@
 # Licenced under Academic Free License version 2.0
 # Review ps2sdk README & LICENSE files for further details.
 
-SUBDIRS = \
-	netman \
-	smap \
-	smap-ps2ip \
-	smbman \
-	udptty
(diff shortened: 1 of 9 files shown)


rickgaiser:
This PR will make ps2eth completely obsolete?

What needs to be done after this PR to use this smap driver instead of the ps2eth driver?
Used by ps2link only? CI changes?

uyjulian:
Yes, for using any smap driver, ps2eth would be obsolete. As far as I know, there are no users of the USB KLSI driver.  

In order to use this smap driver when using the `ps2ip` interface, simply change the module to use `smap-ps2ip.irx` from ps2sdk instead of `smap.irx` from ps2eth.  
ps2link no longer uses ps2eth. wLaunchELF is currently not using ps2eth, but that is broken at the moment so I plan to make it use the version in this PR instead.  
Projects that are using ps2eth in CI are manually cloning and building it. It is not included in `ps2dev` docker image. So there are no changes to CI needed.

PR #299, opened 2022-06-28
Add simple BDM sector cache

rickgaiser:
This PR adds a simple 32KiB sector cache to all BDM devices.

Algorithm:
- All reads smaller than 4K will go through the cache (so 512b, 1K, 2k)
- All reads of 4K and larger will **not** use the cache, they are already fast, and only waste cache space.
- Cache will always read in 4K blocks from device (so always more than the requested 512b, 1K, 2k) becouse it is faster
- Cache reads start at the requested sector (so not aligned to 4K)
- Cache has 8 blocks of 4K = 32K
- LRU algorithm
- Read only cache (writes will invalidate the written sectors)

Possible improvements:
- Dynamic cache size (adding an extra parameter to bd_cache_create)
- Better algorithm

Note for OPL:
- This improves GUI loading speeds for all BDM devices
- This **does not** improve ingame speed
- This cache can potentially be used ingame, but that would require a dynamic cache size (see improvements)

Comparison to vfat cache (ccache):
- Has the same 4KiB block size
- Has 32 blocks, instead of 8 (also uses 128KiB instead of 32KiB)
- Has better write support

### DIFF
--- a/iop/fs/Makefile
+++ b/iop/fs/Makefile
@@ -6,7 +6,7 @@
 # Licenced under Academic Free License version 2.0
 # Review ps2sdk README & LICENSE files for further details.
 
-SUBDIRS = bdm libbdm bdmfs_vfat bdmfs_fatfs devfs fakehost filexio http netfs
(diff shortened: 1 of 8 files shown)


uyjulian:
Looks good

PR #302, opened 2022-07-01
libbdm: improve caching by a LOT

rickgaiser:
I wasn't happy with the performance of the 8 block LRU (Least Recently Used) cache. So I've been looking into different algorighms that would fit our usecase:

**LRU** - Least Recently Used
With LRU the least recently used blocks will be evicted. But the problem with LRU is that many blocks that are only used once can quickly flush out the (32 block) cache.

**LFU** - Least Frequently Used
With LFU the least frequently used blocks will be evicted. This solves the LRU problem, but creates the problem of cache pollution. Where for instance 31 blocks of the 32 block cache have high usage counts, and so only 1 block can be used for all new accesses. Highly used blocks in the past will stay present in the cache forever.

**LFUDA** - Least Frequently Used with Dynamic Aging
To overcome the issue of LFU, blocks must somehow "age" in time. So only the blocks that have been used frequently AND recently are kept in the cache.

This PR changes the algorithm to my personal implementation of LFUDA. Tests with browsing through game lists in OPL show an almost 20x improvement over the old algorithm. So where the old cache needed to read 20MB from the device for this certain test, the new cache only needs to load about 1MB for the same test.
Note that the cache still ONLY caches reads that are SMALLER than 4KiB. So the 20x improvement is ONLY for those reads. Large block reads are not cached, and cannot be cached. They where already fast and don't need much omprovement.

### DIFF
--- a/iop/fs/libbdm/src/bd_cache.c
+++ b/iop/fs/libbdm/src/bd_cache.c
@@ -5,21 +5,20 @@
 //#define DEBUG  //comment out this line when not debugging
 #include "module_debug.h"
 
-
-#define SECTORS_PER_BLOCK 8 // 8 * 512b =  4KiB
-#define BLOCK_COUNT       8 // 8 * 4KiB = 32KiB
-
+#define SECTORS_PER_BLOCK     8 //  8 * 512b =   4KiB
+#define BLOCK_COUNT          32 // 32 * 4KiB = 128KiB
+#define BLOCK_WEIGHT_FACTOR 256 // Fixed point math (24.8)
 
 struct bd_cache
 {
     struct block_device *bd;
-    u32 lru_current;
-    u32 lru[BLOCK_COUNT];
+    int weight[BLOCK_COUNT];
     u32 sector[BLOCK_COUNT];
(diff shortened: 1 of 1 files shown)


iop/fs/libbdm/src/bd_cache.c
```diff
@@ -57,56 +56,71 @@ static int _read(struct block_device *bd, u32 sector, void *buffer, u16 count)
...
-        memcpy(buffer, &c->cache[blkidx_best][offset], count * 512);
-        return count;
+        if (c->weight[blkidx] < blkidx_best_weight) {
+            // Better block found
+            blkidx_best_weight = c->weight[blkidx];
+            blkidx_best = blkidx;
+        }
     }
+#ifdef DEBUG
+    printf(" devread: %*d, evict %*d [%*d], add [%*d] %d\n", 4, c->sectors_dev, 2, blkidx_best, 8, c->sector[blkidx_best], 8, sector);
```
uyjulian, line 110: There is an extra `%d` here without a matching argument

rickgaiser: Ah yes, my last minute cleanup before push. Force pushing the fix.

PR #303, opened 2022-07-04
More libcdvd definitions and documentation

uyjulian:


AKuHAK:
Does led and fan control also have minimum mecha version?

uyjulian:
> Does led and fan control also have minimum mecha version?

Yes

PR #318, opened 2022-08-07
mx4sio fixed

### LINKED ISSUE #317: MCMAN dynamic linking with SIO2MAN causing issues with MX4SIO
MCMAN uses dynamic linking with SIO2MAN here:
https://github.com/ps2dev/ps2sdk/blob/3df67b619704328e1093ca01a729f09f542408f6/iop/memorycard/mcman/src/main.c#L171-L240

For MX4SIO to work it needs to "hook" into these functions, as it does here:
https://github.com/ps2dev/ps2sdk/blob/3df67b619704328e1093ca01a729f09f542408f6/iop/sio/mx4sio_bd/src/sio2man_hook.c#L131-L151

When MX4SIO and MCMAN try to use SIO2MAN at the same time this will cause issues. As is the case with OPL here:
https://github.com/ps2homebrew/Open-PS2-Loader/issues/663

Is there a reason for MCMAN in ps2sdk to have this dynamic linking?
Can we simplify MCMAN to use a standard import table? Then import SIO2MAN / XSIO2MAN / RSIO2MAN
Also what does the "xfromman.irx" compile option in MCMAN do?

rickgaiser:
These fixes will make mx4sio work in OPL again.

@uyjulian can you check the mcman changes? I'm not sure why this was coded the way it was? Also see https://github.com/ps2dev/ps2sdk/issues/317

closes #317


uyjulian:
These changes look to be OK.

I don't think any program will be using ps2sdk's mcman with ROM sio2man.

PR #322, opened 2022-08-14
Implement truncate

AJenbo:
Implementation taken from pspsdk. This function is required for proper save game handling in DevilutionX (Diablo 1 port)


fjtrujy:
Thanks for doing it.
I probably implemented it in `PSPSDK` 😄 .
By the way, you have missed adding that new function in the `makefile`
https://github.com/ps2dev/ps2sdk/blob/master/ee/libcglue/Makefile#L23

AJenbo:
It was already in the make file, just missing the implementation :)

PR #324, opened 2022-08-15
MCMAN and MCSERV old API in separate module variants

uyjulian:
Use separate module variants for older API interface of MCMAN and MCSERV.


rickgaiser:
Thanks, I wanted to do something similar since my PR #318 broke wLE.
Merging.

Issue #332, opened 2022-08-21
`printf` not working fine with `fileXio`


fjtrujy:
When we use `fileXio` the `printf` messages stop working when running `PCSX2`.

When using `ps2client` is still working because the `udptty` driver is closing and opening again the `fd` `0` and `1`, as you can see here, for redirecting the output:
https://github.com/ps2dev/ps2sdk/blob/master/iop/network/udptty/src/udptty.c#L200-L208

I suppose that something similar should be done in `fileXio`.
Are the `printf` messages with `fileXio` working in `DevTool machines`? (@uyjulian could you double-check it)?

The example:

```c
int main(int argc, char **argv) {
  SifExecModuleBuffer(&iomanX_irx, size_iomanX_irx, 0, NULL, NULL);	
  SifExecModuleBuffer(&fileXio_irx, size_fileXio_irx, 0, NULL, NULL);
  fileXioInit();

  while(1) {
    printf("Hello world with fileXio drivers!\n");
  }
}
```

uyjulian:
Last time I checked, `printf` messages with `fileXio` are working with DTL-T machines.  

Once I reverse the `STDIO` and `IOMAN` module, I'll probably have a better understanding of how this all integrates and how to resolve this.

fjtrujy:
> Last time I checked, `printf` messages with `fileXio` are working with DTL-T machines.
> 
> Once I reverse the `STDIO` and `IOMAN` module, I'll probably have a better understanding of how this all integrates and how to resolve this.

Was it before or after the newlib integration, #161 ?

uyjulian:
> > Last time I checked, `printf` messages with `fileXio` are working with DTL-T machines.
> > Once I reverse the `STDIO` and `IOMAN` module, I'll probably have a better understanding of how this all integrates and how to resolve this.
> 
> Was it before or after the newlib integration, #161 ?

After

rickgaiser:
These printf messages that magically appear in PCSX2, do you know how they get there?

I'm assuming they are IOP SIO writes, not EE SIO? And I'm assuming ioman somehow uses fd=0 and fd=1 to write to the IOP serial port. Can you confirm the PCSX2 output you see is the IOP serial port output? And if not, do you know what it is?

Searching in the ps2sdk sources I can find no driver for using the iop serial port.
We could probably make an iop serial port driver (using ps1 code as example), then do the same thing we do with udptty:
[...]

PR #333, opened 2022-08-21
Write also to `sio_write` meanwhile issue with fileXio is fixed

fjtrujy:
This is a temporal solution meanwhile we find something better for #332 , at least in this way we still can use `PCSX2` for development purposes.

Cheers

### DIFF
--- a/ee/libcglue/src/glue.c
+++ b/ee/libcglue/src/glue.c
@@ -312,8 +312,9 @@ int (*_ps2sdk_write)(int, const void*, int) = fioWrite;
 
 int _write(int fd, const void *buf, size_t nbytes) {
 	// HACK: stdout and strerr to serial
-	//if ((fd==1) || (fd==2))
-	//	return sio_write((void *)buf, nbytes);
+	// meanwhile https://github.com/ps2dev/ps2sdk/issues/332 is fixed
+	if (((fd==1) || (fd==2)) && (_ps2sdk_write != fioWrite))
+		sio_write((void *)buf, nbytes);
 
 	return __transform_errno(_ps2sdk_write(fd, buf, nbytes));
 }


uyjulian:
Seems OK to me. DTL-T also forwards stuff in EESIO so this should be fine for now while a better solution is found.

AKuHAK:
This means that printf will also output into ee-sio?

fjtrujy:
When using fileXio, yes

rickgaiser:
With this temporary solution you will only see the printf messages from the EE code. You will still loose the printf messages from the IOP code.

Perhaps a better temporary/hack solution would be to:
- have iomanX import the write function from ioman
- have iomanX forward the writes to fd=0 and fd=1 to the write function of ioman?

Then we still don't know what magic ioman/PCSX2 do with fd=0 and fd=1, but at least we know it still works the same way.

fjtrujy:
> With this temporary solution you will only see the printf messages from the EE code. You will still loose the printf messages from the IOP code.
> 
> Perhaps a better temporary/hack solution would be to:
> 
> * have iomanX import the write function from ioman
> * have iomanX forward the writes to fd=0 and fd=1 to the write function of ioman?
> 
> Then we still don't know what magic ioman/PCSX2 do with fd=0 and fd=1, but at least we know it still works the same way.

I think that messages from IOP they weren’t lost, they are still visible, anyway I will double check it

fjtrujy:
I'm closing this PR because it wasn't in the end an issue in the `PS2SDK`, it was something wrong in `PCSX2`.
This PR has fixed the issue.
https://github.com/PCSX2/pcsx2/pull/7007

israpps:
Would be nice to have a solution for EE SIO on newer Toolchain that doesn't require devs fooling around with `fopencookie()`

uyjulian:
> Would be nice to have a solution for EE SIO on newer Toolchain that doesn't require devs fooling around with `fopencookie()`

This will be dealt with once EE-side device is implemented

PR #334, opened 2022-08-24
Adding support for `std:thread` and `pthread`

fjtrujy:
This PR is part of a collection for the `std::tread` support in the PS2 EE Toolchain

This PR follows a similar approach as `newlib` in the implementation plan.

As a requirement we MUST have `ps2toolchain-ee` layer totally agnostic to `ps2sdk`, we should be able to compile `psptoolchain-ee` without knowing anything about `ps2sdk` yet.
So in order to achieve this with `pthread-embedded` what I have done is to remove from the `pthread-embedded` repository all the `ps2` specific code, and it is now under the `ps2sdk/ee/libpthreadglue`

So basically :

`pthread-embbeded` generates `libpthread.a`, which calls some functions that now are within `ps2sdk/ee/libpthreadglue`,
`ps2sdk/ee/libpthreadglue` generates `libpthreadglue.a` which uses PS2 kernel functions to deal with threading.
Why `pthread-embedded` is going to be in `ps2toolchain-ee` instead of being as `port` and allocate in `ps2sdk-port` as we are doing with other libraries?

As you probably know, in the toolchain we have 2 phases for `GCC`. The first phase allows us to use C code, and the second phase is for providing `CPP support`.
In order to make phase 2 to work, we used to require to have `newlib` support, now If we want to support `std::thread` would mean to have `pthread-embbeded` compiled in advance.
This is why `pthread-embbeded` is going to be in the toolchain and before gcc phase 2.

The majority of the changes that you see have been basically taken from the `libpthreadglue` from the  `PSPSDK` repo, I just took the PSP implementation and do the required adjustment to make it work for PS2 (As I did it months ago for the PSP).

## ⚠️ IMPORTANT:
Now `libpthread.a` and `libpthreadglue.a` is going to be part of the standard libraries, it means that you won't even need to add anything for using them, is going to be out of the box.
It means, that binary size will increase automatically around 15KB, however, I'm offering a way to disable the pthread support if you aren't going to use it.
You just need to put in your code the next macro (usually in the main.c file):

```c
PS2_DISABLE_AUTOSTART_PTHREAD();
```
That macro is defined in the `ps2sdkapi.h` file (I didn't find a better way of doing it pinging @rickgaiser and @uyjulian for bring some ideas)

That's actually needed because during the initialization process we need to call explicitly the function `pthread_init()`, otherwise `pthread` and obviously `std::thread` won't work.

This PR, requires to merge first https://github.com/ps2dev/gcc/pull/2 and https://github.com/ps2dev/ps2toolchain-ee/pull/22


rickgaiser:
There's still many references to "psp" in this PR.
I think these should be changed to "ps2" or some platform independent name.

fjtrujy:
> There's still many references to "psp" in this PR. I think these should be changed to "ps2" or some platform independent name.

Done!

AKuHAK:
Samples are failing.

PR #339, opened 2022-09-02
64 byte alignment changes

uyjulian:


ee/rpc/sior/src/sior_rpc.c
```diff
@@ -26,7 +26,7 @@
 
 static SifRpcDataQueue_t qd __attribute__((aligned(64)));
 static SifRpcServerData_t Sd0 __attribute__((aligned(64)));
-static u32 buffer[32] __attribute__((aligned(64)));
+static u32 buffer[64] __attribute__((__aligned__(64)));
```
rickgaiser, line 29: 128 byte to 256 byte ?

iop/fs/fakehost/src/nprintf.c
```diff
@@ -46,7 +46,7 @@ static void *naplinkRpcHandler(int cmd, void *buffer, int size)
 ////////////////////////////////////////////////////////////////////////
 static SifRpcServerData_t server __attribute((aligned(16)));
 static SifRpcDataQueue_t  queue __attribute((aligned(16)));
-static unsigned char rpc_buffer[512] __attribute((aligned(16)));
+static unsigned char rpc_buffer[512] __attribute__((__aligned__(64)));
```
rickgaiser, line 49: For the IOP 4 byte alignment would be enough.

rickgaiser (changes requested):
Please explain why you think this is needed.

On the IOP side, DMA -> EE restrictions are:
- 4 byte alignment
- multiples of 16bytes (the EE DMA minimum)

On the EE side, DMA restrictions are:
- 16 byte alignment
- multiples of 16 byte

But on the EE, there's also cache-line restrictions, the easy way around this is to use:
- 64 byte alignment
- multiples of 64 byte

sp193:
To deal with the EE cache, it is possible to bypass cached memory by using one of the uncached segments too. In that case, it is not necessary to have 64-byte alignment for the buffer in EE memory. However, mixing strategies (writing/invalidating the cache & bypassing the cache) is a recipe for disaster.

rickgaiser:
So the most simple strategy I think would be:
On the IOP: **4 byte** alignment, **16 byte** multiples
On the EE: **64 byte** alignment, **64 byte** multiples for allocation, **16 byte** multiples for transmission

Using anything larger than this will not give any improvements and only cost extra RAM.

ee/rpc/sior/src/sior_rpc.c
uyjulian: Ah whoops, this was an unintended change.

uyjulian:
I changed it so that it will be aligned to 16 bytes on IOP and 64 bytes on EE.

iop/network/netman/src/rpc_server.c
```diff
@@ -17,7 +17,7 @@
 
 /* Data used for registering the RPC server. */
 static SifRpcServerData_t rpc_sdata;
-static unsigned char rpc_buffer[80];
+static unsigned char rpc_buffer[80] __attribute__((__aligned__(4)));
```
rickgaiser, line 20: 80 is not a multiple of 16.

rickgaiser:
Looks good, merging. Thanks.

PR #342, opened 2022-09-20
Add alignement to iox_stat_t and iox_dirent_t

fjtrujy:
I was suffering some issues with `ps2link/ps2client` when listing directories and file stats.
The error was just happening with `fileXio` however with `fio` it worked properly.

After hours of investigation, we faced (with the help of @F0bes) that it was an alignment issue. Something really similar to the error in `wLaunchELF` https://github.com/ps2homebrew/wLaunchELF/pull/89/files

I have tested it in real hardware and in `PCSX2`, in both places, it is working as expected.

Cheers

### DIFF
--- a/common/include/iox_stat.h
+++ b/common/include/iox_stat.h
@@ -106,14 +106,14 @@ typedef struct
     unsigned int private_4;
     /** Sector start.  */
     unsigned int private_5;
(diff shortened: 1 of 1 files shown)


uyjulian (approved):
lgtm

PR #343, opened 2022-09-22
font and cursor customizable colors for scr_printf

israpps:
I haven't tested it, only saw that putchar actually spits colored chars so i wanted to try


fjtrujy (approved):
LGTM

uyjulian (approved):
lgtm

PR #347, opened 2022-10-01
mcman: Always forward McDetectCard to mcman_detectcard

uyjulian:
Attempt to remove the workaround as done by the following: https://github.com/ps2homebrew/wLaunchELF/pull/88

**untested**


uyjulian:
This isn't needed.

PR #352, opened 2022-10-22
Implement SMAP modular hook interface

uyjulian:
Mainly intended for lightweight usages, like UDPBD, without needing to add its network stack code into the SMAP module itself.

### DIFF
--- a/iop/network/smap-modular-netman/Makefile
+++ b/iop/network/smap-modular-netman/Makefile
@@ -6,15 +6,13 @@
 # Licenced under Academic Free License version 2.0
 # Review ps2sdk README & LICENSE files for further details.
 
-SUBDIRS = \
-	netman \
-	smap \
-	smap-modular-netman \
-	smap-modular-none \
-	smap-none \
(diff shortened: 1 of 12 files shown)


sp193:
Isn't that the purpose of NETMAN, to provide a layer of abstraction from the network protocol stack?

uyjulian:
The consumer of this interface (UDPBD) intends to operate at a much lower level then netman.

Also, netman cannot handle multiple network stacks at once, only one. I could try to implement multiple network stack support into netman, but I'll probably do that at a later time.

sp193:
Out of curiosity, how much lower-level can it get? NETMAN provides basic functions for sending, receiving and doing some operations with the network interface.

It is true that NETMAN cannot support concurrent usage of multiple network stacks, as that was not planned because I saw no need for it.

uyjulian:
> Out of curiosity, how much lower-level can it get?

The UDPBD driver interacts directly with the SMAP hardware for performance and memory reasons.

PR #356, opened 2022-11-09
Upload containers to Github Container Registry

uyjulian:


fjtrujy:
I tried to this time ago... but I think that we need some grants from @ooPo

uyjulian:
I asked, and public package publishing is enabled, so this works now.

PR #365, opened 2022-11-26
ELF loader no debug

### LINKED ISSUE #361: [feature]: add 'silent' variant of ELF loader
this 'silent' variant should __not__ change the BG colors 

This FR purpose is mainly to make OPL Launcher maintain the screen on black color without the need of making a modified copy of this loader for the OPL Launcher repo

uyjulian:
Fixes #361

### DIFF
--- a/ee/Rules.make
+++ b/ee/Rules.make
@@ -13,7 +13,21 @@ EE_SRC_DIR ?= src/
 EE_INC_DIR ?= include/
 EE_SAMPLE_DIR ?= samples/
 
+# Helpers to make easy the use of newlib-nano
+EE_NEWLIB_NANO ?= 0
+EE_COMPACT_EXECUTABLE ?= 0
+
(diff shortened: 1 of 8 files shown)


ee/Rules.make
```diff
@@ -44,25 +54,25 @@ EE_SAMPLES := $(EE_SAMPLES:%=$(EE_SAMPLE_DIR)%)
 
 EE_OBJS := $(EE_OBJS:%=$(EE_OBJS_DIR)%)
 
+EE_LINKFILE ?= $(PS2SDKSRC)/ee/startup/src/linkfile
+
+EE_BIN_MAPFILE ?= $(shell basename $(CURDIR)).map
+
+ifneq (x$(EE_COMPACT_EXECUTABLE), x0)
+EE_LDFLAGS += -Wl,-zmax-page-size=128 -s -Wl,--gc-sections
+EE_CFLAGS += -fdata-sections -ffunction-sections
```
fjtrujy, line 63: Do we need `-Os` here as well?

uyjulian: I have it set in `EE_OPTFLAGS` in the loader Makefile, but I think it would be better to move it inside Rules.make

fjtrujy (approved):

PR #372, opened 2022-12-11
Improve colors in the ELF loader

fjtrujy:
The red color was duplicated, now we have a separate color per step

### DIFF
--- a/ee/elf-loader/src/loader/src/loader.c
+++ b/ee/elf-loader/src/loader/src/loader.c
@@ -24,6 +24,7 @@
 #define SET_GS_BGCOLOUR(colour)
 #endif
 
+// Color status helper in BGR format
 #define WHITE_BG 0xFFFFFF // start main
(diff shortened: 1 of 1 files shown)


fjtrujy:
@israpps you could create a readme in this folder where to put the color documentation

israpps:
> @israpps you could create a readme in this folder where to put the color documentation

Could be possible to make a commit (by someone more experienced than me)

Where all debug colors are documented on comments

Then make a script that reads them all and tabulates on a markdown file maybe?

Just an idea.

israpps:
Oh, now I see that unlike OPL EE core all colors are defined on macro and has small description on comments

What about using those [doxygen groupings](https://www.doxygen.nl/manual/grouping.html) and make the color a group, so descriptions are rendered together

fjtrujy:
I just suggested you in case you were interested.
Having something automatic would be really great.

PR #375, opened 2022-12-12
Implement types to modload and secrman from loadfile

uyjulian:
More useful symbol names and types.


fjtrujy (approved):
Good improvements

PR #379, opened 2022-12-17
HDD checking modules

uyjulian:
Mainly for easier reference and code sharing with the APA and PFS drivers.


AKuHAK:
Is it based on hdd checker modules, or red from official?

uyjulian:
It is based on HDDChecker modules

sp193:
I do not really recommend including these modules into the PS2SDK, as nearly no homebrew other than HDDChecker (and perhaps the fsck tool I built for FHDB) will use these modules. Otherwise, every utility IRX under the sun could be considered as being necessary to become part of the PS2SDK. So when do you draw that line?

One thing I didn't do in the past, was to build libapa and libpfs as static libraries, which can then be used as dependencies for building other IRX (e.g. HDD.IRX, PFS.IRX, FSCK.IRX etc). Would that be better?

israpps:
> I do not really recommend including these modules into the PS2SDK, as nearly no homebrew other than HDDChecker (and perhaps the fsck tool I built for FHDB) will use these modules. Otherwise, every utility IRX under the sun could be considered as being necessary to become part of the PS2SDK. So when do you draw that line?
> 
> One thing I didn't do in the past, was to build libapa and libpfs as static libraries, which can then be used as dependencies for building other IRX (e.g. HDD.IRX, PFS.IRX, FSCK.IRX etc). Would that be better?

Hi @sp193 

Sorry for bothering...

I sent you an email but i had no answer in almost 3 weeks.

Would you be so kind of disclosing the MBR loader feature of FreeMcBoot?
It will be very useful for some on-development projects!

fjtrujy:
> I do not really recommend including these modules into the PS2SDK, as nearly no homebrew other than HDDChecker (and perhaps the fsck tool I built for FHDB) will use these modules. Otherwise, every utility IRX under the sun could be considered as being necessary to become part of the PS2SDK. So when do you draw that line?
> 
> One thing I didn't do in the past, was to build libapa and libpfs as static libraries, which can then be used as dependencies for building other IRX (e.g. HDD.IRX, PFS.IRX, FSCK.IRX etc). Would that be better?

That's a really good argument....
I think that `ps2sdk` is already too complicated with the amount of `IRX` drivers that we have.
Maybe could be a good idea to create a specific repo where we group all these "additional IRX drivers"

uyjulian:
Yeah, I think at least it would be a good idea to split the filesystem modules (like `apa`, `pfs`, `mcman`, `cdfs`, `bdm_vfat` etc) off to a separate repository to at least make those easier to handle.  

Also, regarding "utility" functions in ps2sdk, I believe since the modules already include functionality of formatting a partition or filesystem, that adding checking and fixing for those filesystems is not that much bigger of a stretch.  

For another example, Linux has e.g. ext2 drivers for reading and writing the drive, but it does not have the functionality of e.g. formatting and checking the drive, which is handled by e.g. e2fsprogs.

PR #380, opened 2022-12-23
Patch, taken from kernelloader, to allow spu2 and usb to work together

uyjulian:
Patch to allow SPU2 and USB to work together on newer slims
Taken from https://gitlab.com/ps2max/linux/kernelloader/-/blob/7255483bf03e42fa3b34c4a990ffcb0e61a7b7da/loader/loader.c#L564


sp193:
Should this really be merged in? It looks like a hack to me. And I don't suppose it is applicable to us because we don't actually have this problem, do we? From what I do recall, the issue is with the implementation of DECKARD, preventing logical OR of the IOP registers from being completed properly from the EE.
Unlike normal homebrew, PS2Linux had support for some peripherals implemented as kernel modules, with I/O done from the EE.

uyjulian:
Hmm... I'll probably rework this as an option to make it optional.

kernelloader has hardcoded the find/replace code, so I think it would be a better idea to at least document what it actually does in some way.

PR #385, opened 2022-12-28
Changes to use and cache real time clock value based on offset from timer

uyjulian:
Mainly for applications that call `time` or related functions very often

### DIFF
--- a/ee/libcglue/src/glue.c
+++ b/ee/libcglue/src/glue.c
@@ -699,7 +699,7 @@ int clock_getres(clockid_t clk_id, struct timespec *res) {
 
 	/* Return the actual time since epoch */
 	res->tv_sec = tv.tv_sec;
-	res->tv_nsec = 0;
+	res->tv_nsec = tv.tv_usec * 1000;
 
 	return 0;
 }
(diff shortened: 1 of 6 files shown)


fjtrujy:
I think that this PR will require updating `default libraries` in GCC
Now, `ps2time` is not used which means that libcdvd is not part of the "default libraries", right?

fjtrujy:
Sorry, I see that this is still used...

ee/libcglue/src/glue.c
```diff
@@ -632,27 +633,30 @@ void * _sbrk(size_t incr) {
...
+int _gettimeofday(struct timeval *tv, struct timezone *tz)
+{
+	if (tv == NULL)
+	{
+		errno = EINVAL;
+		return -1;
+	}
 
-time_t ps2time(time_t *t);
+	{
```
fjtrujy, line 644: this bracket is not needed

uyjulian: This is just done to ensure that the scope for the variables enclosed is gone when the ending bracket is reached.

ee/libcglue/src/rtc.c
```diff
@@ -0,0 +1,40 @@
...
+#include <time.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ps2sdkapi.h"
+#define OSD_CONFIG_NO_LIBCDVD
+#include "osd_config.h"
+#include "timer_alarm.h"
+
+// The definition for this function is located in ee/rpc/cdvd/src/scmd.c
+extern time_t ps2time(time_t *t);
```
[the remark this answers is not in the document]
uyjulian: `extern` is optional, but I placed it there to make it explicit that it is a forward declaration.

PR #390, opened 2023-01-12
Fix alignment of variables in IOP modules to be max 16 bytes

uyjulian:
The IRX loader can only handle max alignment of 16 bytes.  
Future updates to the toolchain will make this a hard requirement.


sp193:
Do correct me if I am wrong. The alignment of DMA buffers on the IOP only have to be aligned to addresses that are multiples of 4 bytes. Even if we considered how the SIF DMA channels work for the EE, that would be 16 bytes (128 bits).

PR #391, opened 2023-01-13
Move `crt0` from `newlib` to `ps2sdk`

fjtrujy:
Currently `crt0` is in `newlib` with this PR together with https://github.com/ps2dev/newlib/pull/5 we will have crt0 as `.c` and in `ps2sdk` which will make easier the usage/improvements and development.

Cheers


AKuHAK:
@fjtrujy @rickgaiser this PR broke ELF loading support in opl:
https://github.com/ps2homebrew/Open-PS2-Loader/actions/runs/4845852591/jobs/8635017026
this one is still working
https://github.com/ps2homebrew/Open-PS2-Loader/actions/runs/4935525224/jobs/8821937135
isnt working anymore 9that was triggered by this pr merging)

related issue [here](https://github.com/ps2homebrew/Open-PS2-Loader/issues/962)

AKuHAK:
wLaunchELF become broken as well

fjtrujy:
But what is not actually working? The loading of ELFs?

AKuHAK:
> But what is not actually working? The loading of ELFs?

yes, exactly, black screen on elf loading

AKuHAK:
wLaunchELF btw was broken long time ago, so it is not really good example in this error

fjtrujy:
Do you know by chance if legacy wLE version can load new binaries generated after this PR without problem?

AKuHAK:
> Do you know by chance if legacy wLE version can load new binaries generated after this PR without problem?

yes as we can run opl without problem, elfs itself are not broken. It seems.

fjtrujy:
> > Do you know by chance if legacy wLE version can load new binaries generated after this PR without problem?
> 
> yes as we can run opl without problem, elfs itself are not broken. It seems.

I think that I have found the issue, it is basically a race condition in `elf-loader` which is using a wrong `crt0.o` file

PR #397, opened 2023-01-21
Add support for 8bit wave files in adpenc

AJenbo:
Since conversion from 8bit wave to 16bit is pretty trivial this saves us the need to use an additional tool for the conversion


fjtrujy (approved):
LGTM!
@uyjulian ?

uyjulian (approved):
Lgtm

PR #398, opened 2023-01-21
Implement positional audio in adpcm

AJenbo:
This adds a 3rd parameter to `audsrv_adpcm_set_volume()` which makes it possible to pan the sound left or right, allowing for positional audio in games.

-100: Sound is all the way to the left
-50: Sound is 45° to the left (50% volume on the right)
0: Sound is centered
50 Sound is 45° to the right (50% volume on the left)
100: Sound is all the way to the right

### DIFF
--- a/ee/rpc/audsrv/include/audsrv.h
+++ b/ee/rpc/audsrv/include/audsrv.h
@@ -221,9 +221,11 @@ int audsrv_adpcm_init();
 /** Sets output volume for the specified voice channel.
  * @param ch       Voice channel ID
  * @param vol      volume in percentage (0-100)
+ * @param pan      left/right offset [-100 .. 0 .. 100]
  * @returns 0 on success, negative otherwise
  */
-int audsrv_adpcm_set_volume(int ch, int vol);
+int audsrv_adpcm_set_volume_and_pan(int ch, int vol, int pan);
+#define audsrv_adpcm_set_volume(ch, vol) audsrv_adpcm_set_volume_and_pan(ch, vol, 0) //For backward-compatibility
 
 /** Uploads a sample to SPU2 memory
  * @param adpcm    adpcm descriptor structure
--- a/ee/rpc/audsrv/samples/playadpcm/playadpcm.c
+++ b/ee/rpc/audsrv/samples/playadpcm/playadpcm.c
@@ -70,7 +70,7 @@ int main(int argc, char *argv[])
 
 	audsrv_adpcm_init();
 	audsrv_set_volume(MAX_VOLUME);
-	audsrv_adpcm_set_volume(0, MAX_VOLUME);
+	audsrv_adpcm_set_volume_and_pan(0, MAX_VOLUME, 0);
 	audsrv_load_adpcm(&sample, buffer, size);
 	audsrv_ch_play_adpcm(0, &sample);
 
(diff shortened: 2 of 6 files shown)


ee/rpc/audsrv/include/audsrv.h
```diff
@@ -221,9 +221,10 @@ int audsrv_adpcm_init();
 /** Sets output volume for the specified voice channel.
  * @param ch       Voice channel ID
  * @param vol      volume in percentage (0-100)
+ * @param pan      left/right offset [-100 .. 0 .. 100]
  * @returns 0 on success, negative otherwise
  */
-int audsrv_adpcm_set_volume(int ch, int vol);
+int audsrv_adpcm_set_volume(int ch, int vol, int pan);
```
fjtrujy, line 227: in order to keep retro compatibility, should we create another method instead? and just call that method with the default value for the existing function?

KrahJohlito: was thinking the same.. https://github.com/ps2dev/ps2sdk/blob/master/ee/rpc/audsrv/include/audsrv.h#L247
for example

AJenbo: Done

AJenbo:
I'm unable to see what the issue is in codacy, from the looks of it I would say it's a fault on there part and not an issue with the actual PR

ee/rpc/audsrv/include/audsrv.h
```diff
@@ -221,9 +221,11 @@ int audsrv_adpcm_init();
 /** Sets output volume for the specified voice channel.
  * @param ch       Voice channel ID
  * @param vol      volume in percentage (0-100)
+ * @param pan      left/right offset [-100 .. 0 .. 100]
  * @returns 0 on success, negative otherwise
  */
-int audsrv_adpcm_set_volume(int ch, int vol);
+int audsrv_adpcm_set_volume_and_pan(int ch, int vol, int pan);
+#define audsrv_adpcm_set_volume(ch, vol) audsrv_adpcm_set_volume_and_pan(ch, vol, 0) //For backward-compatibility
```
glebm, line 228: Make this a regular function with the same signature as before?

```c
int audsrv_adpcm_set_volume(int ch, int vol);
```

AJenbo: Hmm I think it's better to stick to the same pattern that has been used with other functions that changed.

glebm: The 2 functions here have different names and the non-pan version is still a useful function, it's not just for backwards compatibility.

ee/rpc/audsrv/include/audsrv.h
fjtrujy: Welll done!

fjtrujy (approved):
This PR makes totally sense, @uyjulian waiting your feedback

ee/rpc/audsrv/include/audsrv.h
AJenbo: how so? withe pan set to zero it gives the same result as the old one

glebm: Well it's probably nice to not have to pass zero when you don't need to pan but either way is fine by me

uyjulian (approved):
Looks good to me

PR #401, opened 2023-01-28
BDM interface variant of atad

uyjulian:


rickgaiser:
This is great! I always wanted to do this :-).

So with this ATA driver ("ata_bd") we can use the internal HDD as a BDM block device. Meaning we can format it to fat32 or exfat.

Is this ready for merge?

Another thing that would be great is to change the PFS driver to use the BDM file system interface. That way we can use the PFS file system on all BDM devices too (OSDSYS on a slim with USB stick a possibility?).

AKuHAK:
@uyjulian maybe it will be useful to add some sort of version, or flag somewhere? For example:
0 - use pure APA
1 - use pure BDM
2 - use mixed scheme

PR #417, opened 2023-03-25
Tweak IOP modules printf

israpps:
- enforce modname as a prefix on printf's of some modules
- encase some printf inside macros, so they're not added when building non debug (left all those "Welcome" printf intact
- unify printf macros into single header for certain modules

### DIFF
--- a/iop/cdvd/cdfs/src/cdfs_iop.h
+++ b/iop/cdvd/cdfs/src/cdfs_iop.h
@@ -1,8 +1,9 @@
 #ifndef _CDFS_IOP_H
 #define _CDFS_IOP_H
 
+#define MODNAME "cdfs_driver"
 #ifdef DEBUG
-#define DPRINTF(args...)	printf(args)
+#define DPRINTF(args...)	printf(MODNAME ": "args)
 #else
 #define DPRINTF(args...)	do { } while(0)
 #endif
--- a/iop/network/smap/src/smap.c
+++ b/iop/network/smap/src/smap.c
@@ -117,7 +117,7 @@ static u16 _smap_read_phy(volatile u8 *emac3_regbase, unsigned int address)
     } while (i < 100);
 
     if (i >= 100)
-        printf("smap: %s: > %u ms\n", "_smap_read_phy", i);
+        printf("SMAP: %s: > %u ms\n", "_smap_read_phy", i);
 
     return result;
 }
@@ -140,7 +140,7 @@ static int DisplayHelpMessage(void)
 static inline void RestartAutoNegotiation(volatile u8 *emac3_regbase, u16 bmsr)
 {
     if (EnableVerboseOutput)
-        DEBUG_PRINTF("smap: restarting auto nego (BMCR=0x%x, BMSR=0x%x)\n", _smap_read_phy(emac3_regbase, SMAP_DsPHYTER_BMCR), bmsr);
+        DEBUG_PRINTF("restarting auto nego (BMCR=0x%x, BMSR=0x%x)\n", _smap_read_phy(emac3_regbase, SMAP_DsPHYTER_BMCR), bmsr);
     _smap_write_phy(emac3_regbase, SMAP_DsPHYTER_BMCR, SMAP_PHY_BMCR_ANEN | SMAP_PHY_BMCR_RSAN);
 }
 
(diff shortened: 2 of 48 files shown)


israpps:
dont forget to squash instead of common merge xd

iop/cdvd/cdfs/src/cdfs_iop.h
```diff
@@ -2,7 +2,7 @@
 #define _CDFS_IOP_H
 
 #ifdef DEBUG
-#define DPRINTF(args...)	printf(args)
+#define DPRINTF(args...)	printf("CDFS: "args)
```
AKuHAK, line 5: why not use?
```
#define MODNAME "CDFS"
IRX_ID(MODNAME, 1, 1);

#define M_PRINTF(format, args...) \
    printf(MODNAME ": " format, ##args)
```
like here:
https://github.com/ps2dev/ps2sdk/blob/master/iop/dev9/atad/src/ps2atad.c#L40

israpps: I thought that the string passed to IRX_ID() macro had to remain EXACTLY the same...

If I'm wrong, and that can be done, please let me know

israpps: so...
you want me to use modname on all of them?

i would like to keep it this way

fjtrujy: I think that using @AKuHAK approach will be a better solution

asmblur (approved):
Looks fine to me.

asmblur (approved):

iop/network/smap/src/smap.c
```diff
@@ -154,12 +154,12 @@ static int InitPHY(struct SmapDriverData *SmapDrivPrivData)
 
     LinkSpeed100M = 0;
     if (EnableVerboseOutput != 0)
-        DEBUG_PRINTF("smap: Resetting PHY\n");
+        DEBUG_PRINTF("SMAP: Resetting PHY\n");
```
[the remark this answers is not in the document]
israpps: > "SMAP: " should be removed here (many times in this file), becouse DEBUG_PRINTF also adds "SMAP: "

seems I missed that one.

Will deal with it now

fjtrujy:
@israpps , do you have anything else pending to do in this PR?

I think you still miss something else, which is remove the `printf` function from the import list if not needed. If the IOP driver is always just using `printf` when `DEBUG`, you should remove that import as well.

I think you should rebase the PR and ensure that all the drivers follow this pattern.

Additionally, I think that is good to merge this, as it could save some KBs on the IRX generation as skipping the usage of `printf` functions (I'm not really sure as `printf` is a external function from `stdio`).

israpps:
> @israpps , do you have anything else pending to do in this PR?
> 
> I think you still miss something else, which is remove the `printf` function from the import list if not needed. If the IOP driver is always just using `printf` when `DEBUG`, you should remove that import as well.
> 
> I think you should rebase the PR and ensure that all the drivers follow this pattern.
> 
> Additionally, I think that is good to merge this, as it could save some KBs on the IRX generation as skipping the usage of `printf` functions (I'm not really sure as `printf` is a external function from `stdio`).

Printf comes from system module `STDIO.IRX`

Some modules have a greeting printf on their start function indicating versión and other stuff...

IIRC I had nothing left to do with this PR...

I'll check it tonight if i can

israpps:
Made a quick test

Importing from an IRX module increases binary in 16 bytes, and every imported func is another 16 byte increase

Wich means that the removal of printf implies reducing binary of at least 16 or 32 bytes (depending if IRX uses other exports from stdio).
This ofc does not consider size increase of the strings passed to printf

fjtrujy:
> Made a quick test
> 
> Importing from an IRX module increases binary in 16 bytes, and every imported func is another 16 byte increase
> 
> Wich means that the removal of printf implies reducing binary of at least 16 or 32 bytes (depending if IRX uses other exports from stdio). This ofc does not consider size increase of the strings passed to printf

So then, is it ready to be merged? Have you rebased and checked if new drivers recently added miss this new way of print info?

@uyjulian what do you think?

Cheers.

uyjulian:
These set of changes right now should be fine.

fjtrujy (approved):

PR #418, opened 2023-03-28
64bit LBA support for bdm and associated file system/device modules

grimdoomer:
This PR adds support for 64bit LBAs in bdm and file system/device modules that utilize bdm: 
- Bdm, libbdm, and bdmfs_fatfs modules have full support for 64bit LBAs
- Added support for parsing and mounting multiple partitions in MBR or GPT style
- Atad module has been updated to support a maximum of 48bit LBAs (that is the maximum per the ATA standard)
- Bdmfs_vfat, usb, mx4sio, and iLink modules have been updated for the bdm API changes but still use 32bit LBA commands for the respective devices

These changes allow for using block devices including internal HDDs larger than 2TB. When used with exfat internal HDDs now have a max size limit of 144 petabytes (LBA 48 max size). For block devices that support full 64bit LBAs the max addressable size using exfat is 16 exabytes. 

Additionally I have added some debugging changes to the ee kernel library to re-enable the print syscall for EE uart debug output (see: ee/kernel/src/debug.c). This is helpful in situations where linking in printf via libc is undesirable due to the size increase it adds. By re-enabling the print syscall you can now use existing EE uart printf functionality in the kernel, and it only requires a few lines of code to enable via the InitDebug function.

All of these changes have been tested on my fork of OPL that adds in support for large capacity internal HDDs: https://github.com/grimdoomer/Open-PS2-Loader/tree/hdd_large_capacity_support

Please feel free to squash merge this PR upon acceptance.


uyjulian:
To fix the cppcheck warning, please kindly add before line 700 in `iop/sio/mx4sio_bd/src/mx4sio.c`:

```
// U64_2XU32 expands to 2 arguments
// cppcheck-suppress wrongPrintfScanfArgNum
```

Also, this seems to be hard to review due to the large amount of unrelated changes.
May I suggest splitting up your commits., and removing the unneeded changes, such as comment/uncommenting, adding or removing newlines, adding/removing debug print statements, etc.

One way to do this is the following:
1. `git rebase -i upstream master` to start the rebase process
2. Change all lines to "e", save and quit
3. `git diff HEAD~1 HEAD | xclip -selection clipboard` to copy the differences to your clipboard
4. Paste into a text editor and delete the unneeded hunks. Then copy back into your clipboard
5. `git checkout HEAD~1 .` to checkout the files before the current commit
6. `xclip -out -selection clipboard | patch -p1` 
7. `git add .`
8. `git commit --amend` then save and quit.
9. To add more commits in between use `git add .` and `git commit`
10. `git rebase --continue`, then go back to step 3 if there are more commits to clean

Also, instead of the manual process I described, you can also use Sublime Merge to edit commits.

iop/dev9/atad/src/ps2atad.c
```diff
@@ -888,29 +894,37 @@ static int ata_device_set_transfer_mode(int device, int type, int mode)
 /* Export 9 */
 /* Note: this can only support DMA modes, due to the commands issued. */
 int ata_device_sector_io(int device, void *buf, u32 lba, u32 nsectors, int dir)
+{
+    return ata_device_sector_io_internal(device, buf, (u64)lba, nsectors, dir);
+}
+
+int ata_device_sector_io_internal(int device, void *buf, u64 lba, u32 nsectors, int dir)
```
bignaux, line 901: could we use ata_device_sector_io64 instead of ata_device_sector_io_internal, and add a DECLARE_IMPORT for it in atad.h , in order to let homebrew that use atad able to use 64bits lba ?

iop/dev9/atad/src/ps2atad.c
```diff
@@ -197,11 +199,14 @@ static void ata_multiword_dma_mode(int mode);
 static void ata_ultra_dma_mode(int mode);
 static void ata_shutdown_cb(void);
 
+int ata_device_sector_io_internal(int device, void *buf, u64 lba, u32 nsectors, int dir);
+
 #ifdef ATA_ENABLE_BDM
-static int ata_bd_read(struct block_device *bd, u32 sector, void *buffer, u16 count);
-static int ata_bd_write(struct block_device *bd, u32 sector, const void *buffer, u16 count);
+static int ata_bd_read(struct block_device *bd, u64 sector, void *buffer, u16 count);
+static int ata_bd_write(struct block_device *bd, u64 sector, const void *buffer, u16 count);
```
bignaux, line 206: even if we have to change the prototype for these functions, couldn't we take the opportunity to standardize them to the POSIX.1-2001 standard by using the following prototypes: 
ssize_t pread(int fd, void *buf, size_t count, off_t offset);
ssize_t pwrite(int fd, const void *buf, size_t count, off_t offset);

bignaux (changes requested):

rickgaiser (changes requested):
Congratulations on making this work, and thank you for sharing your code.

However, I agree with @uyjulian that this PR is too large and should be split in separate PR's. Smaller PR's can be discussed and accepted much more easy than a big PR like this.

> The individual commits are worthless on their own, only the aggregate matters. It would take me way longer to redo the commits than it will to review the aggregate. I'm happy to answer any questions you have or address concerns with the changes, but I don't really have the time or interest in redoing the individual commits when they can just be squashed anyway.

I understand this will take time, for you and for all of us. But this is what working together on open source software means. And in general, this will lead to better quality code. May I suggest splitting the code into the following PR's?:
- Add 64bit support to BDM
- Add GPT partition driver
- Add 48bit LBA support to ATA
- Add kernel debugging function

Also, as requested on discord, did you test the free/consumed IOP memory before and after this aggregate change? What are the results using which block device?

grimdoomer:
I can split the kernel printf code out into its own PR but there is no way I'm creating and submitting 4 different PRs. Having these changes as separate commits does not provide any value as there's no reason to use one change without the others. Having 64bit support in bdm is useless if it will only ever be provided 32bit sector addresses. Having GPT support is useless if no device driver can address passed 32bits, and the LBA 48 change in atad is less than 10 lines and does not need its own PR.

If someone wants to fork my topic branch and break it up and submit individual PRs they are free to do so. But I do not have the time or interest to do the work and go through the PR process 4 separate times for what is not even a medium sized change.

uyjulian:
Split it into 4 different commits, not 4 different PRs.

It shouldn't take more than 1 hour to do

iop/dev9/atad/src/ps2atad.c
rickgaiser: No, pread and pwrite can write byte sized data and we do not (or want to) support that. Also, our read/write functions are part of the object oriented block_device structure, so they need the block_device structure as their first argument.

iop/dev9/atad/src/ps2atad.c
rickgaiser: Good idea. This can be added in this PR, or in a later one when the import is needed.

iop/fs/bdm/src/part_driver.c
rickgaiser, line 1: GPT and MBR are 2 unrelated partitioning schemes. In BDM, partitioning schemes are implemented as 'file_system' drivers. They should be implemented as 2 separate drivers, and not merged into 1 driver to rule them all.

BDM will ask all drivers 1-by-1 if they can handle the file system on the block device. If 1 file system cannot handle a block device, then the next file system will be asked if it can handle the block device. What I'm expecting for a GPT partitioned block device (for example an 8TB usb stick, usb0p0) is:

Stage 1, mounting the "Protective MBR":
- BDM asks the MBR driver if it can handle usb0p0. MBR will find the "Protective MBR" header of GPT, and registers a new block device of 2TB (type EEh), this will become usb0p1.

Stage2, mounting the GPT patition
- BDM asks the MBR driver if it can handle usb0p1, it cannot.
- BDM asks the GPT driver if it can handle usb0p1, it can and registers a new block device of 8TB. This will become usb0p2.

Stage 3, mounting the exFat file system:
- BDM asks the MBR driver if it can handle usb0p2, it cannot.
- BDM asks the GPT driver if it can handle usb0p2, it cannot.
- BDM asks the exFat driver if it can handle usb0p2, it can, and will be accessible as "mass" or "mass0"

iop/fs/bdmfs_fatfs/src/ff.c
```diff
@@ -3385,8 +3387,9 @@ static FRESULT mount_volume (	/* FR_OK(0): successful, !=0: an error occurred */
 	if (SS(fs) > FF_MAX_SS || SS(fs) < FF_MIN_SS || (SS(fs) & (SS(fs) - 1))) return FR_DISK_ERR;
 #endif
 
-	/* Find an FAT volume on the drive */
-	fmt = find_volume(fs, LD2PT(vol));
+	/* Check for a valid file system on the device. */
+	fmt = check_fs(fs, 0);
```
rickgaiser, line 3391: What does this do?

iop/fs/bdmfs_fatfs/src/fs_driver.c
```diff
@@ -627,9 +639,9 @@ static int get_frag_list(FIL *file, void *rdata, unsigned int rdatalen)
             // Fragment or file end
             M_DEBUG("fragment: %uc - %uc + 1\n", iClusterStart, iClusterCurrent + 1);
             if (iFragCount < iMaxFragments) {
-                f[iFragCount].sector = clst2sect(file->obj.fs, iClusterStart);
+                f[iFragCount].sector = clst2sect(file->obj.fs, iClusterStart) + bd->sectorOffset;
```
rickgaiser, line 642: Why is the sectorOffset needed here, when it was not needed before?

iop/fs/bdmfs_fatfs/src/fs_driver.c
```diff
@@ -667,12 +679,26 @@ int fs_ioctl2(iop_file_t *fd, int cmd, void *data, unsigned int datalen, void *r
...
+            {
+                ret = -EINVAL;
+                break;
+            }
+
+            // Get the block device backing the file so we can get the starting LBA of the file system.
+            struct block_device* bd = fatfs_fs_driver_get_mounted_bd_from_index(file->obj.fs->pdrv);
+            
+            *(u64*)rdata = clst2sect(file->obj.fs, file->obj.sclust) + bd->sectorOffset;
+            ret = 0;
```
rickgaiser, line 693: Same here, why is this needed, when it was not needed before?

iop/fs/bdmfs_fatfs/src/fs_driver.c
```diff
@@ -681,6 +707,18 @@ int fs_ioctl2(iop_file_t *fd, int cmd, void *data, unsigned int datalen, void *r
...
+        {
+            struct block_device *mounted_bd = fatfs_fs_driver_get_mounted_bd_from_index(fd->unit);
+            if (mounted_bd == NULL)
+                ret = -ENXIO;
+            else if (mounted_bd->ioctl == NULL)
+                ret = -ENOTSUP;
+            else
+                ret = mounted_bd->ioctl(mounted_bd, cmd, data, datalen, rdata, rdatalen);
+            break;
+        }
```
rickgaiser, line 721: Why is it needed to know the 'bits' that are supported by the driver and/or the interface?

The number of bits needed to address the block device can also be extracted from the size.

iop/sio/mx4sio_bd/src/mx4sio.c
```diff
@@ -875,15 +912,15 @@ static void sd_detect_thread(void *arg)
...
-
         M_DEBUG("Check card, inserted=%d, used=%d\n", card_inserted, card_used);
 
         // Detect card if it has not been used recently
         if (card_used == 0)
             sd_detect();
         card_used = 0;
+
+        // Sleep for 1 second
+        DelayThread(1000 * 1000);
```
rickgaiser, line 923: Originally this delay was right there where you put it, but there where issues with this. By putting it at the beginning of the loop there is an (intentional) delay before the first poll starts after the thread is created.

Is there a reason why this is changed, besides being a little faster, has this also been proven stable?

rickgaiser (changes requested):

iop/fs/bdm/src/part_driver.c
AKuHAK: I personally think that GPT and MBR can be merged. Moreover, my opinion is that we do not have to support either MBR or GPT and mount whole device as one raw block (removing part_driver completely) and use GPT/MBR support that is built in FatFS driver itself (it can handle both in some minimal level).

iop/fs/bdmfs_fatfs/src/ff.c
AKuHAK: I think this check is for checking if after all mountings we have FatFS compatible system. This actually may be good idea, cause I once corrupted NTFS drive with wLaunchELF that incorrectly handled it it as FAT32 filesystem in partition table.

common/include/bdm-common.h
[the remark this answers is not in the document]
grimdoomer: The ioctl interface is specifically to get information from the block device on the EE side. This is important if you want to be able to determine the device index, LBA size, or name of the underlying block device driver (ie: what USBMASS_IOCTL_GET_DRIVERNAME was doing, but since this applies to all block devices it's not really a USBMASS specific ioctl code).

This is especially important with the changes I've made in OPL:
1. Supporting multiple devices of the same type (ex: supporting two USB devices at the same time). This requires to know the device index (devNr) on the EE side in order to pass this info to OPL cdvdman so that it can bind to the correct device index.
2. The LBA size is required to inform OPL cdvdman if it should use LBA48 or LBA28 ATA commands, as the atad code in OPL does not include ata_identify to query it dynamically. Previously this information was retrieved via the xhdd IOP module. However, now that bdm supports 64bit LBAs it would be trivial to add support to the usb and iLink modules to also support LBAs > 32bits, in which case the LBA size would again need to be known in order to know which command set to use when accessing the device in cdvdman/etc.

Rather than add device specific ioctl codes for each device type (some of which depend on additional IOP modules being loaded), it made more sense to add an ioctl interface to bdm and have each device specific module that supports bdm to "pass-through" bdm specific ioctl codes.

iop/fs/bdm/src/part_driver.c
grimdoomer: I can split them into two separate fs drivers but there are a couple issues I have with the flow you outlined.

1. The MBR driver should explicitly fail to mount the protective MBR, as it exists only as a fallback for devices which are unaware of the protective MBR partition type, and is a "best effort" approach. It exists to tell the device "the entire disk has already been accounted for", and in the case of actually trying to mount a file system via the protective MBR, it will almost always fail once you get to 3TB+ device sizes due to how most file systems actually place data on the disk.
2. The MBR/GPT drivers should only create new block devices for individual partitions. The LBA range for the MBR/GPT sectors is already "owned" by the underlying storage device driver (ex: usb, ata, etc). Mounting this LBA range twice will just lead to recursion issues and waste IOP memory.
3. Once bdm finds a fs driver that can mount a given LBA range it stops asking subsequent fs drivers to attempt to mount. In your flow above this means it would stop after mounting the protective MBR, which is why you need to "double" mount the LBA range, only to have the GPT driver later violate the new LBA range as specified by the protective MBR. This is not only wasteful but also bad design to have the MBR driver specify an LBA range for the new block device that the GPT driver will just violate.

Overall I think the steps to take here are:
1. Split MBR and GPT into separate fs drivers
2. Have the MBR driver explicitly fail to mount protective MBR type partitions so a subsequent fs driver can try to handle it properly

iop/fs/bdmfs_fatfs/src/fs_driver.c
grimdoomer: Before these changes exfat "owned" the entire storage device, as in, the block device it would "bind" to had a sector offset of 0. For each exfat partition the file->obj.fs->database value would get set to the sector offset for the start of the partition (relative to the start of the storage device as a whole). When clst2sect was called this would get added to the fragment sector offset to get an absolute sector address.

After these changes exfat only "binds" to individual partitions that have non-zero sector offsets. Thus the sector offset for the partition is now 0 (because we're now relative to the start of the partition, not the storage device). So we now need to add the starting LBA of the block device (which is the LBA of the partition relative to the start of the storage device) to the fragment sector to get the absolute sector address.

iop/fs/bdmfs_fatfs/src/fs_driver.c
grimdoomer: See explanation for the previous comment.

iop/fs/bdmfs_fatfs/src/fs_driver.c
grimdoomer: Explained in the comment about why add an ioctl interface to bdm: this is used on the EE side to inform OPL cdvdman which device command set to use for proper device access.

iop/sio/mx4sio_bd/src/mx4sio.c
grimdoomer: With the delay in its original location my mx4sio card would take a long time to mount in OPL. This particular thread runs at low priority, and from my debugging would take around 5 seconds before it got scheduled in even though the wait is supposed to be 1 second. Since the comment didn't note that the location of the wait was intentional I moved it, which resulted in the mx4sio card getting mounted immediately.

I have not experienced any issues after moving it, but if there previously were issues regarding it's location then a larger sample size would be needed to say if it's stable or not.

common/include/bdm-common.h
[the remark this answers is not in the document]
rickgaiser: So there's 2 things you would like to know from the block device:

1) The id is already known in the block device struct. The filesystem driver can return `bd->devId`. Or am I missing something here?

2) Why do we need the usb/ilink/mx4sio/udpbd drivers to have code to tell the file system driver if it's 32/64bit? Does not make sense. Especially when also used ingame, when there's not even a file system running. It's a waste of code and space.
The problem you're trying to solve with ATA can be solved differently too I think:
-- per read/write to <2TB can be LBA28, >=2TB must be LBA48
-- you can check the location of the ISO first, then have all reads/writes be LBA28/48
-- have the ingame ata driver figure out what it should do
But more generically speaking. You would like to have full-featured block device drivers and small block device drivers. Then ask the full-featured driver for information that the small driver cannot know. Like a connection string or struct.
I think the ingame smb support has a similar problem, only solved differntly. It has a separate initialization irx that is no longer present after initialization.

Anyway, I think this change is not really needed or can be solved differently without bloating all block device drivers.

iop/fs/bdm/src/part_driver.c
rickgaiser: 1. I agree, failing for partition type FFh (as it currently does I think) would be better. This would remove stage1 as described above, and save 1 entry in the file_system table in BDM.
2. There's nothing wrong with recursion when it comes to partitions. This is actually what makes extended MBR partitions work. It's literally 4 partitions inside another partition. For instance in linux:
image
In this example /dev/sda2 is an extended partition. So the sectors of for instance /dev/sda5 are used by ntfs+sda5+sda2+sda. But they don't all "own" the data, there's no "ownership" in the same way we normally speak ow ownership when allocating/deallocating dynamic memory. BDM uses the same model as Linux.
3. In my example MBR would mount 1 partition, then provide a partition of size-1 sector. But see 1., I agree that failing in the MBR driver would be better, then mounting only once in GPT (without recursion).

I fully agree with your conclusion 1 and 2.

iop/fs/bdmfs_fatfs/src/ff.c
rickgaiser: Great! Thank you for clarifying and fixing this bug.

iop/fs/bdmfs_fatfs/src/fs_driver.c
rickgaiser: Again great! Thank you for clarifying and fixing this bug.

iop/sio/mx4sio_bd/src/mx4sio.c
rickgaiser: Your debugging must have been using 100% cpu for 5 seconds with a higher priority if it took so long.

The whole card detection thing is terrible and in need of a rewrite if you're interested? When there is no card inserted, but the mx4sio driver is loaded, this loop will hang the sio2 bus every second, causing OPL (or any other application) to become unresponsive since controller input is handled over the same sio2 bus.

But anyway, this change can couse side effect and it's off-topic for this PR, so I think it's better to remove it.

common/include/bdm-common.h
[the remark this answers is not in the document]
grimdoomer: 1. LBA28 cannot address up to 2TB, only 128GB.
2. LBA28/48 is specific to ATA, other devices like SCSI, USB, (SD as well I believe) use 32/64 bit LBA commands. This means the exfat driver would have to first check the underlying device driver and then handle accordingly to determine if the device needs small/large LBA commands. Which command set to use is an implementation detail of the device driver, the file system drivers, which are completely independent of the device drivers, should not have device-specific logic in them. This creates unnecessary relationships between them resulting in harder to maintain code.
3. Checking the location of the ISO does not work, you will need to walk the entire fragment chain and make sure all fragments are below the LBA split address, which requires adding more code to the exfat driver to do so, and again, requires device specific logic.
4. What do you consider the "ingame ata driver"? Is this atad? Cdvdman in OPL? Or something else entirely?

"You would like to have full-featured block device drivers and small block device drivers. Then ask the full-featured driver for information that the small driver cannot know."

This makes no sense. You only have 1 block device driver. Your "mini" block device drivers have the exact same amount of code as the full versions. The only difference is the mini ones don't have debug output, and use smaller stacks/number of support devices. But all the code is the same. 

You say the "ingame" driver should figure it out and that's exactly what this does. The underlying device driver is the one determining what command set should be used and reporting that information via the ioctl interface. The file system driver is just passing through the ioctl to the underlying device driver, this is exactly how the IO stack in every major OS works for the past 30+ years.

Adding device specific implementation details to the exfat driver means anytime someone creates a new bd device driver they will have to go modify the exfat driver to handle logic specific to that device. If someone creates another file system driver (say for ext or something), they will have to go add the same device specific logic to it. This is a bad design pattern and creates harder to maintain code. These details should be left up to the device specific driver, not the file system drivers.

You call this bloat yet you just asked me to split the MBR and GPT code which will result in duplicating ~100 lines of code AND static arrays for partition tracking. Your suggestions are what is creating bloat and introducing generally bad design patterns which ends up creating harder to maintain code.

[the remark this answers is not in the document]
rickgaiser: This is the problem you're trying to solve, correct?:

> The LBA size is required to inform OPL cdvdman if it should use LBA48 or LBA28 ATA commands, as the atad code in OPL does not include ata_identify to query it dynamically.

So we're basically talking about 1 bit needed in cdvdman (`lba_48bit`). But why do you even need this bit?
I think becouse there are 2 ATA drivers:
1. The full version in ps2sdk, with BDM support already added by @uyjulian (https://github.com/ps2dev/ps2sdk/commit/4340cf9007ad572764eed3eb800b96d122b2b8e8)
2. The crippled/copied version in OPL ingame (cdvdman), with BDM support added by you (https://github.com/grimdoomer/Open-PS2-Loader/blob/hdd_large_capacity_support/modules/iopcore/cdvdman/atad.c)

We're talking about a PR to ps2sdk, but I think some background on OPL is required here as well. The OPL ingame support currently has three modes:
- hdd_cdvdman.irx
- smb_cdvdman.irx
- bdm_cdvdman.irx + bdm block device driver from ps2sdk (usb / ilink / mx4sio / udpbd? / ata?)

Adding BDM support for ATA would have been as simple as loading the ps2sdk ATA driver, with the BDM version of cdvdman. Changing only a few lines of code. Take for instance the addition of mx4sio:
https://github.com/ps2homebrew/Open-PS2-Loader/commit/883fc1848396e57b24eae0d2420a633490cc8763
All that was needed to add mx4sio was: "8 changed files with 32 additions and 2 deletions". ATA BDM support should have been the exact same change.

But instead it seems like you took the difficult route and created a completely new mode: "bdm_ata_cdvdman.irx". Along with an ATA driver that does not seem to know what sort of ATA drive it's talking to. And not we're discussing in a ps2sdk PR how to fix this issue.

> This means the exfat driver would have to first check the underlying device driver and then handle accordingly to determine if the device needs small/large LBA commands. Which command set to use is an implementation detail of the device driver, the file system drivers, which are completely independent of the device drivers, should not have device-specific logic in them. This creates unnecessary relationships between them resulting in harder to maintain code.

I see we want the same thing: easy to maintain code without unnecessary relationships between the FS driver and the BD driver. All the FS driver needs to know about the BD, is in the existing interface: `struct block_device`. That is why I don't want to add an extra ioctl with dummy functions to all block devices.

> Checking the location of the ISO does not work, you will need to walk the entire fragment chain and make sure all fragments are below the LBA split address, which requires adding more code to the exfat driver to do so, and again, requires device specific logic.

Why would you do this in the exFat driver? Do it in OPL GUI. This is the place where you have both the fragment chain, and knowledge about what drivers you're loading. It requires less code and less dependencies that the ioctl's you're proposing.

> You say the "ingame" driver should figure it out and that's exactly what this does.

Yes and no. the ps2sdk driver should be the ingame driver. Then it knows how to figure things out. But you used another driver that does not know how to figure things out.

> Adding device specific implementation details to the exfat driver means anytime someone creates a new bd device driver they will have to go modify the exfat driver to handle logic specific to that device. If someone creates another file system driver (say for ext or something), they will have to go add the same device specific logic to it. This is a bad design pattern and creates harder to maintain code. These details should be left up to the device specific driver, not the file system drivers.

Did I say otherwise? I fully agree with this.
"These details should be left up to the device specific driver" <- this is exactly what is wrong with the ata driver in OPL.

> You call this bloat yet you just asked me to split the MBR and GPT code which will result in duplicating ~100 lines of code AND static arrays for partition tracking. Your suggestions are what is creating bloat and introducing generally bad design patterns which ends up creating harder to maintain code.

Don't make this personal and try to stay polite. If you can convince me that your solution is the better one, then I would be happy to admit, as I've done with for instance your idea on MBR/GPT mounting/failing.

iop/dev9/atad/src/ps2atad.c
TnA-Plastic: This is good!

common/include/bdm-common.h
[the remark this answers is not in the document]
grimdoomer: >So we're basically talking about 1 bit needed in cdvdman (lba_48bit). But why do you even need this bit?

It's needed to determine if the atad code should use LBA28 commands or LBA48 commands. While it would be great to just always use LBA48 commands there are people using drives that only support LBA28 commands. So the distinction is needed. And again, it would be trivial to add large LBA support to USB and iLink drivers as well, and they would run into the same issue.

>But instead it seems like you took the difficult route and created a completely new mode: "bdm_ata_cdvdman.irx". Along with an ATA driver that does not seem to know what sort of ATA drive it's talking to.

This is only half correct, the bdm_ata_cdvdman is just the existing bdm_cdvdman using the atad code that is normally only compiled into hdd_cdvdman, only now it registers the hdd as a block device. Everything I used was already existing in OPL. Why not use the ps2sdk atad driver? Because it has a bunch of code that will never be used and only takes up IOP memory (smart status checking, locking/unlocking hdd, etc.). This slimmed down version is ~45% smaller than the version in ps2sdk.

I'm not even going to argue on the rest of the points. I already found another way to implement this for ATA, and I have no intentions of updating usb or iLink for large LBA support so it doesn't even matter.

iop/dev9/atad/src/ps2atad.c
grimdoomer: Changed to ata_device_sector_io64 and exported in the latest changes

iop/fs/bdm/src/part_driver.c
grimdoomer: MBR and GPT have been split into two separate fs drivers in latest changes.

rickgaiser (approved):

israpps:
https://github.com/ps2dev/ps2sdk/actions/runs/4617178853/jobs/8163660743?pr=418#step:5:1003

rickgaiser:
Squash merged, thanks @grimdoomer

PR #421, opened 2023-04-10
[EE Makefile]: imitate IOP Makefiles

israpps:
Allow developers to change source files location and object files output location without copy pasting recipes. just declaring the `EE_*_DIR` variables

### DIFF
--- a/samples/Makefile.eeglobal_sample
+++ b/samples/Makefile.eeglobal_sample
@@ -50,6 +50,8 @@ endif
 
 # Externally defined variables: EE_BIN, EE_OBJS, EE_LIB
 
+EE_OBJS := $(EE_OBJS:%=$(EE_OBJS_DIR)%)
+
 # These macros can be used to simplify certain build rules.
 EE_C_COMPILE = $(EE_CC) $(EE_CFLAGS) $(EE_INCS)
 EE_CXX_COMPILE = $(EE_CXX) $(EE_CXXFLAGS) $(EE_INCS)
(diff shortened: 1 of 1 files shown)


samples/Makefile.eeglobal_sample
```diff
@@ -50,30 +50,40 @@ endif
...
+$(EE_OBJS_DIR)%.o: $(EE_SRC_DIR)%.cpp
 	$(DIR_GUARD)
 	$(EE_CXX) $(EE_CXXFLAGS) $(EE_INCS) -c $< -o $@
 
-%.o: %.S
+$(EE_OBJS_DIR)%.o: $(EE_ASM_DIR)%.S
+	$(DIR_GUARD)
+	$(EE_CC) $(EE_CFLAGS) $(EE_INCS) -c $< -o $@
+
+$(EE_OBJS_DIR)%.o: $(EE_ASM_DIR)%.s
```
fjtrujy, line 78: I think you’re duplicating this rule

israpps: I'm on my cellphone and line wrapping is horrible

If you mean the duplication of the ASM source files. Then yes
Two pairs for the ASM die, another for the SRC for

If there's a makefile trickery to achieve this without duplication of recipes please let me know

#### the purpose behind this?
OPL, KELFBinder, Athena and other use the ASM subdir for ditching the resulting files of `$(bin2s)`

This ensures that if the program has assembly source files, they're searched on both places

PR #422, opened 2023-04-13
FatFS: Mount immediately so we can return ok/error

rickgaiser:
The description of this changed parameter:
`Mount option: 0=Do not mount (delayed mount), 1=Mount immediately`

Setting this to 0 will always return succes, even if the filesystem is not supported. We need to "mount immediately" so we can report back to BDM if this is a supported filesystem, or not.

This also solves an issue in OPL when you have 2 partitions:
- the first being an usupported partition (like ntfs, ext4, etc...) - that would falsely be mounted as "mass0"
- the second being the partition with games - that would not show up

@grimdoomer do you agree on this PR, or do you have other things in the works that solve the same issue?


grimdoomer:
Yes this looks good, I don't have anything that conflicts with it.

PR #427, opened 2023-04-23
Revert "[EE Makefile]: imitate IOP Makefiles"

fjtrujy:
Reverts ps2dev/ps2sdk#421

Because it is making `ps2sdk-ports` to not compile, some `makefile` over there needs to be updated.

@israpps is already working on it, he will create the PR back, when the other PRs from `ps2sdk-ports` were ready.

Cheers.


israpps:
i still dont understand the issue on ps2sdk ports tho

I think it would be sane to bring my idea into another makefile sample

will not break any existing project and anyone can change their makefiles to use the new sample saving some sample duplications

PR #429, opened 2023-04-30
Improve errno usage

fjtrujy:


AKuHAK:
@fjtrujy @uyjulian 
this pr started to produce strange issues in OPL:
https://github.com/ps2homebrew/Open-PS2-Loader/issues/988
confirm to work if it is reverted

fjtrujy:
I need more information about what function is wrong….. where is it failing?
Watching the issue you attached I don’t know anything…..

AKuHAK:
I wasnt able to debug the issue

israpps:
> @fjtrujy @uyjulian this pr started to produce strange issues in OPL: [ps2homebrew/Open-PS2-Loader#988](https://github.com/ps2homebrew/Open-PS2-Loader/issues/988) confirm to work if it is reverted

Ask them to move the OPL config somewhere else, configure again.

If issue is fixed, then it is another case of https://github.com/ps2homebrew/Open-PS2-Loader/issues/978 and they should upload the original config

AKuHAK:
@israpps he tested games on fresh usb. They both confirmed that reverting this pr solved dvd9 problem.

fjtrujy:
But maybe the issue is on the driver that replace the dvd/cdrom unit

AKuHAK:
@fjtrujy it is really hard to track down this, can it be this function:
```c
int configGetStat(config_set_t *configSet, iox_stat_t *stat)
{
    return (fileXioGetStat(configSet->filename, stat) >= 0 ? 1 : 0);
}
```

fjtrujy:
> @fjtrujy it is really hard to track down this, can it be this function:
> 
> ```c
> int configGetStat(config_set_t *configSet, iox_stat_t *stat)
> {
>     return (fileXioGetStat(configSet->filename, stat) >= 0 ? 1 : 0);
> }
> ```

It can't be this one. I have just changed `POSIX` return error function. 
So if you take a look at the changes functions in the `ee/rpc/filexio/src/fileXio_rpc.c`, I have just changed the `helpers` functions, which are just used for `ee/libcglue`.

So this PR just will change the value in POSIX functions, such as `read, open, close, write....` (it could affect if you use these "private" functions, that you shouldn't https://github.com/fjtrujy/ps2sdk/blob/bf4b4e5db0c5986cea574ba17604fe942d08c9b6/ee/libcglue/include/ps2sdkapi.h#L19-L39) but it shouldn't change anything in these `fio` or `fileXio` functions, because I didn't change any behavior there.

Cheers

fjtrujy:
We have found the issue @AKuHAK , I will create a PR right now

PR #432, opened 2023-05-22
Added some vnops to allow time for the clip flags to update in calculate_vertices

hello01-debug:
On the PS2, the clip flag is pipelined, and takes approximately 3-4 cycles to complete after a `vclipw`. This pull request adds 4 `vnops` after `vclipw` in the `calculate_vertices` function to allow time for the flag to properly update. This should fix the broken clipping seen in `cube.elf`.


uyjulian:
Thanks

PR #438, opened 2023-06-05
Fix issue with lseek64

fjtrujy:
This PR has been created thanks to @AKuHAK for investigating an issue after merging #429 
Create a specific method for transform errno returning 64 bits


uyjulian (approved):
lgtm

PR #440, opened 2023-06-13
For assembly code, use .set push ... .set pop instead of .set X ... .set noX

uyjulian:


asmblur:
Absolutely the right way.

rickgaiser:
I agree, but at the same time this pr can cause very strange issues that will be difficult to trace back to this PR.
Merge?

asmblur:
I would expect the issues to be few and far between. Worst case scenario, we check every change of anything pops up.

PR #444, opened 2023-07-06
Update slib.c

israpps:
a PR of what has been described here https://github.com/ps2dev/ps2sdk/commit/b60a086d5afe1ed0659ad1bdf6415705d76ca252#r120739506


israpps:
seems like a header file is missing

Lets see if now it compiles

PR #445, opened 2023-07-06
Fix compilation of slib

uyjulian:
Because testing is too hard


israpps:
> Because testing is too hard

hey. I didnt merge it. and if i wanted to. i would wait for CI to finish

PR #451, opened 2023-07-14
mx4sio v1.2

qnox32:
This is equal to the mx4sio driver version used in the OPL BDM 4.8.1 build people have been testing. 

The RPC builds and xprintf.c are not built by default and are only used in conjunction with the sdcard debugger for testing. They serve a limited purpose and can be removed if needed.


israpps:
You could definitely strip away the RPC printf stuff and use EE `freopen()` as a solution.

I tested it on KELFBinder

Moves both IOP and EE printf to file Without issues.
Resulting in an excellent method to get logs.

qnox32:
> You could definitely strip away the RPC printf stuff and use EE `freopen()` as a solution.
> 
> I tested it on KELFBinder
> 
> Moves both IOP and EE printf to file Without issues. Resulting in an excellent method to get logs.

Removed RPC builds and xprintf.c for the time being. In future I might try to re-implement their functionality using freopen() as you suggested, but there is no ETA on that at the moment.

PR #452, opened 2023-07-14
krat0s: fix RPC calls to properly free packets as per original code

AKuHAK:
by krat0s
This will solve some freezes in some applications after multiple IOP resets


rickgaiser:
`rpc_packet_free` is defined 3x. Why not make this a static inline function, not guarded by `#ifdef F_*` ?

AKuHAK:
> `rpc_packet_free` is defined 3x. Why not make this a static inline function, not guarded by `#ifdef F_*` ?

Done.

PR #453, opened 2023-07-15
Set maximum supported UDMA mode, instead of hardcoded UDMA4

AKuHAK:


fjtrujy:
@AKuHAK what's missing here to be merged?

PR #457, opened 2023-08-17
ATAD: fix for Crucial BX500 SSD

AKuHAK:
Do not call proprietary SCE commands on this model as this will lock the hdd.


uyjulian (approved):
lgtm

PR #459, opened 2023-09-01
Add defragmentation write support

rickgaiser:
Add defragmentation write support. This can be used for emulating a memory card (VMC), or a HDD, in neutrino/OPL.


uyjulian (approved):
lgtm

Issue #460, opened 2023-09-03
MIPTBP1 and 2 incorrect bitmask


crypt4489:
In common/include/gs_gp.h, MIPTBP1 and 2 have an incorrect bitmask for the addresses of the mip levels. It should mask the lower 14 bits not 10 bits as it does now.

rickgaiser:
https://github.com/ps2dev/ps2sdk/blob/a8e58fdeea62efb4ee803e3bc2c89ab8bc0305e2/common/include/gs_gp.h#L223-L231

You're right, can you make a PR for changing this?

crypt4489:
@rickgaiser see #461

PR #461, opened 2023-09-04
Fix MIPTBP addresses bitmask

crypt4489:
For #460, this updates the bitmask for the addresses in MIPTBP


rickgaiser:
Thanks, merging!

PR #462, opened 2023-11-25
timer: Fix TimerBusClock2USec

### LINKED ISSUE #416: `TimerBusClock2USec` not reporting microseconds properly
Been toying around with ps2sdk to port a modern C++ project. That one uses `steady_clock` to count time and I noticed that sometimes, value returned by that would sometimes be smaller than a value returned previously. 

I identified the problem to be caused by a bug in `TimerBusClock2USec`: Seems `pUsec` is the amount of clocks instead of actual microseconds. 

Potentially offending line: https://github.com/ps2dev/ps2sdk/blob/master/ee/kernel/include/timer_alarm.h#L70. 

This is used by `gettimeofday` to fill in the `tv_usec` member and that causes it to have values outside the [0, 999999] range.

I found that problem last month, so, details are a bit hazy and I could be missing something else. I might try fixing it myself but that might take a long time, so, I just wanted to report it :)

Lemme know if you need more details!

Thanks!

Ziemas:
Should fix #416

### DIFF
--- a/ee/kernel/include/timer_alarm.h
+++ b/ee/kernel/include/timer_alarm.h
@@ -67,7 +67,7 @@ static inline u64 Sec2TimerBusClock(u64 sec) {
 static inline void TimerBusClock2USec (u64 ulClock, u32 *pSec, u32 *pUsec) {
     u64 sec = ulClock/kBUSCLK;
     *pSec = (u32)sec;
-    *pUsec = ulClock - (sec * kBUSCLK);
+    *pUsec = (1000000 * (ulClock - (sec * kBUSCLK))) / kBUSCLK;
 }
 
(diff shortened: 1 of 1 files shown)


Ziemas:
Tested this now and it seems to work fine.

fjtrujy (approved):
I’m fine wit this change, it makes sense to me.
Anyway, I will wait till @davidgfnet double-check it, because he is the author of this.

Cheers

uyjulian:
Validated against my notes. This is fine, merging

davidgfnet:
Yup that's correct, it was returning clocks instead of usecs. Not sure what happened there :P
Thank you for the fix! Sorry I forgot about this, too much email I guess.

PR #463, opened 2023-11-27
Using more functionality from newlib and clean-up

fjtrujy:
## Description
The main scope of this PR is to remove several helpers we created for open/close dir and use better what's it is provided by newlib.

It will also bring the usage of other functions such as `rewinddir` and some others.

This PR requires to be merged first: https://github.com/ps2dev/newlib/pull/8

The implementation of this PR follows the same approach that PSP did.

I have created an example to check that `readdir` keeps working, however could be nice to test some other examples.
I will try to add more examples later
Screenshot 2023-11-27 at 19 51 56

As a summary the new functions being used from `newlib` are (they belong to the flag `HAVE_OPENDIR`):
- `closedir` https://man7.org/linux/man-pages/man3/closedir.3.html
- `dirfd` https://man7.org/linux/man-pages/man3/dirfd.3.html
- `ftw` https://man7.org/linux/man-pages/man3/ftw.3.html
- `nftw` https://www.man7.org/linux/man-pages/man3/nftw.3p.html
- `opendir` https://man7.org/linux/man-pages/man3/opendir.3.html
- `readdir` https://man7.org/linux/man-pages/man3/readdir.3.html
- `rewinddir` https://man7.org/linux/man-pages/man3/rewinddir.3.html
- `scandir` https://man7.org/linux/man-pages/man3/scandir.3.html
- `alphasort` https://linux.die.net/man/3/alphasort
- `seekdir` https://man7.org/linux/man-pages/man3/seekdir.3.html
- `telldir` https://man7.org/linux/man-pages/man3/telldir.3.html

That also requires to implement in our `glue.c`:
- `getdents` https://man7.org/linux/man-pages/man2/getdents.2.html

The `rename` function is being redirected to the proper IOP `rename` functions, it was previously done through a `unlink` and `link` process.

Some other functions, the ones for the flags, `_NO_POSIX_SPAWN` `_NO_EXECVE` `NO_EXEC`, have been skipped as they are;t possible to use (so far).

As a summary, with this PR, we have made a good step in approaching a more Linux and POSIX-friendly SO (maybe in the future we can even use pipe/sockets for IO functions) and we have removed some unnecessary helpers functions created for `fileIO` and `fileXio`.

Cheers.

### DIFF
--- a/ee/libcglue/src/fdman.c
+++ b/ee/libcglue/src/fdman.c
@@ -1,0 +1,161 @@
+/*
+# _____     ___ ____     ___ ____
+#  ____|   |    ____|   |        | |____|
+# |     ___|   |____ ___|    ____| |    \    PS2DEV Open Source Project.
+#-----------------------------------------------------------------------
+# (C)2001, Gustavo Scotti (gustavo@scotti.com)
+# (c) 2003 Marcus R. Brown <mrbrown@0xd6.org>
+# (c) 2023 Francisco Javier Trujillo Mata <fjtrujy@gmail.com>
+# Licenced under Academic Free License version 2.0
+# Review ps2sdk README & LICENSE files for further details.
+*/
+
+/**
+ * @file
+ * fdman.c - Manager for fd.
+ */
+
+#include <string.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <unistd.h>
+
+#include <kernel.h>
+#include "fdman.h"
+
+#ifdef F___fdman_sema
+int __fdman_sema = -1;
+#else
+extern int __fdman_sema;
+#endif
+
+#ifdef F___descriptor_data_pool
+__descriptormap_type  __descriptor_data_pool[__FILENO_MAX];
+#else
+extern __descriptormap_type  __descriptor_data_pool[__FILENO_MAX];
+#endif
+
+#ifdef F___descriptormap
+__descriptormap_type *__descriptormap[__FILENO_MAX];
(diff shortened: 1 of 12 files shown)


uyjulian (approved):
Code looks good, looks reasonable, good direction

fjtrujy:
I have made some changes and also updated the example using `rewinddir`

Screenshot 2023-11-28 at 21 49 04

fjtrujy:
I have also executed the regress example and I have checked that everything looks to be working fine

Screenshot 2023-11-28 at 23 02 52

ee/libcglue/src/fdman.c
```diff
@@ -0,0 +1,161 @@
...
+{
+	ee_sema_t sema;
+
+	/* Initialize mutex */
+	sema.init_count      = 1;
+    sema.max_count       = 1;
+    sema.option          = 0;
+	__fdman_sema = CreateSema(&sema);
+	if (__fdman_sema < 0) {
+		abort();
```
fjtrujy, line 56: not really sure what to do here....

uyjulian: This is an unusual case, so leaving this as is should be fine.

Semaphore count should be low near the start of execution so shouldn't run out of semaphores here

fjtrujy: Súper!

rickgaiser (approved):
Looks good to me. Great work for the next step in the right direction!

PR #464, opened 2023-11-30
Apply normalized to all functions

fjtrujy:
## Description
This PR is to improve the usage of relative paths in all the POSIX functions. The same approach is also used by PSP.

More tests have been added to the regress sample

The remove file function has also been fixed, as newlib is using `_unlink` when removing a file.

Cheers.


uyjulian:
Lgtm

PR #467, opened 2023-12-03
feature: Add support for BDM and iomanX backing store for apa driver

uyjulian:
Adds support for BDM and iomanX backing store for apa driver.


israpps:
Now I'm curious?

What's the advantage/effect of this?🤔

PR #468, opened 2023-12-04
Fix relative paths

fjtrujy:
## Descritpion
This PR allow to have unit driver with uppercase values, and also it now supports paths without `/` after the `:`, like `drive:folder/file.something`

Cheers.


uyjulian (approved):
Lgtm

PR #470, opened 2023-12-06
Rework EE timer to more closely match official behavior

uyjulian:
For cohesiveness purposes.

### DIFF
--- a/ee/kernel/Makefile
+++ b/ee/kernel/Makefile
@@ -102,9 +102,55 @@ EXIT_OBJS += _exit_internals.o SetArg.o Exit.o ExecPS2.o LoadExecPS2.o ExecOSD.o
 
 ### Timer objects
 
-TIMER_OBJS = __time_internals.o StartTimerSystemTime.o StopTimerSystemTime.o iGetTimerSystemTime.o GetTimerSystemTime.o cpu_ticks.o
-TIMER_ALARM_OBJS = TimerAlarmInternals.o __Timer2Resched.o InitTimerAlarm.o DeinitTimerAlarm.o InitializeTimerAlarm.o \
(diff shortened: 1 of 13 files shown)


uyjulian:
nanosleep_test, timer_alarm, timer_alarm_sleep, and timer_alarm_stop have been fixed and tested working.

fjtrujy:
@davidgfnet I think you should take a look at this, as you were one of the original EE timers implementation

davidgfnet:
Lemme give it a read tonight, thanks for the PR!

davidgfnet:
I see you use TIM2 for the alarms, is that correct? We should probably indicate that, since homebrews might be using tim2 already for other purposes and collide with it.
I'd love to see some comments here and there. I know I'm totally biased since I wrote the previous code, but I feel that code was readable and understandable, whereas yours is much harder to read/follow :) 
Is the intent to reimplement a bug-compatible version of whatever sony's sdk shipped?

uyjulian:
TIM2 was also used by the old code, and libraries like psgl have already been updated to handle this case and disable the usage of TIM2. (But they should be reworked to use the timer code instead of direct accesses to registers such that they can share the timer)  


Mainly this was done because I had issue with conflicting `SetTimerAlarm` definitions.

davidgfnet:
Gotcha! Yeah I noticed that the old sony sdk was using TIM2 but I went for 1/0 so that TIM2 could be used by users. It's a bit arbitrary to be honest.
The code quality is a bit meh (no offense!), I just find many comments missing for some not-so-obvious code paths. Does this code come from another ps2 project or something?
Thanks!

uyjulian:
This code mainly comes from my libosd reversing project. I plan to clean it up in another pass

fjtrujy:
Please @uyjulian don't forget to clean it up, and add some comments and so on, in a near future

PR #471, opened 2023-12-06
Improve mixing usage `fileio` and `fileXio`

fjtrujy:
## Description
The main scope of this PR is to avoid having unused functions within the embedded ELF, unfortunately, the linker and compiler aren't that clever and require some help to make it work better.
More concretely most of the unused functions belong to the library used for performing IO operations, either `fileio` or `fileXio`. Having both in the binary first makes the a waste of space, and secondly could deal with problems if you start mixing by mistake `fd` from/to `fileio`/`filexio`.

How I did it:
- `rom0_info` contains some information about the `ROM0` included in the console, it was always using `fileio` for accessing the `ROM0` content, now, I have created a possibility to inject via driver the functions used to deal with IO operations.
- `osd_config` used functions from `rom0_info`, so the same strategy was created here.
- There was something wrong additionally in `osd_config`, it was using some `_libcglue_timezone_update` functions, and it shouldn't, because `libkernel.a` shouldn't have any kind of dependency, for still allowing a way of using it without `newlib`. In order to facilitate the usage of these logic to the final user, I have created 2 new functions in `ps2sdkapi.h` file, `ps2sdk_setTimezone` and `ps2sdk_setDaylightSaving` which automatically will use the proper IO library and the expected callback to `_libcglue_timezone_update`.
- `_libcglue_timezone_update` used those `osd_config` functions, so now it injects the driver with the proper IO functions. Additionally, the logic for setting the timezone was updated having as reference how `PSP` did.
- The way of setting the `_ps2sdk_XXX` functions, now have changed. Previously we set by default `fileIO` functions, and then later on, it was replaced with `fileXio` functions. Now, however, they have `NULL` value as default, and then `weak` functions called `__set_ps2sdk_XXX` pointing to `fileIO` have been created; then `strong` functions pointing to `fileXio` have been created, in the way that as soon as the user link the binary with `fileXio` automatically all the `fileio` functions will be removed from the binary. All these functions must be called before the `_ps2sd_XXX` is used. This also require to do an extra trick to make it work, https://stackoverflow.com/a/22178564
- `fileXio` EE library has been improved, it now has a specific compilation per function, and it helps to link and strip unused symbols. Other not needed functions were removed.
- A specific sample to check the timezone has been created.
- A specific sample to check `fileXio` usage has been created.
- Other minor clean/up changes.

Results:
- I have used my `filesystem` sample from `ps2_drivers` https://github.com/fjtrujy/ps2_drivers/tree/main/samples/filesystem_sample
```sh
-rwxr-xr-x  1 fjtrujy  staff  535656 Dec  6 22:06 filesystem_sample.elf
-rw-r--r--@ 1 fjtrujy  staff  549524 Dec  6 21:05 filesystem_sample_old.elf
-rw-r--r--  1 fjtrujy  staff  165720 Dec  6 22:09 filesystem_sample_old_packed.elf
-rwxr-xr-x  1 fjtrujy  staff  462900 Dec  6 22:08 filesystem_sample_old_strip.elf
-rw-r--r--  1 fjtrujy  staff  163144 Dec  6 22:10 filesystem_sample_packed.elf
-rwxr-xr-x  1 fjtrujy  staff  451508 Dec  6 22:08 filesystem_sample_strip.elf
```
It reduces `15KB` when the binary contains debug symbols, 11KB when the binary is stripped, and 2.6KB when the binary is packed.

- Let's check now OPL (I have also added `EE_CFLAGS += -fdata-sections -ffunction-sections -flto` and `EE_LDFLAGS += -fdata-sections -ffunction-sections -flto -Wl,--gc-sections`) flags:
```
WIP: I will try to test it tomorrow :(
```

Finally, I would like some of you to test those improvements, to fully assure that everything is working as expected. Especially OPL which is the most critical piece of software we actually support.

### DIFF
--- a/ee/libcglue/include/file-advanced.h
+++ b/ee/libcglue/include/file-advanced.h
@@ -1,0 +1,92 @@
+/*
+# _____     ___ ____     ___ ____
+#  ____|   |    ____|   |        | |____|
+# |     ___|   |____ ___|    ____| |    \    PS2DEV Open Source Project.
+#-----------------------------------------------------------------------
+# Copyright 2001-2004, ps2dev - http://www.ps2dev.org
+# Licenced under Academic Free License version 2.0
+# Review ps2sdk README & LICENSE files for further details.
+*/
+
+#ifndef __FILE_ADVANCED_H__
+#define __FILE_ADVANCED_H__
+
+#define NEWLIB_PORT_AWARE
+#include <fileio.h>
+#include <fileXio_rpc.h>
+#include "iox_stat.h"
+
+extern uint8_t __usingFileXIO;
+
+
+static inline void swapToFileXio() {
+    fileXioInit();
+    __usingFileXIO = 1;
+}
+
+#define FILE_SWAP_PS2SDK_FUNCTIONS() \
+    uint8_t __usingFileXIO = 0; \
+    int __fileXioGetstatHelper(const char *path, struct stat *buf); \
+    int __fileXioDreadHelper(int fd, struct dirent *dir); \
+    int __fioOpenHelper(const char* path, int flags, ...); \
+    off64_t __fioLseek64Helper(int fd, off64_t offset, int whence); \
+    int __fioRenameHelper(const char *oldpath, const char *newpath); \
+    int __fioMkdirHelper(const char *path, int mode); \
+    int __fioGetstatHelper(const char *path, struct stat *buf); \
+    int __fioReadlinkHelper(const char *path, char *buf, size_t bufsize); \
+    int __fioSymlinkHelper(const char *oldpath, const char *newpath); \
+    int __fioDreadHelper(int fd, struct dirent *dir); \
+    void _predefined_ps2sdk_close() { \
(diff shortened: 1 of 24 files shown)


uyjulian (approved):
These set of changes are in a good direction.

LGTM

ee/libcglue/include/file-advanced.h
fjtrujy, line 1: I think this header shouldn't be here, but I would like to hear your opinions... @rickgaiser and @uyjulian

uyjulian: Maybe `ps2sdk-libcglue.h`?

ee/libcglue/src/fdman.h
```diff
@@ -14,7 +14,7 @@
 
 #define _FDMAN_H_
 
-#define __FILENO_MAX 1024
+#define __FILENO_MAX 64
```
fjtrujy, line 17: This is crucial to save RAM, what do you think should be a good number here?
I would say that 64 is fair enough....

PR #472, opened 2023-12-08
Include compressed DWARF2 debug info by default for EE and IOP buildsystem

uyjulian:
Allows `addr2line` to output line number and file path.
Will increase size of libraries and unstripped executables, but stripped and compressed executables should remain same size

### DIFF
--- a/ee/Rules.make
+++ b/ee/Rules.make
@@ -35,18 +35,21 @@ EE_OPTFLAGS ?= -O2
 # Warning compiler flags
 EE_WARNFLAGS ?= -Wall -Werror
 
+# Debug information flags
+EE_DBGINFOFLAGS ?= -gdwarf-2 -gz
+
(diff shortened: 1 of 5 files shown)


fjtrujy:
Isn't this PR going to include by default all the IRX with that info?
Shouldn't we create 2 different IRX binaries? With and without this debug info? like `fileXio.irx` and `fileXio_debug.irx`?

uyjulian:
The IRX, when linked, will not contain the debug info. It is just the object files (currently) that contain the debug info.

(Once I get the elf2irx tool working and integrated, the elf files will contain the debug info while the irx files will not)

fjtrujy:
> The IRX, when linked, will not contain the debug info. It is just the object files (currently) that contain the debug info.
> 
> (Once I get the elf2irx tool working and integrated, the elf files will contain the debug info while the irx files will not)

I see!
Nice then!!

PR #473, opened 2023-12-08
Refactor WaitSema with a timeout into a separate function

uyjulian:
Seems like multiple places are using this pattern, so refactor it into a separate function


fjtrujy (approved):
Well done!

PR #482, opened 2023-12-16
[W.I.P] Add ps2ips header on compat headers

Wolf3s:
These headers need also ps2ips.h


fjtrujy:
I think this PR doesn't make sense after https://github.com/ps2dev/ps2sdk/pull/498 right @Wolf3s ?

Wolf3s:
@fjtrujy Well sounds ok that pull request, i will test it on libsmb2 when it's merged.

PR #483, opened 2023-12-16
fix: Add -call_shared to EE assembler flags

uyjulian:
This resolves the warning "warning: linking abicalls files with non-abicalls files" when assembling EE asm files


fjtrujy:
NIce!

PR #486, opened 2023-12-19
VFAT driver merge

uyjulian:
Merge common usbhdfsd, IEEE1394_disk, and bdmfs_vfat sources


fjtrujy:
Nice, the less code the better, does it requires any additional testing?

uyjulian:
No additional testing required. (This doesn't do any modifications to bdmfs_fatfs which is being used nowadays)

PR #489, opened 2023-12-28
Fix file xio functions override

fjtrujy:
I found that ELF generated still had some symbols to fio when fileXio is embedded.
Additionally I have also found that no `IO` operations can be performed before the fileXio is initied, that's why the timezone update is postponed.

Probably there are better ways to solve this, or maybe even reverting all these changes, and put it back as it was previously.


fjtrujy:
Closing as I'm going to revert to how it was before the latest refactor

PR #490, opened 2023-12-28
EE linkfile changes

uyjulian:
`.spad` section has been added for zero-initialized or uninitialized data to be located in scratchpad region.  
`.reginfo` and `.rodata` sections have been moved around.


fjtrujy:
Should we update also the one for the `elf-loader`?

uyjulian:
Since it doesn’t rely on scratchpad, it shouldn’t need to be updated.

PR #491, opened 2023-12-28
Revert fio fileXio logic to simplify possible issues

fjtrujy:
## Description
On the latest weeks, I have been trying to avoid having within the binary code that was mixing `fio` and `fileXio`, however, that was causing several issues, so instead of trying to fix them (because maybe it wasn't even possible) I have decided to go back using the same implementation we were previously using.

Cheers and sorry


uyjulian (approved):
Lgtm. This can be reworked in a better way later

PR #492, opened 2024-01-02
Fix `rom0` access

fjtrujy:
## Description
Thanks of the testing done by @israpps , I have faced 2 different issues when trying to do IO operation with `rom0`:
1. `rom0` can't have an `/`, and the previous code was always trying to put the `/` after the `:`, like `rom0:/ROMVER` which is wrong, now it properly does `rom0:ROMVER`
2. Additionally when using a custom IO driver for accessing roninfo.c functions, we were always hardcoding the usage of FIO flags, like `FIO_O_RDONLY`, so I added a new field to the `_io_driver` struct for saving the `openFlags`.

Not sure if there are other "units" different than `rom0` that should also skip the `/` after the `:`, please tell me if you are aware of any other drive.

Cheers.

### DIFF
--- a/ee/libcglue/src/cwd.c
+++ b/ee/libcglue/src/cwd.c
@@ -149,9 +149,15 @@ int __path_absolute(const char *in, char *out, int len)
 			return -1;
 	} else if(dr > 0 && in[dr - 1] == ':') {
 		/* It starts with "drive:", so it's already absoulte, however it misses the "/" after unit */
-		strncpy(out, in, dr);
-		out[dr] = '/';
-		strncpy(out + dr + 1, in + dr, len - dr - 1);
+		/* Just do it if drive: is different than rom0:, rom1:, rom2: .. rom9: */
+		if (strncmp(in, "rom", 3) == 0 && in[3] >= '0' && in[3] <= '9' && in[4] == ':') {
+			if(!__safe_strcpy(out, in, len))
+				return -1;
+		} else {
+			strncpy(out, in, dr);
+			out[dr] = '/';
+			strncpy(out + dr + 1, in + dr, len - dr - 1);
+		}
 	} else if(in[0] == '/') {
 		/* It's absolute, but missing the drive, so use cwd's drive */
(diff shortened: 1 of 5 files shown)


israpps:
> Not sure if there are other "units" different than rom0 that should also skip the / after the :, please tell me if you are aware of any other drive.


The whole `rom` device.

We commonly use rom0 because most homebrew does not care for the rest, but there's also `rom1:` (DVD player storage ROM) and there should be other indexes, at least one or two (eg: the extra ROM chip on the mainland china ps2 models)

ee/libcglue/src/cwd.c
```diff
@@ -149,9 +149,15 @@ int __path_absolute(const char *in, char *out, int len)
 			return -1;
 	} else if(dr > 0 && in[dr - 1] == ':') {
 		/* It starts with "drive:", so it's already absoulte, however it misses the "/" after unit */
-		strncpy(out, in, dr);
-		out[dr] = '/';
-		strncpy(out + dr + 1, in + dr, len - dr - 1);
+		/* Just do it if drive: is different than rom0:, rom0 doesn't expect to have an slask */
+		if(strncmp(in, "rom0:", 5) == 0) {
```
uyjulian, line 153: Should also handle the case of `rom1`, `rom2`, ...

fjtrujy: I have updated PR from `rom0` to `rom9`

fjtrujy:
> > Not sure if there are other "units" different than rom0 that should also skip the / after the :, please tell me if you are aware of any other drive.
> 
> The whole `rom` device.
> 
> We commonly use rom0 because most homebrew does not care for the rest, but there's also `rom1:` (DVD player storage ROM, only accesible after uploading `rom0:ADDDRV` to IOP) and there should be other indexes, at least one or two (eg: the extra ROM chip on the mainland china ps2 models)

Thanks, I have updated PR to support from `rom0` to `rom9`

uyjulian (approved):
Lgtm

PR #496, opened 2024-01-05
Fix usage of `CDROM` paths

fjtrujy:
## Description
When normalized path, I'm checking if `drive` requires to have `/`, `\` or none for instance `mc0`, `cdrom` or `rom`

I have used the example done on OPL to check if it works as expected:
https://github.com/ps2homebrew/Open-PS2-Loader/blob/master/labs/opltestiso/ee/main.c

Screenshot 2024-01-05 at 17 38 57

Cheers.

### DIFF
--- a/ee/libcglue/src/cwd.c
+++ b/ee/libcglue/src/cwd.c
@@ -76,30 +76,12 @@ char *getcwd(char *buf, size_t size)
 		return NULL;
 	}
 
-	strcpy(buf, __cwd);
+	strncpy(buf, __cwd, size);
 	return buf;
 }
 #endif
 
 #ifdef F___path_absolute
-/* Like strcpy, but returns 0 if the string doesn't fit */
-static int __safe_strcpy(char *out, const char *in, int maxlen)
-{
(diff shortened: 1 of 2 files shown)


uyjulian:
Hmm… I think it would be better to normalize paths.
Detect the device name, then change all occurrences of `/` or `\` to the correct slash (`/` or `\`) depending on the device.

AKuHAK:
This will only fix Access to the root files, files in subdirectories still will not be accessible.

fjtrujy:
So now do you expect developer to use `/` even for the `cdrom` unit?

uyjulian:
Should be able to either use `/` or `\`, then code corrects it to the correct slash direction for device

fjtrujy:
> Should be able to either use `/` or `\`, then code corrects it to the correct slash direction for device

But I think that previously we didn’t support either, It could be nice to do… I will try it

fjtrujy:
I have updated the PR, but I couldn't make further tests.

Cheers

uyjulian (approved):
Code lgtm

PR #502, opened 2024-01-16
Buildsystem improvements 240116

uyjulian:
The major change is that this fixes the race condition when installing samples with multiple jobs


.github/workflows/compilation.yml
fjtrujy, line 1: Shouldn't you also to put the `$(getconf _NPROCESSORS_ONLN)` instead `$(getconf _NPROCESSORS_ONLN)`

fjtrujy: And the same in the `dockerfile`

uyjulian: They are already POSIX shell expansion, so I don't need to do anything with that.

fjtrujy: Sorry I wrote it wrong previously, I wanted to say that we still have hardcoded `-j 1` in the workflow and docker file

uyjulian: An. Yeah I could do that

PR #508, opened 2024-01-18
[RFC/WIP] IOP: Provide crt0 with global constructor support

Ziemas:
In order to make it easier to use C++ on the IOP we can provide a crt0 with support for global constructors.

Is there anything else we should provide here? Or a reason to not provide it at all?

I know there would need to be additional things required to support threadsafe statics etc. But I'd be inclined to suggest just disabling anything that requires extra support. (i.e. building with  `-fno-exceptions -fno-rtti -fno-threadsafe-statics`)

### DIFF
--- a/iop/startup/Makefile
+++ b/iop/startup/Makefile
@@ -6,15 +6,20 @@
 # Licenced under Academic Free License version 2.0
 # Review ps2sdk README & LICENSE files for further details.
 
-all:
+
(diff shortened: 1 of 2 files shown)


rickgaiser:
lgtm

So if we want to make a C++ IRX now we need to include this crt0.o, and use `main` instead of `_start`. 

And the standard IRX development will stay at 'C' and use `_start`.

PR #510, opened 2024-01-18
add romimg to tools & recipe for IOPRP Images

israpps:
saves time if developer has to deal with ioprp images.
Only thing left is to include `libiopreboot`.


rickgaiser:
That's a nice additional tool. It's made by @sp193 , and updated by you to work under other os'es, am I correct?

israpps:
> That's a nice additional tool. It's made by @sp193 , and updated by you to work under other os'es, am I correct?

Only thing I did was:
- fix on Linux
- change the command line switches from windows style to Linux style (`/c` -> `-c`)

PR #513, opened 2024-01-21
Fix OPL and remove bin2s and bin2o

rickgaiser:
This PR reverts the #483 becouse of the following issues:
https://github.com/ps2homebrew/Open-PS2-Loader/issues/1142
https://github.com/ps2homebrew/Open-PS2-Loader/issues/1156
https://github.com/ps2homebrew/Open-PS2-Loader/issues/1164

The reason for #483 was to fix warnings from .s files, generated by our own bin2s tool. A better fix for this issue is to stop using the bin2s tool, and use the much better bin2c tool.
Note that from the bin2c tool generated .c file, you can still have the toolchain generate a perfect .s or .o file.

So this PR also removes the complex bin2s and bin2o tools, in favor of using only bin2c.


fjtrujy (approved):
Better to simplify paths! Good job!

AKuHAK:
> Note that from the bin2c tool generated .c file, you can still have the toolchain generate a perfect .s or .o file.

Just curious, how to achieve that?

rickgaiser:
`gcc -S file.c` will generate a .s file from a .c file
`gcc -c file.c` will generate a .o file from a .c file

You will need to use the right gcc (like mipsel**-**-gcc) and the flags you like, such as the LTO flags.

PR #515, opened 2024-01-23
Cleanup BDM folders

Wolf3s:


iop/fs/Makefile
```diff
@@ -10,8 +10,8 @@ SUBDIRS = \
 	adddrv \
 	addrom2 \
 	bdm \
-	bdmfs_fatfs \
```
fjtrujy, line 13: Shouldn't we just have here the `bdm` and then inside the `bdm` folder to compile each of the subfolders?

Wolf3s: Hey there i was thinking that the irx compiled would be the same as bdmfs_fatfs, i plan to find a fix to this one or rename back to `bdmfs_fatfs`

Wolf3s: @fjtrujy fixed: https://github.com/ps2dev/ps2sdk/pull/515/commits/a59626e385c28f597dcced2bbb0c3938ee0aa6d2

fjtrujy: Maybe I didn't explain properly before, I wanted to say that I would expect to have in this file just "bdm" right now we still have `bdm/bdm_fatfs` and `bdm/bdm_vfat`.
Those "subfolders" compilation should be done in the `bdm/Makefile` file...

Wolf3s: > Maybe I didn't explain properly before, I wanted to say that I would expect to have in this file just "bdm" right now we still have `bdm/bdm_fatfs` and `bdm/bdm_vfat`. Those "subfolders" compilation should be done in the `bdm/Makefile` file...

I think it shoudln't affect the subfolders compilation, but i'm thinking that you want to remove these `bdm/bdm_fatfs` and `bdm/bdm_vfat` make comands and add only bdm. Sounds fine to me.

rickgaiser:
Why move these modules and rename the folders? The folder name matches the module name.

Wolf3s:
> Why move these modules and rename the folders? The folder name matches the module name.

Moved in way better now.

rickgaiser:
It's just 3 modules in 3 folders. No need to nest them. Currently all folders have the same name as the modules. This integrates nicely into the ps2sdk makefiles. The nested structure you propose puts all 3 modules into a folder that does **not** match the module name.

Besides personal preference (I like them where they are, you like them moved), I think it's a bad idea to move entire folders without reason, becouse it also makes the git workflow for developers difficult. Changes will be difficult to rebase, difficult to track history, etc.

PR #519, opened 2024-01-28
[Discussion] Add SMB2MAN

Wolf3s:


iop/network/smb2man/src/COPYING
```diff
@@ -0,0 +1,24 @@
...
+
+
+The lib and include directories
+===============================
+The smb2 client library itself, i.e. the lib and include directories,
+is licenced under GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+A copy of LGPL 2.1 is included in the file LICENCE-LGPL-2.1.txt but can also be downloaded from http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
```
fjtrujy, line 11: I think this specific license can't be included here...

Wolf3s: Could use libsmb2 as external then?

Wolf3s: Because if use the code as internal it should not conflict inside.

fjtrujy: I’m not expert here, but I would say that this is not even possible to be external…

Wolf3s: I will close it, there's no chance of merge it, if anyone wants run smb2man on ps2sdk it will be hard to do it properly without the conflicts of L-GPL

PR #523, opened 2024-02-06
Add null pointer checks on network mailbox code

uyjulian:


DanielSant0s:
apparently network is working properly now, so it appear to be fixed.

PR #525, opened 2024-02-06
Check if _fstat is existent for stat operation

uyjulian:
Fixes printf buffering


DanielSant0s:
seems to work, tested with some minimal and complex samples and it appear to be fixed.

fjtrujy (approved):

PR #526, opened 2024-02-06
feat: update and restore ERL library to the current SDK

DanielSant0s:
This PR restores the runtime relocatable and linkable ELFs functionality provided by the ERL library. The biggest problem was the fact that the current GCC defaults to the "RELA" (relocation with addend) section type, which was not officially supported in the library, so what I did was basically add support for this type of section.


uyjulian (approved):
Code lgtm

PR #534, opened 2024-02-27
add dogbait module

israpps:
untested yet. awaiting confirmation on real hardware

did some tests on emulator, the cmd is called once per minute

### DIFF
--- a/iop/system2x6/dogbait/README.md
+++ b/iop/system2x6/dogbait/README.md
@@ -1,0 +1,18 @@
+# Dogbait
+
+This module provides a thread to trick the namco arcade system 246/256 watchdog, to avoit it from shutting down the machine if security dongle is not connected for some time.
+
+its implementation could be considered a fussion of `rom0:LED` and `rom0:DAEMON` (without dongle spamming)
+
+## How to use this module in your program
+
+Use `SifLoadStartModule` or `LoadModuleBuffer` directly.
+
+due to the purpose of this module, checking if the module loaded successfully and stayed resident on IOP is important
+
+like this:
+```c
+int ret, id;
+id = LoadModuleBuffer(dogbait_irx, size_dogbait_irx, 0, NULL, &ret);
+if (id<0 || ret == 1) we_have_an_error();
+```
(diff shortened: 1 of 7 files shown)


iop/system2x6/dogbait/README.md
```diff
@@ -0,0 +1,18 @@
+# Dogbait
+
+This module provides a thread to trick the namco arcade system 246/256 watchdog, to avoit it from shutting down the machine if security dongle is not connected for some time.
+
+its implementation could be considered a fussion of `rom0:LED` and `rom0:DAEMON` (without dongle spamming)
```
ijacquez, line 5: Spelling mistake: "fussion" should be "fusion".

sp193:
Out of curiosity, what happens if you boot the IOP without those modules (the DAEMON and LED, if I am not mistaken)?

israpps:
> Out of curiosity, what happens if you boot the IOP without those modules (the DAEMON and LED, if I am not mistaken)?

LED calls the blue led command directly (because arcade CDVDMAN has the export for that stubbed)

And DAEMON just calls mcdetectcard2 from DONGLEMAN once per minute

sp193:
I see. So if you hypothetically booted the IOP without DAEMON by specifying your own IOPBTCONF file, the System 246 will not be shut down?

israpps:
> I see. So if you hypothetically booted the IOP without DAEMON by specifying your own IOPBTCONF file, the System 246 will not be shut down?

Actually, DAEMON and LED are commented on the IOPBTCONF

According to krHACKen. SYSCON watchdog sets a timer. when consumed, a console poweroff is performed. And the blue led control Mechacon ncmd is the key to reset the watchdog timer

israpps:
closing because krHACKen confirmed this is not helping to solve the issue.

`rom0:DAEMON` calls DONGLEMAN function mcdetectcard2 every minute.

I dont know, but somewhere, a command is issued to the mecachon so it resets the syscon arcade watchdog

israpps:
I used the wrong mecha command.

PR #535, opened 2024-02-29
crt0: Fix bss init routine

F0bes:
There was a severe issue where the bss segment was not properly being zeroed.
This caused major issues when using ps2link reset. IE: graph vram allocations not being reset, or `setenv` crashing after a soft reset. 

I originally thought these were two different issues, but newlib uses a static local variable to track if it has previously allocated https://github.com/ps2dev/newlib/blob/ee-v4.4.0/newlib/libc/stdlib/setenv_r.c#L56
Obviously, if you soft reset the system without marking this flag as zero, the next time it tries to use this no-longer-allocated memory, you would get a crash.

TLDR: This fixes globals not being set to zero, which in turn fixes probably that bug you've been chasing for weeks.

### DIFF
--- a/ee/startup/src/crt0.c
+++ b/ee/startup/src/crt0.c
@@ -43,21 +43,21 @@ static struct sargs_start *args_start;
 void __start(struct sargs_start *pargs)
 {
     asm volatile(
-   		"# Clear bss area"
-   		"la   $2, _fbss"
-   		"la   $3, _end"
-		"1:"
-   		"sltu   $1, $2, $3"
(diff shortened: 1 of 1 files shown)


uyjulian:
Now that's a fun issue.

davidmalcolm:
FWIW I've filed [this RFE](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115496) against GCC to perhaps add a compile-time warning about such mistakes.

F0bes:
Good idea! As we've both witnessed, this can cause quite a headache :^)

PR #536, opened 2024-02-29
crt0: use noat for __start routine

F0bes:
Accidentally introduced a "Warning: used $at without ".set noat"" when building with the sdk
Woops!


F0bes:
Hmm, now the CI fails. I'll try to investigate. It builds fine here..

PR #537, opened 2024-03-01
add DONGLEMAN module

israpps:
# Purpose:
- Support COH-H10020 Security Dongles with homebrew MCMAN
- Read COH-H10020 and SCPH-10020 simultaneously

## TODO:
- [ ] get confirmation that it works on real hardware
- [ ] Check if arcade SIO2MAN has any difference
- [ ] Check if arcade MCSERV has any difference

## Optional:
- [ ] fix semaphore feature if dogbait does not work

### notes:
- arcade SIO2MAN diffs:
  + version matches XSIO2MAN v2
  + allows thread priority change via args (param seems to be: `thpri=new_priority`)


israpps:
homebrew xmcman had a different implementation of mcman_getcnum, however, no combination of port & slot can cause the `0xf` fixed return value of dongleman:

Port | Slot | final cnum
---- | ---- | ----- |
`0` | `0` | `0x0`
`0` | `1` | `0x1`
`0` | `2` | `0x2`
`0` | `3` | `0x3`
`1` | `0` | `0x8`
`1` | `1` | `0x9`
`1` | `2` | `0xa`
`1` | `3` | `0xb`

PR #550, opened 2024-03-06
kernel: Add COP0 register enum

F0bes:
Makes using GetCop0 and SetCop0 easier

`ee_cop0_defs.h` exists, but the defines are intended to be used in assembly, rather than passed as an argument.


fjtrujy:
Shouldn’t be better if we set directly values in each section, maybe is cleaner, anyway I don’t have against this PR

F0bes:
> Shouldn’t be better if we set directly values in each section, maybe is cleaner, anyway I don’t have against this PR

Do you mean in a different header file, or on a different line in kernel.h?

PR #552, opened 2024-03-08
add IOP type checker macros

israpps:
This allows us to detect DECKARD models without even opening `rom0:DECKARD`. Thought it might be useful to have it as part of the SDK headers

### DIFF
--- a/common/include/iop_regs.h
+++ b/common/include/iop_regs.h
@@ -435,4 +435,13 @@
 #define IOP_I_STAT_VB   (1 << 0)
 #define IOP_I_STAT_SBUS (1 << 1)
 
+//Borrowed from XPARAM patch applier by krat0s
+//Bit 31 of GM_IF is for the IOP type.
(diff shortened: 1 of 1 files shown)


rickgaiser:
Yes, https://github.com/ps2dev/ps2sdk/blob/master/common/include/iop_regs.h looks like the right place for this.

There's already 2 defines for the same location (I think?):
```
#define A_IOP_IRQ_CTRL 0xBF801450
#define A_IOP_SIF_1450 0xBF801450
```

`#define GM_IF ((vu32 *)0x1F801450) `

Do these point to the same registers and can they be re-used?

israpps:
@rickgaiser this should suffice

rickgaiser:
image

0x1F801450 + 0xA0000000 = 0xBF801450

it's the same address, only accessed differently:
- unmapped = TLB is not used
- uncached = caching is not used

But I'm fine with this change. Merging.

PR #555, opened 2024-03-12
Fix crt0 inline asm

Ziemas:
The registers assigned to constraints and the manually selected ones conflicted resulting in the bss loop trampling them.

Properly labeling clobbers and smarter constraint use should fix this.


F0bes (approved):

uyjulian (approved):
Lgtm

rickgaiser:
Yes, I can also confirm neutrino working again. Merging, thanks!

PR #556, opened 2024-03-12
Fixing MacOS compilation

fjtrujy:
The rooming tool is not compiling for the MacOS environment.

Please check that things still compile for Linux and Windows environments
@rickgaiser and @israpps 

Cheers.


rickgaiser:
on linux it compiles, but with the following warnings:
image

fjtrujy:
Ok, so let's merge it then...

PR #558, opened 2024-03-13
Improvements to drive cwd handling of libcglue

uyjulian:
Drive path parsing is now done in a similar way as ioman(X) does it  
`hdd` added to path exclusions


fjtrujy:
Did we agree at the end to exclude `/` from HDD?

israpps:
> Did we agree at the end to exclude `/` from HDD?

HDD paths cant have `/` 

Just remember how a sony full hdd path looks like
```
hdd0:__partition:pfs:/FILE
```

Also, adding the `/` to hdd will probably break partition creation and IOCTLs

PR #561, opened 2024-03-15
include debug fileXio on SDK distribution

israpps:
Useful if developer wants to track open,close,read,etc,etc while debugging without having to clone and compile PS2SDK again nor store prebuilt copy on repository.

Will also help a lot to see how path changes when passing through the libcglue path normalization


uyjulian (approved):
Lgtm

PR #562, opened 2024-03-16
Libcglue get drive improvement fix

uyjulian:


AKuHAK:
Maybe it will be better to negate this?
Check if path is mass, mc or pfs
use posix
then check cdrom
use windows slash
for other use : only

fjtrujy:
> Maybe it will be better to negate this? Check if path is mass, mc or pfs use posix then check cdrom use windows slash for other use : only

To be honest, I'm not sure, feel free to do whatever you consider the best

uyjulian:
I went ahead and added that change

fjtrujy:
Are you sure this is the solution we want, Imagine the day of tomorrow a new driver added, then most probably we will do supporting the “/“, which means updating that list again.
Then what is happening with HDD partitions? As we can have as many as we want, are these supporting “/“?

israpps:
The list of drives that need special treatment is smaller that the special cases. Deal with edge cases and default to `:/`

Edge cases
- `hdd`
- `rom`
- `cdrom`

Versus
- `mc`
- `pfs`
- `xfrom`
- `mass`
- `dev9x` (?)
- `smb`
- `cdfs`
- etc...

uyjulian:
Yeah I think it would be better to default to posix path handling. I changed it back to to the other behavior

PR #564, opened 2024-03-18
add libxparam

israpps:
applies compatibility modes to the DECKARD Emulator. based on the internal database of `rom0:XPARAM` on DECKARD Slim consoles and games passing custom XPARAM via the `PARAM2` entry in system.cnf

### DIFF
--- a/ee/xparam/Makefile
+++ b/ee/xparam/Makefile
@@ -2,18 +2,13 @@
 #  ____|   |    ____|   |        | |____|
 # |     ___|   |____ ___|    ____| |    \    PS2DEV Open Source Project.
 #-----------------------------------------------------------------------
-# Copyright 2001-2004, ps2dev - http://www.ps2dev.org
+# Copyright 2001-2024, ps2dev - http://www.ps2dev.org
 # Licenced under Academic Free License version 2.0
 # Review ps2sdk README & LICENSE files for further details.
 
-SUBDIRS = startup erl kernel libcglue libpthreadglue rpc debug \
-	eedebug sbv dma graph math3d \
-	packet packet2 draw libgs \
-	libvux font input inputx network iopreboot \
-	mpeg \
-	elf-loader elf-loader-nocolour \
-	xparam
+EE_OBJS = xparam.o resetparam.o erl_support.o
 
 include $(PS2SDKSRC)/Defs.make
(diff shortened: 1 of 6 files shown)


AKuHAK:
Rename krat0s to kr_ps2

fjtrujy:
@rickgaiser could you review this?

rickgaiser:
Looks good to me

israpps:
Since the XPARAM structure is stored on RAM and is accessible from either EE or IOP via simple address access. I will probably refactor this some day to drop the dependency of `rom0:XPARAM`

israpps:
@AKuHAK @fjtrujy @rickgaiser 

What do you think guys? leaving it as is or switch to[ direct RAM manipulations](https://israpps.github.io/PPC-Monitor/docs/XPARAM.html), to drop the dependency of `rom0:XPARAM` and `sifloadmodule`

fjtrujy:
The only thing that I don't like about this "module", is the fact it is starting to include IDs related to games itself.
Not sure if this should be part of the `ps2sdk` or to be a totally separate library, for instance part of `ps2sdk-ports`

israpps:
> The only thing that I don't like about this "module", is the fact it is starting to include IDs related to games itself. Not sure if this should be part of the `ps2sdk` or to be a totally separate library, for instance part of `ps2sdk-ports`

all the mentioned IDs are edge cases to make sure the library has an uniform behaviour accross all the DECKARD models.

Any ID mentioned on source code meets either one of the following points:
- Was added to XPARAM database on a newer DECKARD Model
- Was updated in later XPARAM (eg: adding another compat mode)
- The game passes its own XPARAM value via system.cnf

rickgaiser:
This is code to make a few specific games work on deckard models. It does not belong in ps2sdk.
I know the only 2 common places to put code in is limited to ps2sdk, and ps2sdk-ports. But this code belongs in neither.

You can always create a new ps2 game loader sdk library :).

israpps:
> This is code to make a few specific games work on deckard models. It does not belong in ps2sdk. I know the only 2 common places to put code in is limited to ps2sdk, and ps2sdk-ports. But this code belongs in neither.
> 
> You can always create a new ps2 game loader sdk library :).

this is not only for game loaders, any program capable of running ps2 games in any form can benefit of it. even wLaunchELF

PR #565, opened 2024-03-20
Allow manual workflow execution

israpps:
Allows to recompile SDK without commiting to the branch. Also allows forks to run workflow on other branches without commiting to main branch


uyjulian (approved):
Lgtm

PR #566, opened 2024-03-21
Implementation of loadcore module

uyjulian:


israpps:
Pretty neat!

One step closer to having a homebrew equivalent for the `IOPRPxxx.IMG`'s that shipped with og SDK

fjtrujy (approved):

PR #573, opened 2024-03-25
Module unload support

uyjulian:
Backwards compatible module unload support for pvrdrv, dvr, dvrav, dvrdv, dvrfile, dvripl, dvrmisc, apa, pfs, mcman, mcserv


israpps:
Such a good day to be alive

PR #581, opened 2024-03-27
Adding tools to CI/CD

fjtrujy:
This PR adds the compilation of tools in the available supported systems Linux, Windows, and MacOS.

In this way now we can check if we merge something that is breaking the CI, previously, we didn't face any error till by dispatching it reached the `ps2dev` repo.

Cheers


uyjulian (approved):
Lgtm

PR #584, opened 2024-03-27
[tools CI] upload built binaries and only run workflow if commiting t…

israpps:
…o `tools/` or to the CI script

- allows testing binaries built by CI
- saves workflow run time if the tools folder was not touched by the push event trigger


AKuHAK:
Not compile

fjtrujy:
> Not compile

Is not his fault... this is something else @F0bes is already working on it

PR #588, opened 2024-04-01
Add `F_SETFD` cmd to `_fcntl` function

fjtrujy:
This PR basically makes the `std::filesystem::directory_iterator` work as expected.

I have used this example:
```cpp
#include <stdio.h>
#include <filesystem>

int main(void)
{
    std::filesystem::directory_iterator dir_scanner("./");
    for (auto &entry : dir_scanner) {
        printf("%s\n", entry.path().string().c_str());
    }
    return 0;
}
```

It prints content of folder:
```
[    9.2121] dopen name host:/Users/fjtrujy/Downloads/filesystem-test-48641ffa 
[    9.2127] dopen fd = 256
[    9.2180] read/write allocate memory 4000
[    9.2182] ./CMakeFiles
[    9.2183] ./Makefile
[    9.2186] ./cmake_install.cmake
[    9.2231] ./filesystem-test.elf
[    9.2233] ./CMakeCache.txt
```

Cheers.

### DIFF
--- a/ee/libcglue/Makefile
+++ b/ee/libcglue/Makefile
@@ -90,7 +90,6 @@ GLUE_OBJS = \
 	_isatty.o \
 	chmod.o \
 	fchmod.o \
-	fchmodat.o \
 	pathconf.o \
 	fsync.o \
 	getuid.o \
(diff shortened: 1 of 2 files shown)


uyjulian (approved):
Lgtm. The POSIX.1-2008 calls can be implemented more fully in the future to make more libraries like libarchive working properly

PR #590, opened 2024-04-06
Fix usage of `clock` and `CLOCKS_PER_SEC`

fjtrujy:
## Description
After fixing the implementation of `clock()`, now it returns values in `microseconds`, so now we need to make some adjustments.

Before mergin this PR we need to merge first https://github.com/ps2dev/newlib/pull/12

### DIFF
--- a/ee/libcglue/samples/nanosleep/main.c
+++ b/ee/libcglue/samples/nanosleep/main.c
@@ -36,7 +36,7 @@ int main(int argc, char *argv[])
     struct timespec tv = {0};
     tv.tv_sec = 1;
     tv.tv_nsec = 0;
-    error_tolerance = 5; // 5 miliseconds of error tolerance
+    error_tolerance = 200; // 200 miliseconds of error tolerance
 
 #if defined(SCREEN_DEBUG)
     init_scr();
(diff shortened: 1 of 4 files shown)


ee/libcglue/src/glue.c
```diff
@@ -637,7 +637,7 @@ int _gettimeofday(struct timeval *tv, struct timezone *tz)
 
 #ifdef F__times
 clock_t _times(struct tms *buffer) {
-	clock_t clk = GetTimerSystemTime() / (kBUSCLK / CLOCKS_PER_SEC);
```
fjtrujy, line 640: Even being the same value, I think that using `CLOCKS_PER_SEC` could be "confusing"

ee/libcglue/samples/nanosleep/main.c
```diff
@@ -36,7 +36,7 @@ int main(int argc, char *argv[])
     struct timespec tv = {0};
     tv.tv_sec = 1;
     tv.tv_nsec = 0;
-    error_tolerance = 5; // 5 miliseconds of error tolerance
+    error_tolerance = 200; // 200 miliseconds of error tolerance
```
fjtrujy, line 39: There is room for improvement in the timer, we should be more accurate.

PR #593, opened 2024-04-15
Misalignment workaround for bdmfs_fatfs

uyjulian:


Ziemas:
Does fatfs flush on every store? Would it make a performance difference to split out the unaligned part of the write in `fs_write`?

uyjulian:
I think fatfs flushes when sector buffer is filled, but not sure about this.

I haven't messed with splitting `fs_write` calls.

fjtrujy:
I think we shouldn't merge this, as the misalignment is just happening in the USB, no in all the units as is implementation does.
Anyway, we should find a better solution here, as this is super slow...
Let's wait if some other come with a different solution to the issue.

uyjulian:
Closed in favor of https://github.com/ps2dev/ps2sdk/pull/599

PR #596, opened 2024-04-16
Implementation of SPEED UART driver

uyjulian:


fjtrujy (approved):

israpps:
This could be used as a nice alternative to UDPTTY on non deckard models right?

uyjulian:
> This could be used as a nice alternative to UDPTTY on non deckard models right?

Yeah, it can be used for that purpose.

israpps:
> > This could be used as a nice alternative to UDPTTY on non deckard models right?
> 
> Yeah, it can be used for that purpose.

any picture showing the pins?

uyjulian:
As far as I know, there aren't any

PR #599, opened 2024-04-16
Misalignment workaround for usbmass_bd

uyjulian:
Alternative to #593

### DIFF
--- a/iop/usb/usbmass_bd/src/scsi.c
+++ b/iop/usb/usbmass_bd/src/scsi.c
@@ -214,15 +214,34 @@ static int scsi_write(struct block_device *bd, u64 sector, const void *buffer, u
 {
     struct scsi_interface *scsi = (struct scsi_interface *)bd->priv;
     u16 sc_remaining            = count;
+    unsigned int sectorSize     = bd->sectorSize;
+    void *misalign_buffer       = NULL;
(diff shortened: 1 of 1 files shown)


iop/usb/usbmass_bd/src/scsi.c
```diff
@@ -185,26 +185,48 @@ static int scsi_read(struct block_device *bd, u64 sector, void *buffer, u16 coun
 {
     struct scsi_interface *scsi = (struct scsi_interface *)bd->priv;
     u16 sc_remaining            = count;
+    unsigned int sectorSize     = bd->sectorSize;
+    void *misalign_buffer       = NULL;
     int retries;
 
     M_DEBUG("%s: sector=0x%08x%08x, count=%d\n", __func__, U64_2XU32(&sector), count);
 
+    if (((uiptr)buffer & 3) != 0) {
```
fjtrujy, line 194: I'm actually not sure if we should do it for `read`, as it uses the `sceUsbdOpenPipe` instead of `sceUsbdOpenPipeAligned`, I would say that the proper solution should check if it has or doesn't have the `aligned` flag activated too.
Additionally, maybe we can use `sceUsbdOpenPipe` for both `read & write` as it looks to be the recommended official way...

Issue #594, opened 2024-04-15
USB driver not working well with unaligned content


fjtrujy:
👋 
I have been facing some issues with USB drivers when using `fwrite` from unaligned 4-byte buffers...
Given the next example:
```c
    // Use mass: device to write to a file
    chdir("mass:/");

    // Open a file in binary write mode
    FILE *file = fopen("dummy_file.bin", "wb");
    if (file == NULL) {
        perror("Error opening file");
        return 1;
    }

    unsigned char *dummy_content = (unsigned char *)malloc(2000);

    for (int i = 0; i < 2000; i++) {
        dummy_content[i] = (unsigned char)(i % 256); // Fill with values from 0 to 255
    }

    unsigned char *dummy_content_offset = dummy_content
    printf("Dummy content address is %p, offset address is %p\n", dummy_content, dummy_content_offset);

    // Write the dummy content to the file
    size_t toWrite = 1200;
    size_t elements_written = fwrite(dummy_content_offset, 1, toWrite, file);
    if (elements_written != toWrite) {
        perror("Error writing to file");
        fclose(file);
        free(dummy_content);
        SleepThread();
        return 1;
    }

    // Close the file
    fclose(file);
    free(dummy_content);
    printf("Dummy content written to file successfully.\n");
```

It works as expected, however, if we replace `unsigned char *dummy_content_offset = dummy_content` with `unsigned char *dummy_content_offset = dummy_content + 6` to force having an offset, it fails due to alignment issues on the USB driver.

```c
    // Use mass: device to write to a file
    chdir("mass:/");

    // Open a file in binary write mode
    FILE *file = fopen("dummy_file.bin", "wb");
    if (file == NULL) {
        perror("Error opening file");
        return 1;
    }

    unsigned char *dummy_content = (unsigned char *)malloc(2000);

    for (int i = 0; i < 2000; i++) {
        dummy_content[i] = (unsigned char)(i % 256); // Fill with values from 0 to 255
    }

    // create a pointer with 6 bits offset
    unsigned char *dummy_content_offset = dummy_content + 6;
    printf("Dummy content address is %p, offset address is %p\n", dummy_content, dummy_content_offset);

    // Write the dummy content to the file
    size_t toWrite = 1200;
    size_t elements_written = fwrite(dummy_content_offset, 1, toWrite, file);
    if (elements_written != toWrite) {
        perror("Error writing to file");
        fclose(file);
        free(dummy_content);
        SleepThread();
        return 1;
[...]

PR #604, opened 2024-04-20
Remove extra unaligned buffer as it solved using `sceSifGetOtherData`

fjtrujy:
When executing `fileXioWrite` we were using an extra buffer of 64 bytes for trying to solve 64 bytes alignment:
```
unsigned int unalignedDataLen;
unsigned char unalignedData[64];
```
Producing then 2 separate write calls, firstly for writing the `unalignedData` with `unalignedDataLen` and secondly for the remaining things.
This extra work is not needed at all as the un-alignement is going to be solved by the `SifRpcGetOtherData` call.

This PR is not fixing all the possible scenarios we can face with #594 , however it helps to mitigate the issue.

Cheers

### DIFF
--- a/common/include/fileXio.h
+++ b/common/include/fileXio.h
@@ -195,8 +195,6 @@ struct fxio_write_packet
     int fd;
     const void *buffer;
     int size;
-    unsigned int unalignedDataLen;
-    unsigned char unalignedData[64];
 };
 
 struct fxio_lseek_packet
--- a/ee/rpc/filexio/src/fileXio_rpc.c
+++ b/ee/rpc/filexio/src/fileXio_rpc.c
@@ -615,7 +615,6 @@ int fileXioRead(int fd, void *buf, int size)
 #ifdef F_fileXioWrite
 int fileXioWrite(int fd, const void *buf, int size)
 {
-	unsigned int miss;
 	int rv;
 	struct fxio_write_packet *packet=(struct fxio_write_packet*)__sbuff;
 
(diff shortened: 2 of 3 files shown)


fjtrujy:
This PR together with #599 looks to be a great solution

uyjulian (approved):
Lgtm

sp193:
Maybe I missed something, but where in SifRpcGetOtherData does it compensate for the limitations of SIF DMA and the EE cache line sizes?

fjtrujy:
> Maybe I missed something, but where in SifRpcGetOtherData does it compensate for the limitations of SIF DMA and the EE cache line sizes?

Good question, I just saw that pointer returned was already aligned, I did some tests and it looks to work right (PCSX2 and real hardware), not sure what else could be missing…

sp193:
It has been a long time, so please do double-check the facts.

What I guess you saw, is that the pointer on the IOP side is aligned. It will be aligned by the other (IOP, if the caller is EE) RPC implementation via SifRpcGetOtherData because the FILEIO implementation will use its own buffer. However, SifRpcGetOtherData itself requires the memory pointer to be aligned because it uses SifSetDma to move the data.

What I believe will happen if you remove the code:
* If the pointer is not aligned to a multiple of 16, you will get a bus error.
* If the pointer is aligned to a multiple of 16 but not 64, you could get data corruption due to the cache writeback (SifWriteBackDCache) over an unaligned address.

fjtrujy:
> It has been a long time, so please do double-check the facts.
> 
> What I guess you saw, is that the pointer on the IOP side is aligned. It will be aligned by the other (IOP, if the caller is EE) RPC implementation via SifRpcGetOtherData because the FILEIO implementation will use its own buffer to receive the data that is meant to be written. However, SifRpcGetOtherData itself requires the memory pointer to be aligned because it uses SifSendCmd (which internally uses SifSetDma) to move the data.
> 
> What I believe will happen if you remove the code:
> 
> * If the pointer is not aligned to a multiple of 16, you will get a bus error.
> * If the pointer is aligned to a multiple of 16 but not 64, you could get data corruption due to the cache writeback (SifWriteBackDCache) over an unaligned address.

So I’m pretty sure the EE pointer is not aligned to 16 bytes as I’m using the same example/code that previously was using 58 bytes on the extra mis buffer…
Anyway I will do further tests there.
Can you recommend a proper test to force these scenarios to happen?

sp193:
You could allocate memory +3 bytes, add 3 to the returned pointer and use the result to write/read data.

Even if not being a multiple of 16 works, it might not work if the address is not a multiple of 4.

By the way, the code says that SifWriteBackDCache is always used, regardless of whether the pointer is aligned or not. Since it is a cache write back, it should not corrupt data, even if the address is not aligned. I now think there is no problem unless you mix memory access modes. Such as using uncached memory access access in the adjacent memory region. I believe that this was mentioned in the Sony tips/limitations document, as it is a potential pitfall.

PR #610, opened 2024-05-08
input: Fix uninitialized use.

Ziemas:
Ran into an error for this when building with gcc14. It looks like the right thing to do.


uyjulian (approved):
Lgtm

PR #623, opened 2024-05-23
Add ps2sdkinfo tool for version info

uyjulian:


fjtrujy:
What's the proposal of this tool?

uyjulian:
Basically this tool can output version info, build date, commit id so it can be easier to identify how a program was built.

PR #624, opened 2024-05-29
[libcglue] add macro for system 2x6 conflictive code override

### LINKED ISSUE #601: [System 2x6] default behaviour of `_libcglue_init()` causes system 2x6 to freeze before reaching `main()`
adding this fixed the issue
```c
   void _libcglue_init() {}
   void _libcglue_deinit() {}
```

however, some apps may depend on whatever is inside them.

I will take a closer look when I have time

I have only tested this on emulator. but I guess real COH-H's will have same issue

israpps:
fixes #601


fjtrujy:
Solve #601

PR #625, opened 2024-05-30
add generic stub for unsupported ops on ioman(X)

israpps:
- improves readability (IMO)
- unified return value for unsupported ops (`-ENOTSUP`) making it clear from EE side that the operation that just failed is simply stubbed on the IOP


iop/kernel/include/iomanX.h
```diff
@@ -87,6 +87,11 @@ typedef struct _iomanX_iop_device {
 	struct _iomanX_iop_device_ops *ops;
 } iomanX_iop_device_t;
 
+
+#define DECL_NOT_SUPPORTED_OP() \
```
fjtrujy, line 91: instead of using `macro`s you can also use:
```c
static inline int not_supported_op (void) {return -134;}
```

israpps: You're right, I'll apply the change later.

Although, I would like to read a review of @uyjulian on this regard, specially with the DVR related IRX modules

uyjulian:
Suggest to use `ENOSYS` instead of making up a new error code `134`

fjtrujy:
> > Suggest to use `ENOSYS` instead of making up a new error code `134`
> 
> 134 corresponds to -`ENOTSUP` without including errno everywhere.

But I think this is where the issue is coming from, to include `errno` in every file where it is needed it is a good practice. In this way you don't care about the "actual value" however you care about the "meaning" `ENOSYS`.

israpps:
> > > Suggest to use `ENOSYS` instead of making up a new error code `134`
> > 
> > 
> > 134 corresponds to -`ENOTSUP` without including errno everywhere.
> 
> But I think this is where the issue is coming from, to include `errno` in every file where it is needed it is a good practice. In this way you don't care about the "actual value" however you care about the "meaning" `ENOSYS`.

I thought including errno also includes the array of descriptions that comes with it. Or that is stripped away by the compiler? If that's the case I'll correct it

fjtrujy:
> > > > Suggest to use `ENOSYS` instead of making up a new error code `134`
> > > 
> > > 
> > > 134 corresponds to -`ENOTSUP` without including errno everywhere.
> > 
> > 
> > But I think this is where the issue is coming from, to include `errno` in every file where it is needed it is a good practice. In this way you don't care about the "actual value" however you care about the "meaning" `ENOSYS`.
> 
> I thought including errno also includes the array of descriptions that comes with it. Or that is stripped away by the compiler? If that's the case I'll correct it

It shouldn’t increase binary size, it must be exactly the same, if not something else is wrong

iop/dvrp/dvripl/src/dvripl.c
```diff
@@ -200,14 +199,9 @@ int dvripl_df_ioctl2(iomanX_iop_file_t *f, int cmd, void *arg, unsigned int argl
...
 
-int dvripl_df_null()
-{
-    return -48;
-}
-
 s64 dvripl_df_null_long()
 {
-    return -48LL;
+    return -134LL;
```
fjtrujy, line 204: Why have you changed the value here?

fjtrujy (commented):
I’m ok merging this change @rickgaiser @uyjulian .
However this improvement is quite difficult to check if other devs will maintain.
Just to put an example, uyjulian has created 2 extra PRs where he adds more IOP modules, he is not using that function there…

israpps:
> Just to put an example, uyjulian has created 2 extra PRs where he adds more IOP modules, he is not using that function there…

Well of course he's not using something that has not been merged yet.😬

If this gets merged let's just propagate the change to any open PR.

uyjulian:
Including `errno.h` **won't** increase binary size as it is just preprocessor macro definitions. Basically the equivalent of copy and paste.

israpps:
> Including `errno.h` **won't** increase binary size as it is just preprocessor macro definitions. Basically the equivalent of copy and paste.

excellent. I thought it did. at least my local copy of errno.h (did not check the one on ps2sdk) had a char** buffer for the descriptions of errno values

fjtrujy:
Resolve first conflicts and then I will review it

iop/dvrp/dvripl/src/dvripl.c
israpps: -ENOTSUP. but as `s64`

israpps:
done, should b ok now

fjtrujy (approved):
It should work nice! And I don't expect any memory increase at all

israpps:
> It should work nice! And I don't expect any memory increase at all

I hope so lol

PR #626, opened 2024-05-31
Missing functions

fjtrujy:


ee/libcglue/include/sys/socket.h
```diff
@@ -82,6 +82,13 @@
 #include <tcpip.h>
 #endif
 
+/*
+ * Maximum queue length specifiable by listen(2).
+ */
+#ifndef SOMAXCONN
```
fjtrujy, line 88: This variable is defined in the same file but in the commented section with `#if 0`.

PR #629, opened 2024-06-04
Add dummy implementation for some missing net functions

fjtrujy:
## Description

In this PR we are adding dummy implementation for IOP and EE driver for:
- `gethostbyaddr`
- `sndmsg`
- `recvmsg`


uyjulian (approved):
Lgtm

PR #640, opened 2024-07-02
[src_printf]: implement `\r`

israpps:
could be useful


ee/debug/src/scr_printf.c
```diff
@@ -299,6 +299,10 @@ void scr_vprintf(const char *format, va_list opt)
                     X++;
                 }
                 break;
+            case '\r':
+                X = 0;
+                // scr_clearline(Y); //Should we clear the line?
```
fjtrujy, line 304: I think it is not necessary to clean

PR #647, opened 2024-07-10
[ppctty]: set appropiate flags for tty device

israpps:
Like UDPTTY


rickgaiser:
Setting it the same as UDPTTY should be correct. I can merge, but why is this a draft?

israpps:
> Setting it the same as UDPTTY should be correct. I can merge, but why is this a draft?

Oh, yeah, sorry

PR #648, opened 2024-07-12
convert ioplib in a standalone iop library

israpps:
Currently there are 3 IRX modules around (AFAIK) using ioplib:
- MX4SIO_BD (located here)
- SD2PSXMAN (On development)
- [neutrino's patch_freemem](https://github.com/rickgaiser/neutrino/tree/master/iop/patch_freemem)

The Idea is to stop having local copies of ioplib.h and ioplib.c on every project that requires manipulating exports of other modules.

Therefore, converting this into a library wich is part of the PS2SDK seemed like the most reasonable approach, having one implementation for all the use cases. instead of local copies


israpps:
havent tested it yet if the mx4sio driver works normally after this (although it should?)

rickgaiser (commented):
Good idea, I created this 'ioplib' some time ago, based on sources from elsewhere (OPL I think it was). Since then I've made many copies = bad practice.

For naming consistency, if you name the library `modhook` which I think covers what the library does better that `ioplib`, then I think the include header should also be named `modhook.h`. Also the function names should then begin with something like `modhook_*` or `mh_*`.

iop/sio/mx4sio_bd/src/sio2man_hook.c
```diff
@@ -97,26 +97,26 @@ static void _sio2man_unhook(iop_library_t *lib)
...
     }
 
-    ioplib_hookExportEntry(lib, 23, _23_psio2_pad_transfer_init);
-    ioplib_hookExportEntry(lib, 24, _24_psio2_mc_transfer_init);
-    ioplib_hookExportEntry(lib, 25, _25_psio2_transfer);
-    ioplib_hookExportEntry(lib, 26, _26_psio2_transfer_reset);
-    ioplib_hookExportEntry(lib, 46, _46_psio2_pad_transfer_init);
-    ioplib_hookExportEntry(lib, 47, _47_psio2_mc_transfer_init);
-    ioplib_hookExportEntry(lib, 48, _48_psio2_mtap_transfer_init);
+    modhookk_hookExportEntry(lib, 23, _23_psio2_pad_transfer_init);
```
rickgaiser, line 100: How did the CI build this succesfully? `modhookk` with double 'k'?

israpps: Lol, didn't saw that. Will recheck this later!

PR #649, opened 2024-07-15
Call `_InitSys()` before newlib initialization

fjtrujy:
## Description
The current implementation of `crt0.c` looks to be wrong as we are calling `_InitSys()` too late.

Within `_InitSys()` we are initialization timers, alarms, and some others.
In the existing implementation, we are calling `_InitSys()` after calling `libcglue_init` and the C++ constructors; this mean we couldn't use for instance timers on those specific functions.

This is why I think that `_InitSys` should be called way earlier, as it doesn't have any `newlib` requirement.

Cheers.

### DIFF
--- a/ee/startup/src/crt0.c
+++ b/ee/startup/src/crt0.c
@@ -112,12 +112,12 @@ static void _main()
     // initialize libcglue
     _libcglue_init();
 
+    // Enable interruts
+    EI();
+
     // call global constructors (weak)
     if (_init)
         _init();
-
-    // Enable interruts
-    EI();
 
     // call main
     retval = main(pa->argc, pa->argv);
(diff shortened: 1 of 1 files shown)


rickgaiser:
lgtm.

Perhaps another improvement is to move `EI();` to before calling the `_init();` / global contructors. Since constructors are part of user programs, they should be in the same ready-to-use-everything environment as `main`.

fjtrujy:
> lgtm.
> 
> Perhaps another improvement is to move `EI();` to before calling the `_init();` / global contructors. Since constructors are part of user programs, they should be in the same ready-to-use-everything environment as `main`.

Nice let me do it too

PR #650, opened 2024-07-15
Implementing profglue

fjtrujy:
## Description
This PR brings us the option to measure our code in order to find out possible bottlenecks.

 In order to make this to work there are 2 previous PRs that need to be merged:
- Enable `gprof` binary compilation from newlib: https://github.com/ps2dev/binutils-gdb/pull/5
- Add `libprofglue` as part of the standard libraries when using the flags `-pg` https://github.com/ps2dev/gcc/pull/11

## Implementation
The current implementation being used here is inspired by the existing implementation for PSP.
https://github.com/pspdev/pspsdk/tree/master/src/prof

## How to use it:
Basically, you need to compile your code using the `-pg -g` flags for compilation and linking.

Then **when exiting your app**, it will automatically generate a file called `gmon.out`. In order to interpret that file you require to use the `gprof` tool.
More info: https://sourceware.org/binutils/docs/gprof/

You can use for instance:
```
mips64r5900el-ps2-elf-gprof -b retroarch_ps2.elf gmon.out > analysis.txt
```

Where 
```
mips64r5900el-ps2-elf-gprof -b {your binary elf} gmon.out > {output file}
```

## Example
```c
#include <stdio.h>

#include <kernel.h>

int calculateNPrime(int n) {
   int i, j, count = 0;
   for (i = 2; i <= n; i++) {
      for (j = 2; j <= i; j++) {
         if (i % j == 0) {
            break;
         }
      }
      if (i == j) {
         count++;
      }
   }
   return count;

}

int dummy_function()
{
   int i;
   for (i = 0; i < 10000; i++)
   {
      printf(".");
   }
   printf("\n");
   return 0;
}

int main()
{
   printf("Hello, world!\n");
   dummy_function();  
   calculateNPrime(1000);
   printf("Goodbye, world!\n");

   SleepThread();
   return 0;
}
```

Generates:
```
Flat profile:

Each sample counts as 6.78168e-06 seconds.
  %   cumulative   self              self     total           
 time   seconds   seconds    calls  us/call  us/call  name    
 83.33      0.00     0.00        1   271.27   271.27  dummy_function
 16.67      0.00     0.00        1    54.25    54.25  calculateNPrime
  0.00      0.00     0.00        1     0.00   325.52  main

            Call graph


granularity: each sample hit covers 2 byte(s) for 2.08% of 0.00 seconds

index % time    self  children    called     name
                0.00    0.00       1/1           _ftext [2]
[1]    100.0    0.00    0.00       1         main [1]
                0.00    0.00       1/1           dummy_function [3]
                0.00    0.00       1/1           calculateNPrime [4]
-----------------------------------------------
                                                 <spontaneous>
[2]    100.0    0.00    0.00                 _ftext [2]
                0.00    0.00       1/1           main [1]
-----------------------------------------------
                0.00    0.00       1/1           main [1]
[3]     83.3    0.00    0.00       1         dummy_function [3]
-----------------------------------------------
                0.00    0.00       1/1           main [1]
[4]     16.7    0.00    0.00       1         calculateNPrime [4]
-----------------------------------------------

Index by function name

   [4] calculateNPrime         [3] dummy_function          [1] main
```

I hope this tool helps us to improve our toolchain.

Cheers.

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -29,6 +29,7 @@ jobs:
           # Create symbolink links using relative paths
           (cd $PS2DEV/ee/mips64r5900el-ps2-elf/lib && ln -sf ../../../ps2sdk/ee/lib/libcglue.a libcglue.a && cd -)
           (cd $PS2DEV/ee/mips64r5900el-ps2-elf/lib && ln -sf ../../../ps2sdk/ee/lib/libpthreadglue.a libpthreadglue.a && cd -)
+          (cd $PS2DEV/ee/mips64r5900el-ps2-elf/lib && ln -sf ../../../ps2sdk/ee/lib/libprofglue.a libprofglue.a && cd -)
           (cd $PS2DEV/ee/mips64r5900el-ps2-elf/lib && ln -sf ../../../ps2sdk/ee/lib/libkernel.a libkernel.a && cd -)
           (cd $PS2DEV/ee/mips64r5900el-ps2-elf/lib && ln -sf ../../../ps2sdk/ee/lib/libcdvd.a libcdvd.a && cd -)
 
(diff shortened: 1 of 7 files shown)


fjtrujy:
I would like @davidgfnet to take a look at this PR, thanks!

davidgfnet:
Cool stuff. Looks good to me.
As I mentioned I would remove the _busy recursive magic, since you should not ever build prof.c with "-pg", and it only adds overhead otherwise (just deleting mcount.S:20-28, 36-41 and 54-55).
Thanks man!

fjtrujy:
I have added some changes to avoid `prof.c` functions for being excluded if the are compiled with `-pg`.

Cheers

PR #655, opened 2024-08-29
Subfile module

uyjulian:


fjtrujy (approved):

AKuHAK:
@uyjulian do you have some sample code of how lo launch for example kelf from PP.TEST partition?

uyjulian:
I don't have any sample code (yet).

But basically for partition attribute area `PS2ICON3D` item `system.cnf`, following for keys `BOOT2` and `IOPRP` apply:  
* When string starts with `cdrom` or `pfs`, boot from there
* When string is exactly `NOBOOT`, fail here
* When string is exactly `PATINFO`, see below
* Otherwise, fail

For `PATINFO`:
* Read first 512 bytes of partition
* Format string: `subfile:%s%s,%08X,%08X` where first argument is device, second argument is partition, third argument is 32 bit integer at offset 48 of read data (+8 bytes if key is IOPRP), then fourth argument is 32 bit integer at offset 52 of read data (+8 bytes if key is IOPRP)
* Boot from the formatted string

PR #656, opened 2024-08-29
Eromdrv module

uyjulian:


iop/fs/eromdrv/README.md
```diff
@@ -0,0 +1,16 @@
+# Encrypted ROM driver
+
+IOP module for handling the encrypted filesystem contained in the ROM
+chip attached to DEV1.
+
+## Configurations
+
+There are multiple configurations of this library, allowing the choice of
+balancing between size, speed, and features.
```
fjtrujy, line 9: How can we choose these "configurations" I don't see anything like that in the code?

israpps: He uses this generic readme.md on every new module.

Regardless if those modules have different configurations

uyjulian: For now, there is one configuration. If I decide to add more it's just a new line in the readme and some buildsystem additions.

fjtrujy (approved):

PR #657, opened 2024-08-29
Fix GP register access for IOP on newer toolchain

uyjulian:


rickgaiser:
Was GP register usage broke? And could that be the cause of this issue?:
https://github.com/ps2homebrew/Open-PS2-Loader/issues/1286

Since the broken games for instance used a usb keyboard driver. So if our broken usbd calls that driver with the wrong GP register it could crash?

uyjulian:
Modifying of GP register in inline asm was broken:  
```
In file included from src/smap.c:4:
In function 'ChangeGP',
    inlined from 'SMAPStart' at src/smap.c:670:13:
/mnt/data/ps2src/ps2dev/ps2sdk/iop/kernel/include/defs.h:49:9: error: PIC register clobbered by 'gp' in 'asm'
   49 |         __asm__ volatile(       "move %0, $gp\n"
      |         ^~~~~~~

```

So storing of the calling module's GP register couldn't be done. So callbacks needing it were broken. It may or may not fix the issue, but I haven't checked it.  

Without this PR, the GP register was not saved/restored, so modules that didn't do their own handling of the GP register may access the incorrect offset.  
With this PR, the behavior should be corrected.

PR #664, opened 2024-10-01
[WIP/RFC] Cmake

Ziemas:
Opening a PR early in case anyone wants to help or has ideas/feedback.


fjtrujy:
It would be nice to have `cmake` in place as I think it will help make compilation faster.
What was your blocker?

Ziemas:
TBH I completely forgot about this for a while.

I don't think I was blocked on anything in particular, it's just kinda boring/annoying. 

I think I need a different approach to that compile_multiple macro for doing per-function compilation though. IIRC the way I've done it means I can't do the ERL generation for the libraries that use it.

Ziemas:
Idk what to do about `iopreboot`. With cmake we can't build IOP code in the EE cmake project but the EE project depending on IOP doesn't seem right. Maybe we'll have to split it out further into more cmake projects.

uyjulian:
You could have a IOP subproject for `imgdrv` and have EE subproject `iopreboot` depend on that single IOP subproject only.  
Then leave the IOP megaproject for `common` and `iop` subdirs

PR #665, opened 2024-10-04
Errno updates

uyjulian:
Includes descriptions for definitions and also added definitions for 47 and 48


fjtrujy (approved):
Nice improvement!

israpps:
since we are dealing with errno. what about adding PS2 error codes there too?

For example. I see on OPL EE Core
```c
#define SCE_EBINDMISS 0x10000 /** Could not bind with RPC server */
#define SCE_ECALLMISS 0x10001 /** Could not call the RPC function */
#define SCE_ELOADMISS 0x10003 /** ELF/Module load failed */
```
are those actual SCE Errors?

uyjulian:
errno.h would be for standard error codes. The `SCE_` prefixed ones aren't there

israpps:
> errno.h would be for standard error codes. The `SCE_` prefixed ones aren't there

oh. didn't saw they're already in loadfile.h

PR #669, opened 2024-10-19
bin2c: Add O_BINARY for mingw-w64-i686

F0bes:
For whatever reason `open` has completely different behaviour between the i686 and x86_64 MinGW toolchains.
This should fix `bin2c` when it's generated through GitHub actions.


F0bes:
Nevermind, this produces better results but there is still some unknown issue.

PR #673, opened 2024-11-06
fix romimg bundled on docker containers

israpps:
instead of relying on strlen, go straight away with max bufsize. before this change the romimg from docker would segfault while local build works like a charm

New feature, `-c` and `-a` commands will have new variants, `-C` and `-A`, these will manipulate the filename written to ROMFS removing extension and converting to uppercase

So for example
```bash
romimg -C IOPRP.IMG $PS2SDK/iop/irx/secrman.irx
```
Will register secrman.irx as `SECRMAN` on the image, this seems to be needed when the IOPRP is replacing startup IOP modules

### DIFF
--- a/tools/romimg/src/platform.h
+++ b/tools/romimg/src/platform.h
@@ -12,5 +12,8 @@ int GetCurrentWorkingDirectory(char *buffer, unsigned int BufferSize);
 #else
 #define PATHSEP '/'
 #endif
+#ifndef MAX_PATH
+#define MAX_PATH 4096
+#endif
 
 #endif /* __PLATFORM_H__ */
--- a/tools/romimg/src/platform.h
+++ b/tools/romimg/src/platform.h
@@ -13,8 +13,5 @@ void upperbuff(char *temp);
 #else
 #define PATHSEP '/'
 #endif
-#ifndef MAX_PATH
-#define MAX_PATH 4096
-#endif
 
 #endif /* __PLATFORM_H__ */
--- a/tools/romimg/src/platform.h
+++ b/tools/romimg/src/platform.h
@@ -6,6 +6,7 @@ int GetLocalhostName(char *buffer, unsigned int BufferSize);
 unsigned int GetSystemDate(void);
 unsigned int GetFileCreationDate(const char *path);
 int GetCurrentWorkingDirectory(char *buffer, unsigned int BufferSize);
+void upperbuff(char *temp);
 
 #if defined(_WIN32) || defined(WIN32)
 #define PATHSEP '\\'
--- a/tools/romimg/src/romimg.c
+++ b/tools/romimg/src/romimg.c
@@ -6,10 +6,15 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
+#include <limits.h>
 
 #include "platform.h"
 #include "romimg.h"
 #include "SonyRX.h"
+
+#define BUFCHK(X) (X[0] == '\0') ? "" : X
+#define IMAGE_COMMENT_BASESIZE 31
 
 struct ROMImgStat
 {
@@ -139,12 +144,7 @@ static int GetExtInfoStat(const struct ROMImgStat *ImageStat, struct RomDirFileF
 int CreateBlankROMImg(const char *filename, ROMIMG *ROMImg)
 {
 	unsigned int CommentLength;
-	char LocalhostName[32], cwd[128];
-#if defined(_WIN32) || defined(WIN32)
-	char UserName[32] = "";
-#else
-	char* UserName;
-#endif
+	char LocalhostName[32] = "\0", cwd[MAX_PATH] = "\0", UserName[32] = "\0";
 	struct FileEntry *ResetFile;
 	struct ExtInfoFieldEntry *ExtInfoEntry;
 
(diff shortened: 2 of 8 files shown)


tools/romimg/src/romimg.c
```diff
@@ -154,13 +150,15 @@ int CreateBlankROMImg(const char *filename, ROMIMG *ROMImg)
...
 #else
-	UserName = getenv("USER");
+    getlogin_r(UserName, sizeof(UserName));
 #endif
 	GetLocalhostName(LocalhostName, sizeof(LocalhostName));
 	GetCurrentWorkingDirectory(cwd, sizeof(cwd));
 	/* Comment format: YYYYMMDD-XXXYYY,conffile,<filename>,<user>@<localhost>/<image path> */
-	CommentLength = 31 + strlen(filename) + strlen(UserName) + strlen(LocalhostName) + strlen(cwd);
-	ROMImg->comment = (char *)malloc(CommentLength);
+	CommentLength = 31 + strlen(filename) + 32 + 32 + 128;
```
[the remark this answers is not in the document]
fjtrujy: Another alternative is to use `sizeof(LocalhostName)`, this should return 32 as well

tools/romimg/src/romimg.c
```diff
@@ -139,12 +140,7 @@ static int GetExtInfoStat(const struct ROMImgStat *ImageStat, struct RomDirFileF
 int CreateBlankROMImg(const char *filename, ROMIMG *ROMImg)
 {
 	unsigned int CommentLength;
-	char LocalhostName[32], cwd[128];
-#if defined(_WIN32) || defined(WIN32)
-	char UserName[32] = "";
-#else
-	char* UserName;
-#endif
+	char LocalhostName[32] = "\0", cwd[128] = "\0", UserName[32] = "\0";
```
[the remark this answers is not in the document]
israpps: I guess there is no cross platform definition of MAX_PATH. linux seems to use PATH_MAX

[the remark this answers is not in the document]
fjtrujy: Yeah sorry PATH_MAX or whatever is called, I don't remember right now the POSIX solution for that hehe

uyjulian:
Add or remove one from either buffer or length

israpps:
> Add or remove one from either buffer or length

will take a look. buffer is 11 and i'm reading 10 already

uyjulian:
The buffer is being truncated.

That is, the destination buffer is smaller than the source buffer. So increase the destination buffer size or decrease the source buffer size

israpps:
im kinda lost now
```
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:10: error: ‘__builtin_strncpy’ output may be truncated copying 9 bytes from a string of length 9 [-Werror=stringop-truncation]
```

tools/romimg/src/romimg.c
```diff
@@ -6,11 +6,16 @@
...
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
+#include <limits.h>
 
 #include "platform.h"
 #include "romimg.h"
 #include "SonyRX.h"
 
+#define BUFCHK(X) (X[0] == '\0') ? "" : X
```
fjtrujy, line 16: Do you think this is really needed? 
I think it makes no sense at all, the behavior should be the same, no matter if the first character is `\0`

israpps: yeah, it remained from older approach. will deal with it now

israpps:
fixed. seems to work as expected!

fjtrujy:
Finally 😄

tools/romimg/src/romimg.c
```diff
@@ -154,14 +153,15 @@ int CreateBlankROMImg(const char *filename, ROMIMG *ROMImg)
...
-	UserName = getenv("USER");
+    getlogin_r(UserName, sizeof(UserName));
 #endif
 	GetLocalhostName(LocalhostName, sizeof(LocalhostName));
 	GetCurrentWorkingDirectory(cwd, sizeof(cwd));
 	/* Comment format: YYYYMMDD-XXXYYY,conffile,<filename>,<user>@<localhost>/<image path> */
-	CommentLength = 31 + strlen(filename) + strlen(UserName) + strlen(LocalhostName) + strlen(cwd);
-	ROMImg->comment = (char *)malloc(CommentLength);
-	sprintf(ROMImg->comment, "%08x,conffile,%s,%s@%s/%s", ROMImg->date, filename, (UserName[0] =='\0')?"":UserName, LocalhostName, cwd);
+	CommentLength = IMAGE_COMMENT_BASESIZE + strlen(filename) + sizeof(LocalhostName) + sizeof(UserName) + MAX_PATH;
```
fjtrujy, line 161: use `sizeof` at the end

tools/romimg/src/platform.h
```diff
@@ -6,11 +6,15 @@ int GetLocalhostName(char *buffer, unsigned int BufferSize);
...
 int GetCurrentWorkingDirectory(char *buffer, unsigned int BufferSize);
+void upperbuff(char *temp);
 
 #if defined(_WIN32) || defined(WIN32)
 #define PATHSEP '\\'
 #else
 #define PATHSEP '/'
 #endif
+#ifndef MAX_PATH
+#define MAX_PATH 4096
```
fjtrujy, line 17: `PATH_MAX` is available in all the OS, so use it instead and remove this from here

tools/romimg/src/romimg.c
```diff
@@ -139,12 +143,7 @@ static int GetExtInfoStat(const struct ROMImgStat *ImageStat, struct RomDirFileF
 int CreateBlankROMImg(const char *filename, ROMIMG *ROMImg)
 {
 	unsigned int CommentLength;
-	char LocalhostName[32], cwd[128];
-#if defined(_WIN32) || defined(WIN32)
-	char UserName[32] = "";
-#else
-	char* UserName;
-#endif
+	char LocalhostName[32] = "\0", cwd[MAX_PATH] = "\0", UserName[32] = "\0";
```
fjtrujy, line 146: I recommend to initializate them with `= {0}` instead, it is equivalent to put all the fields with `\0` directly

tools/romimg/src/romimg.c
```diff
@@ -452,15 +452,26 @@ static int AddExtInfoStat(struct FileEntry *file, unsigned char type, void *data
...
 	FILE *InputFile;
 	int result;
 	unsigned int FileDateStamp;
 	unsigned short FileVersion;
 	if ((InputFile = fopen(path, "rb")) != NULL) {
 		const char* fname = strrchr(path, PATHSEP);
 		if (fname == NULL) fname = path; else fname++;
+        if (upperconv) {
+            strncpy(tbuf, fname, sizeof(tbuf));
+            tbuf[sizeof(tbuf) - 1] = '\0';
```
fjtrujy, line 467: if you initialize with `= {0}` then this line shouldn't be required either

tools/romimg/src/platform.h
israpps: windows has the name flipped `MAX_PATH` instead of  `PATH_MAX`

fjtrujy: I'm using `MAX_PATH` in my fork and all the CI/CD was passing.
You are probably speaking about compiling natively, not through MinGW, in theory, we shouldn't care about these platforms...

israpps: not sure where it is defined then... everything in google says on limits.h but it's not the case

fjtrujy (approved):

PR #675, opened 2024-11-08
[SECRMAN] rename funcs related to kbit/kc obtainment

israpps:
these functions retrieve the kbit and kc, after they were decrypted and encrypted with session key on the mechacon, after being sent via secrdownloadheader


uyjulian (approved):
Lgtm

PR #687, opened 2024-12-01
iomanX null terminate after strncpy in parsefile

### LINKED ISSUE #683: Issues with BDM devices and fileXio after #671
@fjtrujy @uyjulian 

It looks like #671 introduced a serious issue with mixed usage of `fileXio` and POSIX calls (at least) on BDM devices.
[NHDDL](https://github.com/pcm720/nhddl) needs to get the device driver name and device number for launching Neutrino.
Currently, the only way to get those is to use the `fileXioIoctl2` call.

Since after #671 has been merged, I keep getting strange errors on real PS2 hardware (PS2 SCPH-70000), even when compling using the latest PS2SDK commit.
When multiple BDM modules and devices are loaded, I get errors when using `opendir`/`closedir` calls.
Most of the times `opendir` can't open `mass0:` even though it was successfully opened and closed with `fileXioDopen` previously. Sometimes trying to call `closedir` on an opened directory when traversing directories crashes PS2 because the DIR pointer suddenly becomes NULL.

In addition to this, when only one BDM device is initialized, using `mkdir` call after `fileXioIoctl2` causes `open("rom0:FONTM"...)` to fail. This particular issue can be reliably reproduced with the following code when launched from PS2LINK:
https://github.com/pcm720/ps2sdk-test/blob/main/main.c

Note that the issue occurs only on a real hardware when using `fileXio.irx`.
It never fails with `fileXio_verbose.irx` or when running on PCSX2.

Reverting 2a131ac831c686d7851b3b39a250eb4768bd1211 or calling `fileXioExit` after `fileXioIoctl2` fixes all of those issues.

---

@fjtrujy's example that demonstrates this issue on PCSX2 with USB mass drivers:
https://github.com/fjtrujy/helloWorldPS2/blob/mass_driver_bug/main.c

uyjulian:
Closes #683

Thanks to @pcm720

### DIFF
--- a/iop/system/iomanx/src/iomanX.c
+++ b/iop/system/iomanx/src/iomanX.c
@@ -1052,6 +1052,7 @@ static const char *parsefile(const char *path, iomanX_iop_device_t **p_device, i
 	if ( devname_len > (sizeof(canon) - 1) )
 		return NULL;
 	strncpy(canon, path_trimmed, devname_len);
+	canon[devname_len] = 0;
 	unit = 0;
 	// Search backward for the unit number.
 	while ( isnum(canon[devname_len - 1]) )


iop/system/iomanx/src/iomanX.c
```diff
@@ -1052,6 +1052,7 @@ static const char *parsefile(const char *path, iomanX_iop_device_t **p_device, i
 	if ( devname_len > (sizeof(canon) - 1) )
 		return NULL;
 	strncpy(canon, path_trimmed, devname_len);
+	canon[devname_len] = 0;
```
fjtrujy, line 1055: Is this the same than `{0}`

uyjulian: No. `{0}` is only used for initializer of an array, but we are setting an element of an array (terminating the string after  `strncpy`)

PR #689, opened 2024-12-03
refactor: use extern for forward decls in headers

uyjulian:


fjtrujy:
What's the main difference we get with this change?

uyjulian:
It basically makes it easier to determine (e.g. from regex) what is a function prototype.

fjtrujy:
Please solve conflicts

uyjulian:
Done

PR #691, opened 2024-12-05
Implement more functions in libcglue 20241204

uyjulian:
`sys/termios.h`, `sys/uio.h`, and `sys/utime.h` were adapted from newlib cygwin port  
`readv` and `writev` were implemented  

Document where functions are called from in newlib

This gets libarchive mostly building with minimal  patches (2 lines)

### DIFF
--- a/ee/libcglue/src/glue.c
+++ b/ee/libcglue/src/glue.c
@@ -21,11 +21,15 @@
 
 #include <pwd.h>
 #include <time.h>
+#include <grp.h>
+#include <sys/termios.h>
 #include <sys/time.h>
 #include <sys/timeb.h>
 #include <sys/times.h>
 #include <sys/utime.h>
+#include <sys/uio.h>
 #include <sys/stat.h>
+#include <sys/signal.h>
 #include <sys/types.h>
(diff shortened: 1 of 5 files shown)


ee/libcglue/src/glue.c
```diff
@@ -453,8 +484,18 @@ int chdir(const char *path) {
...
+#ifdef F_fchdir
+int fchdir(int fd)
+{
+	errno = ENOSYS;
+	return -1;
+}
+#endif
+
+#ifdef F__mkdir
+// Called from newlib mkdirr.c
```
fjtrujy, line 496: I think then we must put `-DHAVE_MKDIR` in the options of compiling `newlib`

uyjulian: Yeah that might be better

fjtrujy: How do you want to do this? are you going a create to newlib too? I would like to merge this PR.

uyjulian: I made a newlib https://github.com/ps2dev/newlib/pull/16

PR #693, opened 2024-12-09
Avoid endless loops when RPC binding

israpps:


fjtrujy:
Please let's add some description to the PRs to understand better what's the PR about

israpps:
> Please let's add some description to the PRs to understand better what's the PR about

@fjtrujy done

israpps:
Seems like my branch was based on my master branch. That's why that change I there. It wasn't intentional.

Will fix

israpps:
I see lots of mess with merge upstream 

Could you squash and merge instead of normal merge @fjtrujy ?

PR #695, opened 2024-12-10
Use original error return values for IO stub ops

uyjulian:


Wolf3s:
I will test this.

Wolf3s:
Looks fine in the same state.

fjtrujy:
@uyjulian any reason why we want to do this? any particular error being reported?

uyjulian:
> @uyjulian any reason why we want to do this? any particular error being reported?

imgdrv was broken because it was returning the wrong result for open

PR #698, opened 2024-12-12
[EE RPC]: break all RPC binding after 100 retries

israpps:
applications will no longer hang eternally on a loop.

Applications that bind to RPC should check return values of rpc binding functions and throw errors if necesary to avoid the program to continue execution as if server is ok


rickgaiser:
I saw your comment here
https://github.com/rickgaiser/neutrino/pull/74#issuecomment-2637827748

Question:

If we load an IOP module and it takes for instance 500ms to have the service running, will this then break existing applications? Waiting 1 second before failing seems like an ok time to me. The user will then have to wait for 1s to see the application has failed to start. Instead of it freezing forever.

How much time is 100 x nopdelay ?

israpps:
I haven't tested it, not sure how much time it takes

It's open to changes tho.

But the base idea is IMO very good, don't use endless loops without a failsafe for RPC binding

sp193:
I am personally not so sure. This is a game console and the original idea was likely to never fail because there is nothing else to do. If your RPC is unexpectedly unavailable, what can the user do?

israpps:
> I am personally not so sure. This is a game console and the original idea was likely to never fail because there is nothing else to do. If your RPC is unexpectedly unavailable, what can the user do?

the intention of this is not to throw the problem to the user, it's to reduce the reports of "hangs at black screen".

Instead display an error message, and if the app is in a very early setup stage, maybe just change bg color and call sleepthread.

PR #703, opened 2024-12-20
Re-Add ioman return value of 64 bits.

Wolf3s:


iop/system/ioman/include/ioman.h
```diff
@@ -71,8 +71,10 @@ typedef struct _iop_device {
 } iop_device_t;
 
 #define IOMAN_RETURN_VALUE_IMPL(val) \
-	static inline int my_ioman_retval_##val##_int(void) {return -val;}
+	static inline int my_ioman_retval_##val##_int(void) {return -val;} \
+	static inline signed long long my_ioman_retval_##val##_s64(void) {return -val;}	
 #define IOMAN_RETURN_VALUE(val) ((void*)&my_ioman_retval_##val##_int)
+#define IOMAN_RETURN_VALUE_S64(val) ((void*)&my_ioman_retval_##val##_s64)
```
fjtrujy, line 77: But it is not being used, where do you want to use it?

fjtrujy:
Please creates description for what do you want to merge this PR

Wolf3s:
Open PS2 Loader.

Obter o Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

uyjulian:
Should use `IOMANX_RETURN_VALUE_S64` because the iomanX ops is the only place where the lseek64 is available

fjtrujy:
Please @uyjulian take care of this PR as you are more aware than me about this.

Wolf3s:
> Should use `IOMANX_RETURN_VALUE_S64` because the iomanX ops is the only place where the lseek64 is available

Doesn't work on opl they used to a custom implementation of ioman to do cdvdman stuff, Well i think there is might be a way to remove that custom structure and add ps2sdk standard

Wolf3s:
Thx, But tommorow i will test removing the custom structure from cdvdman if doesn´t work then i use the ioman stuff. Btw this should stay if appear another function which contains 64 bit values. So this come in handy.

PR #704, opened 2024-12-24
Implementation of thmon module

uyjulian:


iop/system/threadman/include/thbase.h
```diff
@@ -168,6 +170,8 @@ extern int GetSystemStatusFlag();
 // clang-format off
 #define thbase_IMPORTS \
 	thbase_IMPORTS_start \
+ \
```
fjtrujy, line 173: Wrong alignement?

fjtrujy (approved):

iop/system/threadman/include/thbase.h
uyjulian: Fixed

fjtrujy (approved):

PR #711, opened 2025-01-01
fix ioctl failing with ENOSYS.

hakanrw:
Currently, ioctl calls are failing, as `__fioOpsInitializeImpl` and `__fileXioOpsInitializeImpl` are checking for the wrong symbol when setting operation handlers.

Offending lines:

https://github.com/ps2dev/ps2sdk/blob/69f54b9f923db871ea1582a62c9db8866a017fe5/ee/libcglue/src/ps2sdkapi.c#L489

https://github.com/ps2dev/ps2sdk/blob/69f54b9f923db871ea1582a62c9db8866a017fe5/ee/rpc/filexio/src/fileXio_ps2sdk.c#L506

`_ioctl` is not defined anywhere, thus `x_ops_file.ioctl` will remain as `NULL`. This in turn will result in an ENOSYS in `_ps2sdk_ioctl`. The correct symbol to check here would be `_ps2sdk_ioctl`. I corrected the error, and defined `_ioctl` in `glue.c` for consistency.

Tested and validated via PCSX2 emulator.

### DIFF
--- a/ee/libcglue/Makefile
+++ b/ee/libcglue/Makefile
@@ -75,6 +75,7 @@ GLUE_OBJS = \
 	lstat.o \
 	_fstat.o \
 	access.o \
+	_ioctl.o \
 	_fcntl.o \
 	getdents.o \
 	_lseek.o \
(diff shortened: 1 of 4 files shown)


uyjulian:
Thanks!

PR #716, opened 2025-01-10
Delete semaphore on E_SIF_PKT_SEND error

fjtrujy:
## Description
This PR solves an issue when `SifSendCmd` fails and the Semaphore wasn't being deleted.
Additionally, I'm now deleting the semaphore in case of success in the `_request_end` callback function, which simplifies the logic a bit.

Cheers

### DIFF
--- a/ee/kernel/src/sifrpc.c
+++ b/ee/kernel/src/sifrpc.c
@@ -148,6 +148,7 @@ int SifBindRpc(SifRpcClientData_t *cd, int sid, int mode)
 
     if (!SifSendCmd(SIF_CMD_RPC_BIND, bind, RPC_PACKET_SIZE, NULL, NULL, 0)) {
         rpc_packet_free(bind);
+        DeleteSema(cd->hdr.sema_id);
         return -E_SIF_PKT_SEND;
     }
 
@@ -214,6 +215,7 @@ int SifCallRpc(SifRpcClientData_t *cd, int rpc_number, int mode,
 
     if (!SifSendCmd(SIF_CMD_RPC_CALL, call, RPC_PACKET_SIZE, sendbuf, cd->buff, ssize)) {
         rpc_packet_free(call);
+        DeleteSema(cd->hdr.sema_id);
         return -E_SIF_PKT_SEND;
     }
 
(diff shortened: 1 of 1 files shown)


fjtrujy:
Hi @F0bes  take a look if you have some time!

ee/kernel/src/sifrpc.c
```diff
@@ -148,11 +148,11 @@ int SifBindRpc(SifRpcClientData_t *cd, int sid, int mode)
 
     if (!SifSendCmd(SIF_CMD_RPC_BIND, bind, RPC_PACKET_SIZE, NULL, NULL, 0)) {
         rpc_packet_free(bind);
+        DeleteSema(cd->hdr.sema_id);
         return -E_SIF_PKT_SEND;
     }
 
     WaitSema(cd->hdr.sema_id);
-    DeleteSema(cd->hdr.sema_id);
```
uyjulian, line 155: SCE 3.0.0 keeps this instead of moving `iDeleteSema` inside intr handler.

fjtrujy: Isn’t the same? Actually it would be better implementation mine, as it will generate smaller binary

uyjulian: Actually this looks fine. Since `iSignalSema` puts the waiting thread in the ready queue the reference to the sema isn't needed anymore. Also it appears when `iDeleteSema` is called waiting on threads are automatically put in the ready queue.

fjtrujy: Nice!

F0bes: As long as the threads are properly queued when you delete the sema this looks fine to me.

uyjulian: The only potential problem I see is that this creates a race condition on `WaitSema`. For debug EE kernel when debugging functions are not disabled, if the semaphore becomes invalid before `WaitSema` is called, it will trap.

fjtrujy: How can be this possible? I can put it back if you think is better.

fjtrujy: Updated!

uyjulian:
Looks good

PR #717, opened 2025-01-11
cleanup: remove unneeded memset in SifExitIopHeap

### LINKED ISSUE #434: kr_ps2: possible typo
https://github.com/ps2dev/ps2sdk/blob/c45ad1d9421ec4d75ab80dc283b9165487374713/ee/kernel/src/iopheap.c#L65

kr_ps2:
Looks like this is a typo, there should be:
`memset(&_ih_cd, 0, sizeof _ih_cd);`

uyjulian:
It is redundant. It may be a typo on the version variable, but we don't need to check version in that case

Closes #434

### DIFF
--- a/ee/kernel/src/iopheap.c
+++ b/ee/kernel/src/iopheap.c
@@ -62,7 +62,6 @@ int SifInitIopHeap()
 void SifExitIopHeap()
 {
     _ih_caps = 0;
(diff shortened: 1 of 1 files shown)


sp193:
That's strange. Why do we need this? It's actually also in SifInitIopheap().

PR #723, opened 2025-01-23
Link timer, timezone, and RTC functions only if they are used

uyjulian:
Allows smaller executables when not using those functions.

Nanosleep and timezone samples continue to function.

### DIFF
--- a/ee/kernel/Makefile
+++ b/ee/kernel/Makefile
@@ -125,6 +125,8 @@ EXIT_OBJS += _exit_internals.o SetArg.o Exit.o ExecPS2.o LoadExecPS2.o ExecOSD.o
 
 TIMER_OBJS = \
 	timer_data.o \
+	_ps2sdk_init_timer.o \
+	_ps2sdk_deinit_timer.o \
 	SetT2.o \
 	SetT2_COUNT.o \
 	SetT2_MODE.o \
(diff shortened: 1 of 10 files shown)


fjtrujy:
I have downloaded artefacts generated here, I have run `mips64r5900el-ps2-elf-strip` on the elfs and binaries are bigger:
Yours
```
-rwxr-xr-x@ 1 fjtrujy  staff  174484 24 ene 15:12 nanosleep_test.elf
-rwxr-xr-x@ 1 fjtrujy  staff  72308 24 ene 15:13 nanoHelloWorld.elf
```

Current master
```
-rwxr-xr-x@ 1 fjtrujy  staff  174036 24 ene 15:12 nanosleep_test.elf
-rwxr-xr-x@ 1 fjtrujy  staff  72180 24 ene 15:13 nanoHelloWorld.elf
```

Here we have 2 basic examples.
1. `nanosleep` example is using timer, so I was expecting somehow to increase sizes (about the new symbols defined)
2. `nanoHelloWorld` however, is not using anything about timer, is true is disabled too, but it is still bigger

Not sure if some other are reducing space...

uyjulian:
Fixed; it is now smaller in size:

```
1333276 old/nanosleep_test.elf
1314488 new/nanosleep_test.elf
```

fjtrujy (approved):
Well done!

PR #724, opened 2025-01-24
SetMemoryMode syscall related changes

uyjulian:
This mainly avoids a fileio call at init

### DIFF
--- a/ee/libcglue/samples/mem_test/main.c
+++ b/ee/libcglue/samples/mem_test/main.c
@@ -16,6 +16,7 @@
 #include <assert.h>
 
 #include <kernel.h>
+#include <syscallnr.h>
 #include <rom0_info.h>
 
 #if defined(SCREEN_DEBUG)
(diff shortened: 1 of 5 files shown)


israpps:
Just curiosity 

Is this really PSX only? Isn't it applicable to namco system 256 since it also recycles the PSX EE+GS chip revision

ee/libcglue/samples/mem_test/main.c
```diff
@@ -52,7 +53,7 @@ void _libcglue_rtc_update() {}
 // "weak" function called by crt0.o
 void _ps2sdk_memory_init()
 {
-    if (GetMemorySize() == (32 * 1024 * 1024) && IsDESRMachine() == 1) {
+    if (GetMemorySize() == (32 * 1024 * 1024) && GetSyscallHandler(__NR_GetMemoryMode) && GetMemoryMode() == 1) {
```
fjtrujy, line 56: should we create a macro to make easier the usage of this `GetSyscallHandler(__NR_GetMemoryMode) && GetMemoryMode() == 1`? or even the whole `if` condition

uyjulian:
> Just curiosity
> 
> Is this really PSX only? Isn't it applicable to namco system 256 since it also recycles the PSX EE+GS chip revision

Yes, it is not applicable to System 256/148 or DTL-T due to missing relevant system calls

ee/libcglue/samples/mem_test/main.c
fjtrujy: 🆗

fjtrujy:
@rickgaiser any concern?

PR #725, opened 2025-01-25
SIF API definition merge

uyjulian:
This merges the exposed API definitions of SIF RPC and SIF CMD.  
It should also make it easier to see usages of those functions.  

wLaunchELF can load itself.


ee/kernel/src/sifcmd.c
```diff
@@ -155,19 +158,14 @@ int _SifCmdIntHandler(int channel)
 
     if (header->cid & SIF_CMD_ID_SYSTEM) {
         if ((u32)id < cmd_data->nr_sys_handlers)
-            cmd_handlers = cmd_data->sys_cmd_handlers;
-        else
-            goto out;
+            if (cmd_data->sys_cmd_handlers[id].handler)
```
fjtrujy, line 161: this function changed its logic, is that right?

fjtrujy:
I see plenty of `__attribute__((aligned(XX)))` are they needed? why weren't they set before? Was it a bug?

ee/kernel/src/sifcmd.c
uyjulian: The function behavior is the same. However, changes were needed due to the fact that the structure types are now different (due to the added member)

uyjulian:
The alignment stuff was synchronized across the implementations. It was a bug

sp193:
I recall that address alignment for DMA only needs to be 4 bytes at IOP, at 16 at EE. 64 is only required at the EE, if the D-cache line(s) over the region are to be written back.

The lack of alignment attributes, meant that the compiler default would be used.

I suppose that it is better to clearly specify the alignment.

PR #727, opened 2025-01-26
[ps2hdd] Do not restore journal on unformatted HDDs to avoid corrupting non-APA HDDs

### LINKED ISSUE #726: Loading ps2hdd.irx corrupts non-APA HDDs
Simply loading ps2hdd.irx corrupts primary GPT on non-APA HDD.

The culprit seems to be the use of `apaJournalRestore` function during module initalization:
https://github.com/ps2dev/ps2sdk/blob/748cd5c69129783be293d89f656695e21a3e28a6/iop/hdd/apa/src/hdd.c#L342-L346

This function apparently always calls `apaJournalReset` that flushes on return without checking if the drive is actually APA-formatted.
https://github.com/ps2dev/ps2sdk/blob/748cd5c69129783be293d89f656695e21a3e28a6/iop/hdd/libapa/src/journal.c#L29-L45

Commenting out `apaJournalRestore` in `hdd.c` prevents the corruption and keeps primary GPT intact.
Either journal functions in `libapa` or entrypoint function in `hdd.c` must make sure that drive is APA-formatted before trying to restore the journal.

I don't know enough about APA, but I think checking HDD status and skipping journal restore if HDD is unformatted (`1`) is enough to prevent this:
 https://github.com/ps2dev/ps2sdk/blob/748cd5c69129783be293d89f656695e21a3e28a6/iop/hdd/apa/src/hdd.c#L342
```c
if((hddDevices[i].status != 1) && (apaJournalRestore(i) != 0))
```
This fix was tested in PCSX2 and seems to work properly.

pcm720:
This PR adds a check before `apaJournalRestore` to avoid corrupting non-APA HDDs during ps2hdd initialization.
Fixes #726.

### DIFF
--- a/iop/hdd/apa/src/hdd.c
+++ b/iop/hdd/apa/src/hdd.c
@@ -339,7 +339,7 @@ int APA_ENTRYPOINT(int argc, char *argv[])
 	{
 		if(hddDevices[i].status<2)
 		{
-			if(apaJournalRestore(i) != 0)
+			if((hddDevices[i].status != 1) && apaJournalRestore(i) != 0)
 			{
 				APA_PRINTF(APA_DRV_NAME": error: log check failed.\n");
 				return hddInitError();


uyjulian (approved):
lgtm

PR #733, opened 2025-01-27
Cleanup s assembly files

uyjulian:
There were some unused s files, so those have been removed.  
There were some s files that have intermixed C preprocessor statements and assembly, so those have been cleaned up and made such that the C preprocessor statements have been more isolated from the assembly code.  

Tested wLE can launch itself  
Tested OPL can run games from USB


iop/system/udnl/src/udnl.c
```diff
@@ -367,6 +383,36 @@ static volatile unsigned int *func_00000f80(volatile unsigned int *address)
 
 // 0x00001b38
 void func_00001b38(unsigned int arg1);
+__asm__
```
fjtrujy, line 386: Shouldn't be better if we use the `asm` code directly inside of the function? instead of defining the header and then implementing the "whole" function using ASM?

uyjulian: That would be better, but I plan to do that in another pass (not in this PR) because it requires figuring out what registers are read/written/clobbered and that has a higher change of breaking.

fjtrujy:
Anyway, I would like to split this PR into smaller PRs, to check that the implementation is exactly the same.
Sorry for that, but could you create a PR per ee/iop library involved?

uyjulian:
Yeah, I can do that

PR #735, opened 2025-02-11
Fix SPU2 DMA address calculation

jrdnrs:
I was seeing some strange behaviour when ADPCM voice samples were being uploaded past a certain address.

It seems that the Transfer Start Address was being truncated to 16 bits when it should be 20. This adds the extra 4 bits to the TSAH register, meaning that the full 2MB is now addressable (in 16 bit units). As far as I can tell, this fixes the problem.


uyjulian:
Whoops, that was an oversight on my part. Thanks!

PR #737, opened 2025-02-28
[IOP] warn of IOP_BIN without .irx extension breaks makefile

israpps:
new toolchain does extension replacement. this will warn the programmer of such issue.

If you have a better approach let me know


fjtrujy:
@uyjulian @rickgaiser any alternative here?

PR #739, opened 2025-03-19
fix: recalculate addresses when lower bits are greater than 0x8000

DanielSant0s:
On assembly level, the PS2 need to allocate addresses divided in two 16bit immediate loads into upper and lower 32bits of a register in order to get a pointer to the memory address. The issue happens when you are combining addresses which lower bits are greater than 0x8000, ERL library linker is supposed to add 1 to the upper 16 bit of the address. 
(E.g.: 0x20, 0x8001 -> 0x218001) Well, of course it doesn't do it, which leads to a lot of invalid pointers in bigger ERL binaries, This PR fixes it (in a dirty way), predicting the lower address and then adding a bit significance compensation, in order to relocate to the right address. That fix was widely tested on my plugin system, which injects compiled C code in GTA San Andreas.

### DIFF
--- a/ee/erl/src/erl.c
+++ b/ee/erl/src/erl.c
@@ -583,7 +583,7 @@ static int read_erl(int elf_handle, u8 * elf_mem, u32 addr, struct erl_record_t 
     struct elf_header_t head;
     struct elf_section_t * sec = 0;
     struct elf_symbol_t * sym = 0;
-    struct elf_reloc_t reloc;
+    struct elf_reloc_t reloc, next_reloc;
     int i, j;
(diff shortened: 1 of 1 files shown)


fjtrujy:
@uyjulian @rickgaiser could you take a look at this PR?

rickgaiser:
> @uyjulian @rickgaiser could you take a look at this PR?

I have no objections. @DanielSant0s is the expert and one of the few (if not the only) user of ERL. If it works for him then it's good to merge.

uyjulian (approved):
Lgtm

PR #741, opened 2025-03-24
feat: add addend to target relocation address

DanielSant0s:
This PR adds some level of support for addend on relocations. Since RELA type was added recently, I didn't implement any support for it. It should cover most of the cases, tested while developing AthenaEnv ERL support.


uyjulian:
lgtm

PR #746, opened 2025-03-29
sio2man changes 28032025

uyjulian:


fjtrujy:
What's the purpose of all these callbacks that are always called?

uyjulian:
It will basically avoid a branch for checking if the callback pointer is NULL, since it is always set to something

PR #747, opened 2025-03-30
graph: fix callback function ptr

Ziemas:
Specify argument type of function ptr, empty args will be equavalent to `(void)` in future C standards. Fixes a GCC15 error.


uyjulian (approved):
lgtm

PR #748, opened 2025-03-30
Implement DECI2 TTY functions

Ziemas:
These are handy for people using RDB/dsedb since they put messages on the expected DECI2 protocol, they also lets you print from interrupt handlers.


uyjulian (approved):
lgtm

PR #752, opened 2025-04-22
usbd doGetDeviceLocation index fix

DKWDRV:
Makes it like the original.

if cpCount and count are both 0 (only root hub device registered) it would access index -1.

resolves DKWDRV/DKWDRV#82


rickgaiser:
Thanks! Merging.

PR #759, opened 2025-05-05
fix: restore libmpeg

DanielSant0s:
This PR fixes libmpeg and it's sample code.

Change registers to N32 ABI standards.
Fix $at usage, changing it for $t8 and/or using "use no/at".
Fix sample code primitive offset values.
Restore Assembly source compilation, since the C source doesn't work yet.

### DIFF
--- a/ee/mpeg/Makefile
+++ b/ee/mpeg/Makefile
@@ -6,7 +6,7 @@
 # Licenced under Academic Free License version 2.0
 # Review ps2sdk README & LICENSE files for further details.
 
-EE_OBJS = libmpeg.o libmpeg_core_c.o erl-support.o
+EE_OBJS = libmpeg.o libmpeg_core.o erl-support.o
 
 include $(PS2SDKSRC)/Defs.make
 include $(PS2SDKSRC)/ee/Rules.lib.make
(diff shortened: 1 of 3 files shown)


fjtrujy:
Please, @uyjulian, take a look. I remember you were trying to make it work.

Why were you trying to make the C version? 
Usually, I prefer to have the C version as it is easy to understand, but watching the ASM code, I see how it is using plenty of ASM instructions that our current GCC version is not emitting isn't?

DanielSant0s:
> Why were you trying to make the C version? Usually, I prefer to have the C version as it is easy to understand, but watching the ASM code, I see how it is using plenty of ASM instructions that our current GCC version is not emitting isn't?

We could keep the ASM code until C version get fixed

uyjulian:
The main issue I was having is the site where the sample files were located is no longer up

DanielSant0s:
> The main issue I was having is the site where the sample files were located is no longer up

I did some sample to test it using TMPGEnc, you can get by demuxing a PSS too. So it's actually tested.

DanielSant0s:
Here's the video file I used to test: [Link](https://www.mediafire.com/file/edxvycyr2yikypf/VLOAD.m2v/file)

Ziemas:
Do we know if this code was originally decompiled from some known library?

uyjulian:
SMS or libmpeg, probably

fjtrujy:
Let's merge it as it is not hurting anyone.
@uyjulian, @DanielSant0s , @Ziemas feel free to improve the existing tests attaching new files if needed, in this way we can keep the feature working in a easier way

Issue #753, opened 2025-04-22
Provide compiled latest IRX archive with workflows


DKWDRV:
IRX archive for latest SDK would be very useful as download. IRX barely take any space. 
Maybe someone can add it in workflows?

israpps:
I implemented such thing for my personal fork of PS2SDK. Super useful to quickly get debug IRX. I could make a PR.

DKWDRV:
@israpps very impressive! Why not PR into main branch?
One single archive download will be fine. release/debug folder inside for the two types.

DKWDRV:
Bump @fjtrujy  @uyjulian @rickgaiser @israpps

fjtrujy:
You need to be more concrete about what specifically you want and where (and the use case too)
To upload to the artifacts of the workflows is quite easy

israpps:
He wants the generated IRX modules both normal and debug uploaded along with the samples

fjtrujy:
Done in #764

PR #760, opened 2025-05-05
Use BDM bd.devNr for port numbers

rickgaiser:
This PR fixes the use of the block_device `devNr` field. Whenever possible the `devNr` should hold a value that is persistent across reboots.

- **ATA**: master=0, slave=1 (was already implemented)
- **USB**: 0=first port, 1=second port, 2=hub, fixed in this PR
- **MX4SIO**: 1=second port, fixed in this PR
- **IEEE1394**: not implemented (incrementing number)
- **UDPBD**: not implemented (incrementing number)

The already existing ioctl can be used to query this device number:
https://github.com/ps2dev/ps2sdk/blob/88fdeb04289a0bf5c3dbf1e9ebfcd825cd82105f/common/include/usbhdfsd-common.h#L38-L39

@DKWDRV I have put this code together for you, to end the dicussion in https://github.com/ps2dev/ps2sdk/pull/757. I only compile tested it. Please check [this commit](https://github.com/ps2dev/ps2sdk/commit/6a28311284bad6bf49ea01e91c24af76d06dbb7f) and test if this works for you.

### DIFF
--- a/iop/iLink/IEEE1394_bd/src/include/scsi.h
+++ b/iop/iLink/IEEE1394_bd/src/include/scsi.h
@@ -5,6 +5,7 @@ struct scsi_interface
 {
     void *priv;
     char *name;
+    unsigned int devNr;
     unsigned int max_sectors;
 
     int (*get_max_lun)(struct scsi_interface *scsi);
--- a/iop/iLink/IEEE1394_bd/src/include/scsi.h
+++ b/iop/iLink/IEEE1394_bd/src/include/scsi.h
@@ -5,6 +5,7 @@ struct scsi_interface
 {
     void *priv;
     char *name;
+    // cppcheck-suppress unusedStructMember
     unsigned int devNr;
     unsigned int max_sectors;
 
--- a/iop/iLink/IEEE1394_bd/src/sbp2_driver.c
+++ b/iop/iLink/IEEE1394_bd/src/sbp2_driver.c
@@ -97,6 +97,7 @@ void init_ieee1394DiskDriver(void)
 
         SBP2Devices[i].scsi.priv        = &SBP2Devices[i];
         SBP2Devices[i].scsi.name        = "sd";
+        SBP2Devices[i].scsi.devNr       = i;
         SBP2Devices[i].scsi.max_sectors = XFER_BLOCK_SIZE / 512;
         SBP2Devices[i].scsi.get_max_lun = sbp2_get_max_lun;
         SBP2Devices[i].scsi.queue_cmd   = sbp2_queue_cmd;
--- a/iop/iLink/IEEE1394_bd/src/scsi.c
+++ b/iop/iLink/IEEE1394_bd/src/scsi.c
@@ -275,6 +275,7 @@ void scsi_connect(struct scsi_interface *scsi)
 
             bd->priv = scsi;
             bd->name = scsi->name;
+            bd->devNr = scsi->devNr;
             scsi_warmup(bd);
             bdm_connect_bd(bd);
             break;
@@ -305,7 +306,6 @@ int scsi_init(void)
     M_DEBUG("%s\n", __func__);
 
     for (i = 0; i < NUM_DEVICES; ++i) {
-        g_scsi_bd[i].devNr = i;
         g_scsi_bd[i].parNr = 0;
         g_scsi_bd[i].parId = 0x00;
 
--- a/iop/sio/mx4sio_bd/src/spi_sdcard_driver.c
+++ b/iop/sio/mx4sio_bd/src/spi_sdcard_driver.c
@@ -52,7 +52,7 @@ spisd_t sdcard;
 struct block_device bd = {
     NULL,        /* priv */
     "sdc",       /* name */
-    0,           /* devNr */
+    1,           /* devNr, mx4sio is always in the second port */
     0,           /* parNr */
     0x00,        /* parId */
     SECTOR_SIZE, /* sectorSize */
@@ -64,7 +64,7 @@ struct block_device bd = {
     spisd_stop };
 
 /* NOTE: SIO2 does *NOT* allow for direct control of /CS line.
- * It's controlled by the SIO2 hardware and automatically asserted at the start 
+ * It's controlled by the SIO2 hardware and automatically asserted at the start
  * of a transfer and deasserted at the end. This has lead to the need for some
  * conditions surrounding dummy writes to avoid timing disruption. */
 static uint8_t spisd_send_cmd(uint8_t cmd, uint32_t arg)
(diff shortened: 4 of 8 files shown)


DKWDRV:
@rickgaiser  I hope #753 gets more attention too! Instead of just samples release having IRX package.

rickgaiser:
> @rickgaiser [ps2homebrew/Open-PS2-Loader#1316](https://github.com/ps2homebrew/Open-PS2-Loader/issues/1316) should show improvements with this (after reboot part)?

I know it will work for neutrino now but for OPL I don't know. That needs to be tested, but I see OPL using the `USBMASS_IOCTL_GET_DEVICE_NUMBER` ioctl, and that should return a reboot persistent number after this PR.

> Where can I get compiled irx?

We are developers not users: git pull the source code to a separate local branch and compile.

> I think the code is mostly fine. sceUsbdGetDeviceLocation should return 0 on success.
> 
> I don't think it's a good idea to default to devNr = 2 for other cases of path[0]. In the future if many devices will be registered it would be default to it. The two usb port are considered as "root usb hub" with port nr = 0. The ports are 01 and 02. If a hub is connected to any of the ports it will be the actual hub itself that takes 01 or 02 as device and any other found USB device will be 02 in this case. I think it's better check for sceUsbdGetDeviceLocation return value and if that is valid either leave path[0] as it is 01 02 or if that causes problems then in the else clause check if the number is higher than 2 and just do number -1?

I was thinking about that also. But if the function returns an error, then path should be unchanged and still contain the initialized value `path[0] = 0;`. Resulting in `devNr = 2;`, so I removed the return value check but I agree it's nicer to explicitly check the return value and act on it, preventing confusion.

When the usb stick cannot be identified as being in the first or the second port, then using `devNr = 2;` is a bit oversimplified, I admit. But becouse the usb driver only supports at maximum 2 usb devices it should be ok, as long as not both devices are in a hub. In that case they will both become "usb2" for BDM. But it should still work and they will also become "mass0" and "mass1". The only disadvantage is that they can not be uniquely identified.
A better solution would be to use an incrementing number for all usb drives that are not directly in one of the two ports. So they become "usb2" and "usb3".

If you can try the current code that would be nice. I'll try to make the above changes in the next few days but I don't have much time.

DKWDRV:
> > @rickgaiser [ps2homebrew/Open-PS2-Loader#1316](https://github.com/ps2homebrew/Open-PS2-Loader/issues/1316) should show improvements with this (after reboot part)?
> 
> I know it will work for neutrino now but for OPL I don't know. That needs to be tested, but I see OPL using the `USBMASS_IOCTL_GET_DEVICE_NUMBER` ioctl, and that should return a reboot persistent number after this PR.

Yes this PR should fix that issue. That PR has two problems. Second one was random crashes after IOP reboot related to the device number.
 
> > Where can I get compiled irx?
> 
> We are developers not users: git pull the source code to a separate local branch and compile.

That seems so much for such a small irx. 
> > I think the code is mostly fine. sceUsbdGetDeviceLocation should return 0 on success.
> > I don't think it's a good idea to default to devNr = 2 for other cases of path[0]. In the future if many devices will be registered it would be default to it. The two usb port are considered as "root usb hub" with port nr = 0. The ports are 01 and 02. If a hub is connected to any of the ports it will be the actual hub itself that takes 01 or 02 as device and any other found USB device will be 02 in this case. I think it's better check for sceUsbdGetDeviceLocation return value and if that is valid either leave path[0] as it is 01 02 or if that causes problems then in the else clause check if the number is higher than 2 and just do number -1?
> 
> I was thinking about that also. But if the function returns an error, then path should be unchanged and still contain the initialized value `path[0] = 0;`. Resulting in `devNr = 2;`, so I removed the return value check but I agree it's nicer to explicitly check the return value and act on it, preventing confusion.

Well there is nothing using or supporting hubs out there. Neither opl or DKWDRV support hubs so for now this will do for at least 2 USB devices support.

> When the usb stick cannot be identified as being in the first or the second port, then using `devNr = 2;` is a bit oversimplified, I admit. But becouse the usb driver only supports at maximum 2 usb devices it should be ok, as long as not both devices are in a hub. In that case they will both become "usb2" for BDM. But it should still work and they will also become "mass0" and "mass1". The only disadvantage is that they can not be uniquely identified. A better solution would be to use an incrementing number for all usb drives that are not directly in one of the two ports. So they become "usb2" and "usb3".

I had given deep thoughts to this. My original PR would solve this perfectly. You could keep everything as is, and check the port number for everything without messing with a million device ids. Even with a hub that would still be the same unique port number since the port index number. 

> If you can try the current code that would be nice. I'll try to make the above changes in the next few days but I don't have much time.

It's ok, just merge it and close #757

uyjulian:
How about, instead of `massX:` or trying to make new information about how to pass devNr to next program, what about disk ID/label

e.g. `massbyuuid0:/0123-ABCD/yourfile.elf` or `massbyuuid0:/01234567-89AB-CDEF-0123-456789ABCDEF/yourfile.elf`

The same thing linux does with `/dev/disk/by-uuid/`

DKWDRV:
In case of further errors I will open P

> How about, instead of `massX:` or trying to make new information about how to pass devNr to next program, what about disk ID/label
> 
> e.g. `massbyuuid0:/0123-ABCD/yourfile.elf` or `massbyuuid0:/01234567-89AB-CDEF-0123-456789ABCDEF/yourfile.elf`
> 
> The same thing linux does with `/dev/disk/by-uuid/`

There are many solutions to get multiple support but almost all will break most homebrew relying on "mass0" specifically.

For now this is a good compromise for both OPL and DKWDRV. So merge?

In the future  perhaps more can be done.

rickgaiser:
I tested it and it does not work as expected.

`sceUsbdGetDeviceLocation` returns:
1:0:0:0... for the right port, I thought that was the second one
2:0:0:0... for the left port, I thought that was the first one

So... it's reversed? How can this be? Is this consistent across all ps2 models?

MC0 is left, MC1 is right. So I expect things to be counted left-to-right.

DKWDRV:
> I tested it and it does not work as expected.
> 
> `sceUsbdGetDeviceLocation` returns: 1:0:0:0... for the right port, I thought that was the second one 2:0:0:0... for the left port, I thought that was the first one
> 
> So... it's reversed? How can this be? Is this consistent across all ps2 models?
> 
> MC0 is left, MC1 is right. So I expect things to be counted left-to-right.

Ughhhhhhhhhh this will never end. Yes it does so and so it should be. See [here](https://github.com/DKWDRV/DKWDRV/issues/81#issuecomment-2816737819)
If your trying to make things fit the universal goodlines you will be dissapointed. It's the PS2 :)

Really my PR took care of everything and left everything as is, only the developer had to do extra (but safe check).

DKWDRV:
Bump @rickgaiser @uyjulian 
Can we merge this?

uyjulian:
Suppress cppcheck unused struct member with `// cppcheck-suppress unusedStructMember` before the offending lines

rickgaiser:
I was still testing with USB port numbers on FAT model, as they don't have left/right, but top/bottom.

In BDM, the USB numbers are:
- top/left = the first port = "usb0"
- bottom/right = the second port = "usb1"

`USBMASS_IOCTL_GET_DEVICE_NUMBER` will report these numbers as `0` and `1`

This is how it should be, so ready to merge (but it's already merged I see).

PR #766, opened 2025-05-14
Apply needed fixes for GCC 15.1

fjtrujy:
The majority of the changes were just wrong pointer types.
Is good GCC is forcing us to be more strict with this kind of thing.

Cheers


uyjulian (approved):
lgtm

PR #768, opened 2025-05-20
add DONGLEMAN variant of MCMAN

israpps:
This variation of MCMAN offers the same funcionality that game bundled DONGLEMAN drivers.

It can freely access both `mc0:/` and `mc1:/` on arcade PS2s.

Also, the driver seems to work well with DAEMON.IRX based on some quick tests


israpps:
Not sure why it now says `int SecrAuthDongle(int, int, int);` is implicitly declared... but I've locally compiled and tested this IRX and works as expected. It can be used to access mc0 and mc1 freely on arcade PS2s....

However, it's highly recommended to still avoid accessing SoulCalibur2 Conquest card with anything that is not the DONGLEMAN module from SC2

israpps:
@uyjulian @rickgaiser 

Do you guys have any idea why it complains of implicit decl of SecrAuthDongle but locally builds perfectly?

iop/memorycard/mcman/src/mcsio2.c
```diff
@@ -519,14 +530,19 @@ int mcman_eraseblock(int port, int slot, int block, void **pagebuf, void *eccbuf
...
 
-		if (p[3] == 0x5a)
+		if (p[3] == 0x5a) {
+            HAKAMA_SIGNALSEMA();
 			return sceMcResSucceed;
+        }
 	} while (++retries < 100);
 
-	if (p[3] == 0x66)
+    HAKAMA_SIGNALSEMA(); //El_isra: Original impl has two semas (still commented there) replaced with one for simplicity
```
fjtrujy, line 539: I think you can't signal before as it could potentially change the value of `p` before you read in the following "ifs"

PR #769, opened 2025-05-30
CDVD module updates 20250530

uyjulian:
Add code for OSD and XOSD variants
Add enums in more places


uyjulian:
`XOSD` variant is used in XOSD used in DESR systems

PR #771, opened 2025-06-14
Set sio2 module default variant to newest revision by default

uyjulian:
This PR:

Moves modules out of `system` folder to `sio` and `input` folders
Renames the `-old` variants to `-1300`  
Renames the `r` variants to `-2000`  
Adds a `-2000` variant to mcman  
Makes the `-2000` variant default


rickgaiser:
PR https://github.com/ps2dev/ps2sdk/pull/746 / commit https://github.com/ps2dev/ps2sdk/commit/fe8fcb1d48a30da99f8f4603d1bff175cc8b3c81 causes black screen on OPL.

If you ask me, this line is wrong:
https://github.com/ps2dev/ps2sdk/blob/ef3e895ff245f433d5837c1739bf62ed5cfc24f8/iop/system/sio2man/src/sio2man.c#L502

But I don't see any of the lines of code from that commit changed in this PR.
How does this PR fix OPL? Does OPL now use a different module variant?

fjtrujy:
> OPL uses a different module variant. The main issue I saw was there was no 2000 API module variant of mcman so I added that and made it the default to avoid mixing up API versions.

I'm not against merging this PR, but you just made it OPL to work by creating this 2000 API variant, and the 1300 variant still looks wrong.
So, could you first try to take a look at the @rickgaiser comment, as I think we are doing something wrong with the `m_sdk13x_totalflag` and `m_sdk13x_curflag` variables?
It should be quite fast for you, as they are used just in a couple of places. So once found, you can create a tiny PR for fixing OPL to boot.

Then once done with this, please could you attach some description to this PR, why it is being created, needs, and/or improvements? I'm already quite lost with too many IOP libraries and versions...

Thanks

uyjulian:
The description has been updated.

The issue is that 1300 and 1400 versions of sio2man have the same API major version despite changes in locking behavior to handle multitap support. This causes issues (e.g. hangs when waiting on locks) when mixing the module versions between these versions without a clear reason as to why it occurs.

The 2000 version has a higher major version, which always has a separate lock reset function, so this avoids the conflicting API as modules targeting older sio2man will now fail with a version mismatch error.

fjtrujy:
@uyjulian not sure if you still want to merge this PR, but it has some conflicts around

uyjulian:
Merge conflicts resolved

fjtrujy:
LGTM!

10522:
@uyjulian @fjtrujy @rickgaiser it looks like this PR broke the controllers, MX4SIO and MMCE in OPL. 

More details here: [1](https://github.com/ps2homebrew/wOPL/issues/232#issuecomment-4581976143), [2](https://github.com/ps2homebrew/Open-PS2-Loader/issues/1727#issuecomment-4623855448), [3](https://github.com/ps2homebrew/Open-PS2-Loader/issues/1726#issuecomment-4623736761), [4](https://github.com/ps2homebrew/Open-PS2-Loader/issues/1725#issue-4541197172), [5](https://github.com/ps2homebrew/Open-PS2-Loader/issues/1724#issue-4490469871), [6](https://github.com/ps2homebrew/Open-PS2-Loader/issues/1730#issue-4657198237).

uyjulian:
The following PR will fix the remaining issue https://github.com/ps2dev/ps2sdk/pull/862

PR #776, opened 2025-07-09
Support for 8MB APA Partitions, getStat Enhancements, apaRemove per APA specs

AKuHAK:
This PR introduces the following changes:
- Reduces the minimum APA partition size from 128 MB to 8 MB.
- Fixes an issue where `__empty` partitions could exceed APA spec limits — they are now restricted to a maximum of 1 GB.
- Enhances `getStat()` to report the total partition size (main + all sub-partitions) using the `private_1` and `private_2` fields.

### DIFF
--- a/iop/hdd/apa/src/apa-opt.h
+++ b/iop/hdd/apa/src/apa-opt.h
@@ -6,13 +6,14 @@
 
 #ifdef APA_POSIX_VER
 #define APA_ALLOW_REMOVE_PARTITION_WITH_LEADING_UNDERSCORE 1
-#define APA_FORMAT_LOCK_MBR 1
-#define APA_FORMAT_MAKE_PARTITIONS 1
-#define APA_STAT_RETURN_PART_LBA 1
-#define APA_SUPPORT_HDL 1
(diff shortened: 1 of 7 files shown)


AKuHAK:
I enabled in POSIX flag 8MB support as I didn't found any problems with it after massive testing. All this changes necessary for future pfs-shell improvements.
getStat change will also be useful for Athena project [OSD-XMB](https://github.com/HiroTex/OSD-XMB)

uyjulian (approved):
lgtm

AKuHAK:
Doh, @uyjulian thanks for fast merge, but i kept one workaround there :D 
https://github.com/ps2dev/ps2sdk/blob/9c436971eea642197a1648dc48298a1cf5514976/iop/hdd/libapa/src/apa.c#L260-L261
here wanna ask you if you know how to safely access HDDDevice.partitionMaxSize . Currently I limited empty partitions to be max 1Gb, but for disks larger than 80Gb APA specs allows larger single partitions (2Gb+), but current code will still limit empty to one Gb. Its definetly better than it was before, cause smaller partitions does not break APA specs, but still enforces fragmentation.

PR #822, opened 2026-03-21
Implement netdev API

uyjulian:
For future implementation of USB ethernet and interoperability with SCE network API  

The original netman API remains and has been tested to work.


israpps:
Is this compatible (or at least an usable foundation?) for system246 network PCB?

uyjulian:
The USB ethernet UE board driver is not implemented in this PR.

rickgaiser:
Nice!

What's the status of "interoperability with SCE network API", and how would we be able to use that?

Is it for using ps2sdk SMAP driver with the SCE TCP/IP stack?

uyjulian:
> Nice!
> 
> What's the status of "interoperability with SCE network API", and how would we be able to use that?
> 
> Is it for using ps2sdk SMAP driver with the SCE TCP/IP stack?

The netdev variant of the SMAP driver has an interface that allows it to be used with the inet driver, and netman has added support for one additional driver with netdev interface (more coming in additional PRs)

PR #826, opened 2026-04-25
Lwip mainstream

fjtrujy:
## Summary

Switch the lwIP dependency from the long-standing **ps2dev/lwip** fork (branch `ps2-v2.0.3`) to **upstream `lwip-tcpip/lwip` `STABLE-2_0_3_RELEASE`**, and absorb the small set of accommodations into ps2sdk so the lwIP source tree stays unmodified going forward.

This is the foundation for moving to newer lwIP releases — it removes the fork as a blocker. A follow-up stacked PR ([fjtrujy:upgrade_lwip_v2](https://github.com/fjtrujy/ps2sdk/tree/upgrade_lwip_v2)) upgrades to lwIP 2.2.1 on top of this.

## Why

The ps2dev fork carried a small set of PS2-specific patches against lwIP 2.0.3 source. Most of them masked workarounds we can express through `lwipopts.h` / glue code, and some are no longer needed at all in current code paths. Owning patches against upstream lwIP makes every future version bump painful (rebases, conflict resolution against a moving target). Switching to clean upstream removes that maintenance tax.

## What's in this PR

7 commits on top of `master`:

```
fix:     [tcpip] use upstream lwIP 2.0.3 without ps2dev patches
fix:     [smap]  handle unaligned buffers in CopyToFIFO/CopyFromFIFO
opt:     [tcpip] disable IP reassembly and fragmentation on IOP
style:   [tcpip] one-object-per-line formatting in ps2api/ps2ip lists
cleanup: [tcpip] prune lwipopts.h entries that restate lwIP 2.0.3 defaults  (IOP)
cleanup: [tcpip] prune ee lwipopts.h entries that restate lwIP 2.0.3 defaults
cleanup: [tcpip] move sys/time.h include from Makefile -include into arch/cc.h
```

### Switch to upstream lwIP 2.0.3
- `download_dependencies.sh`: clone `lwip-tcpip/lwip` at `STABLE-2_0_3_RELEASE` instead of `ps2dev/lwip` at `ps2-v2.0.3`.

### ps2sdk-side accommodations replacing the fork's patches
- `iop/tcpip/tcpip-base/include/arch/cc.h`: pre-include `<sys/time.h>`. Upstream `sockets.h` references `struct timeval` without including it when `LWIP_TIMEVAL_PRIVATE=0`.
- `iop/tcpip/tcpip-base/include/stdlib.h`: minimal IOP shim that defines `atoi` via sysclib's `strtol`. Upstream lwIP expects `atoi` to be declared.
- `iop/tcpip/tcpip-base/include/lwipopts.h`: set `LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT=1` so `mem_free()` uses `SYS_ARCH_PROTECT` rather than a mutex. Required because SMAP RX calls `pbuf_free()` from interrupt-disabled context — same effect as the fork's `mem.c` patch but achieved purely via lwipopts.

### Driver-side fix uncovered by the move
- `iop/network/smap`: handle 2-byte-aligned PBUF_RAM payloads in `CopyToFIFO` / `CopyFromFIFO` using `lwl/lwr` (R3000A has no unaligned word access — pbufs landing on a 2-byte boundary tripped AdEL on real hardware).

### Configuration tightening for the IOP/EE targets
- Disable `IP_REASSEMBLY` and `IP_FRAG`: PS2 networking targets a local LAN with MTU=1500 and TCP negotiates MSS=1460, so fragmentation never arises in the common case. Frees `MEMP_NUM_REASSDATA` / `MEMP_NUM_FRAG_PBUF` pool entries.
- Prune `lwipopts.h` entries that just restate lwIP 2.0.3 defaults (separate cleanups for the IOP and EE variants).
- Style: one-object-per-line formatting in `ps2api`/`ps2ip` Makefile lists.
- Move `sys/time.h` include from the Makefile's `-include` to `arch/cc.h` so it's visible to every consumer of `arch/cc.h`, not just lwIP source.

## What was extracted into smaller already-merged PRs

The earlier version of this PR also bundled a few independent fixes that have since been split out and merged on their own:

- #830 (`fix: place exported variables in .data instead of bogus section("data")`) — load-bearing for the upcoming 2.2.1 upgrade, but applied cleanly to 2.0.3 too.
- #831 (`cleanup: drop dead PRE_LWIP_130_COMPAT branches from smap and tcpip-netman`) — was previously bundled here, extracted as a focused cleanup PR.
- #832 (`fix: [udptty] return size from udp_send for pre-link-up writes`).
- #833 (`fix: [ee/libcglue] correct __retarget_lock_acquire_recursive race`).
- #834 (`fix: [ee/libcglue] pass iop_max_fd+1 to underlying select instead of n`).
- #835 (`fix: [ee/tcpip] wire _libcglue_fdman_inet_ops in _ps2sdk_ps2ipee_init`).
- #836 (`fix: [ps2ip_rpc] explicit fixed-layout select_pkt across EE/IOP RPC`).

This PR is now strictly the "switch to upstream lwIP 2.0.3 and clean up the lwIP options" change.

## Compatibility

- The build still produces `ps2ip-nm.irx` and `ps2ip.irx` with the same export table; downstream IRXs (smap, netman) rebuild without changes to their imports.
- `ps2link` and other consumers continue to work without modification.

## Test plan

- [x] Build the full IOP networking chain: `tcpip-netman` → `smap` → `netman`, install IRXs, build `ps2link`.
- [x] Real hardware: confirmed working — `ps2link execee` and `reset` succeed the same as with the fork's 2.0.3.

## Follow-up

A stacked PR ([fjtrujy:upgrade_lwip_v2](https://github.com/fjtrujy/ps2sdk/tree/upgrade_lwip_v2)) upgrades on top of this branch to lwIP 2.2.1, adding only what's required to make 2.2.1 work on PS2 (MEM_ALIGNMENT alignment match between lwIP and newlib, a recursive-sema sys_arch_protect, and a few API renames).

### DIFF
--- a/ee/network/tcpip/Makefile
+++ b/ee/network/tcpip/Makefile
@@ -35,9 +35,41 @@ endif
 
 EE_INCS += -I$(LWIP)/src/include -I$(LWIP)/src/include/ipv4
 
-ps2api_OBJECTS = api_lib.o api_msg.o api_netbuf.o err.o sockets.o tcpip.o
-ps2api_IPV4 = icmp.o ip.o ip4.o ip4_addr.o ip4_frag.o inet_chksum.o
-ps2ip_OBJECTS = sys.o lwip_init.o mem.o netif.o pbuf.o stats.o tcp_in.o tcp_out.o udp.o memp.o tcp.o ethernet.o etharp.o raw.o def.o timeouts.o $(ps2api_IPV4) $(ps2api_OBJECTS)
+ps2api_OBJECTS = \
+	api_lib.o \
+	api_msg.o \
+	api_netbuf.o \
+	err.o \
+	sockets.o \
+	tcpip.o
+
+ps2api_IPV4 = \
+	icmp.o \
+	ip.o \
+	ip4.o \
+	ip4_addr.o \
+	ip4_frag.o \
+	inet_chksum.o
+
+ps2ip_OBJECTS = \
+	sys.o \
+	lwip_init.o \
+	mem.o \
+	netif.o \
+	pbuf.o \
+	stats.o \
+	tcp_in.o \
+	tcp_out.o \
+	udp.o \
+	memp.o \
+	tcp.o \
+	ethernet.o \
+	etharp.o \
+	raw.o \
+	def.o \
+	timeouts.o \
+	$(ps2api_IPV4) \
+	$(ps2api_OBJECTS)
 
 ifdef PS2IP_DHCP
 ps2ip_OBJECTS += dhcp.o
--- a/download_dependencies.sh
+++ b/download_dependencies.sh
@@ -13,15 +13,15 @@ if [ "x$1" != "xlocked" ]; then
   fi
 fi
 
-## Download LWIP
-LWIP_REPO_URL="https://github.com/ps2dev/lwip.git"
+## Download LWIP (upstream, unpatched)
+LWIP_REPO_URL="https://github.com/lwip-tcpip/lwip.git"
 LWIP_REPO_FOLDER="common/external_deps/lwip"
-LWIP_BRANCH_NAME="ps2-v2.0.3"
+LWIP_BRANCH_NAME="STABLE-2_0_3_RELEASE"
 if test ! -d "$LWIP_REPO_FOLDER"; then
   git clone --depth 1 -b $LWIP_BRANCH_NAME $LWIP_REPO_URL "$LWIP_REPO_FOLDER"_inprogress || exit 1
   mv "$LWIP_REPO_FOLDER"_inprogress "$LWIP_REPO_FOLDER"
 else
-  (cd "$LWIP_REPO_FOLDER" && git fetch origin && git reset --hard "origin/${LWIP_BRANCH_NAME}" && git checkout "$LWIP_BRANCH_NAME" && cd - )|| exit 1
+  (cd "$LWIP_REPO_FOLDER" && git fetch origin --tags && git reset --hard "${LWIP_BRANCH_NAME}" && git checkout "$LWIP_BRANCH_NAME" && cd - )|| exit 1
 fi
 
 ## Download libsmb2
--- a/ee/network/tcpip/src/include/arch/cc.h
+++ b/ee/network/tcpip/src/include/arch/cc.h
@@ -4,6 +4,11 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <stddef.h>
+/* Upstream lwIP's sockets.h references struct timeval without including
+   sys/time.h when LWIP_TIMEVAL_PRIVATE=0. Pull it in via this port header
+   so every lwIP translation unit (and every consumer that #includes
+   arch/cc.h transitively through lwip/opt.h) sees struct timeval. */
+#include <sys/time.h>
 
 #define PACK_STRUCT_FIELD(x) x __attribute((packed))
 #define PACK_STRUCT_STRUCT __attribute((packed))
--- a/ee/network/tcpip/src/include/lwipopts.h
+++ b/ee/network/tcpip/src/include/lwipopts.h
@@ -4,12 +4,6 @@
 #ifndef __LWIPOPTS_H__
 #define __LWIPOPTS_H__
 
-/**
- * NO_SYS==1: Provides VERY minimal functionality. Otherwise,
- * use lwIP facilities.
- */
-#define NO_SYS		0
-
 #define LWIP_TIMEVAL_PRIVATE 0
 
 /* ---------- Thread options ---------- */
@@ -41,43 +35,16 @@
  */
 #define TCPIP_THREAD_PRIO		DEFAULT_THREAD_PRIO
 
-/**
- * SLIP_THREAD_STACKSIZE: The stack size used by the slipif_loop thread.
- * The stack size value itself is platform-dependent, but is passed to
- * sys_thread_new() when the thread is created.
- */
-#define SLIPIF_THREAD_STACKSIZE		DEFAULT_THREAD_STACKSIZE
-
-/**
- * SLIPIF_THREAD_PRIO: The priority assigned to the slipif_loop thread.
- * The priority value itself is platform-dependent, but is passed to
- * sys_thread_new() when the thread is created.
- */
-#define SLIPIF_THREAD_PRIO		DEFAULT_THREAD_PRIO
-
-/**
- * PPP_THREAD_STACKSIZE: The stack size used by the pppInputThread.
- * The stack size value itself is platform-dependent, but is passed to
- * sys_thread_new() when the thread is created.
- */
-#define PPP_THREAD_STACKSIZE		DEFAULT_THREAD_STACKSIZE
-
-/**
- * PPP_THREAD_PRIO: The priority assigned to the pppInputThread.
- * The priority value itself is platform-dependent, but is passed to
- * sys_thread_new() when the thread is created.
- */
-#define PPP_THREAD_PRIO			DEFAULT_THREAD_PRIO
-
 /*
    ------------------------------------
    ---------- Memory options ----------
    ------------------------------------
 */
 /**
- * MEM_LIBC_MALLOC==1: Use malloc/free/realloc provided by your C-library
- * instead of the lwip internal allocator. Can save code size if you
- * already use it.
+ * MEM_LIBC_MALLOC==1: Use malloc/free/realloc provided by the C-library
+ * instead of lwIP's internal allocator. Default is 0; enabled on EE
+ * because newlib's malloc is already linked in and the heap pool is
+ * cheaper than a duplicate lwIP heap.
  */
 #define MEM_LIBC_MALLOC		1
 
@@ -115,12 +82,11 @@
 /**
  * MEMP_NUM_TCPIP_MSG_INPKT: the number of struct tcpip_msg, which are used
  * for incoming packets.
- * (only needed if you use tcpip.c)
+ * SP193: this should be around the size of the TCP window because the
+ * TCPIP thread may take a while to execute (non-preemptive multitasking),
+ * otherwise incoming frames may get dropped.
  */
-//SP193: this should be around the size of the TCP window because the TCPIP thread may take a while to execute (non-preemptive multitasking), otherwise incoming frames may get dropped.
-#ifndef LWIP_TCPIP_CORE_LOCKING_INPUT
 #define MEMP_NUM_TCPIP_MSG_INPKT	50
-#endif
 
 /**
  * MEMP_NUM_TCPIP_MSG_API: the number of struct tcpip_msg, which are used
@@ -145,13 +111,6 @@
  * interrupt context!
  */
 #define LWIP_TCPIP_CORE_LOCKING_INPUT	1
-
-/** SYS_LIGHTWEIGHT_PROT
- * define SYS_LIGHTWEIGHT_PROT in lwipopts.h if you want inter-task protection
- * for certain critical regions during buffer allocation, deallocation and
- * memory allocation and deallocation.
- */
-#define SYS_LIGHTWEIGHT_PROT	1
 
 /*
    ---------------------------------
--- a/iop/network/smap/src/xfer.c
+++ b/iop/network/smap/src/xfer.c
@@ -51,10 +51,23 @@ static inline void CopyFromFIFO(volatile u8 *smap_regbase, void *buffer, unsigne
 
     SMAP_REG16(SMAP_R_RXFIFO_RD_PTR) = RxBdPtr;
 
-    result = SmapDmaTransfer(smap_regbase, buffer, length, DMAC_TO_MEM);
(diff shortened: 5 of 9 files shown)


sp193:
Is it really a good idea, to work around unaligned memory accesses in SMAP? I recall that LWIP had some options to deal with alignment limitations of some hosts, but it has been a very long time since I touched this, indeed.

Copying with the IOP is very slow.

sp193:
Re: LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT

I may be wrong if things have changed, but we are not always using a mutex because semaphores are very slow on the IOP, compared to just disabling interrupts. Signalling a semaphore from an interrupt context is possible with iSignalSema().

fjtrujy:
> Re: LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT
> 
> I may be wrong if things have changed, but we are not always using a mutex because semaphores are very slow on the IOP, compared to just disabling interrupts. Signalling a semaphore from an interrupt context is possible with iSignalSema().

Good question.... I didn't think about that.
I was just looking to use mainstream changes to make future upgrades easier....
Anyway, I didn't notice any slowdown, but it is true that the only test that I did was using `ps2link` & a `transfer_speed test`

```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/time.h>
#include <fcntl.h>
#include <stdint.h>
#include <errno.h>
#include <string.h>

#define BUF_SIZE 32768

void SleepThread(void);

static double time_diff(const struct timeval *start, const struct timeval *end) {
    return (end->tv_sec - start->tv_sec) + (end->tv_usec - start->tv_usec) / 1e6;
}

static void run_read_mode(const char *path) {
    int fd = open(path, O_RDONLY);
    if (fd < 0) {
        fprintf(stderr, "open %s: %s\n", path, strerror(errno));
        return;
    }

    char *buf = malloc(BUF_SIZE);
    if (!buf) {
        fprintf(stderr, "malloc failed\n");
        close(fd);
        return;
    }

    ssize_t r;
    uint64_t total = 0;
    struct timeval start, now, last_print;
    gettimeofday(&start, NULL);
    last_print = start;
    uint64_t interval_bytes = 0;

    printf("-- READ mode (read syscall) --\n");
    while ((r = read(fd, buf, BUF_SIZE)) > 0) {
        total += (uint64_t)r;
        interval_bytes += (uint64_t)r;
        gettimeofday(&now, NULL);
        double since_last = time_diff(&last_print, &now);
        if (since_last >= 1.0) {
            double instant_mib = (interval_bytes / (1024.0 * 1024.0)) / since_last;
            double elapsed = time_diff(&start, &now);
            double avg_mib = (total / (1024.0 * 1024.0)) / (elapsed > 0 ? elapsed : 1e-9);
            printf("Elapsed: %.2fs  Instant: %.3f MiB/s  Avg: %.3f MiB/s  Read: %lu bytes\n",
                   elapsed, instant_mib, avg_mib, (unsigned long)total);
            fflush(stdout);
            last_print = now;
            interval_bytes = 0;
        }
    }

    gettimeofday(&now, NULL);
    double total_secs = time_diff(&start, &now);
    if (r < 0) {
        fprintf(stderr, "read error: %s\n", strerror(errno));
    }
    printf("READ Done. Read %lu bytes in %.6fs (%.3f MiB/s)\n\n",
           (unsigned long)total, total_secs, (total / (1024.0 * 1024.0)) / (total_secs > 0 ? total_secs : 1e-9));

    free(buf);
    close(fd);
}

static void run_fread_mode(const char *path) {
    FILE *f = fopen(path, "rb");
    if (!f) {
        fprintf(stderr, "fopen %s failed: %s\n", path, strerror(errno));
        return;
    }

    char *buf = malloc(BUF_SIZE);
    if (!buf) {
        fprintf(stderr, "malloc failed\n");
        fclose(f);
        return;
    }

    size_t r;
    uint64_t total = 0;
    struct timeval start, now, last_print;
    gettimeofday(&start, NULL);
    last_print = start;
    uint64_t interval_bytes = 0;

    printf("-- FREAD mode (stdio fread) --\n");
    while ((r = fread(buf, 1, BUF_SIZE, f)) > 0) {
        total += (uint64_t)r;
        interval_bytes += (uint64_t)r;
        gettimeofday(&now, NULL);
        double since_last = time_diff(&last_print, &now);
        if (since_last >= 1.0) {
            double instant_mib = (interval_bytes / (1024.0 * 1024.0)) / since_last;
            double elapsed = time_diff(&start, &now);
            double avg_mib = (total / (1024.0 * 1024.0)) / (elapsed > 0 ? elapsed : 1e-9);
            printf("Elapsed: %.2fs  Instant: %.3f MiB/s  Avg: %.3f MiB/s  Read: %lu bytes\n",
                   elapsed, instant_mib, avg_mib, (unsigned long)total);
            fflush(stdout);
            last_print = now;
            interval_bytes = 0;
        }
    }

    gettimeofday(&now, NULL);
    double total_secs = time_diff(&start, &now);
    if (ferror(f)) {
        fprintf(stderr, "fread error\n");
    }
    printf("FREAD Done. Read %lu bytes in %.6fs (%.3f MiB/s)\n\n",
           (unsigned long)total, total_secs, (total / (1024.0 * 1024.0)) / (total_secs > 0 ? total_secs : 1e-9));

    free(buf);
    fclose(f);
}

int main(int argc, char **argv) {
    const char *path = "testfile.bin";
    const char *mode = "both"; // read, fread, both
    if (argc > 1) path = argv[1];
    if (argc > 2) mode = argv[2];

    if (strcmp(mode, "read") == 0) {
        run_read_mode(path);
    } else if (strcmp(mode, "fread") == 0) {
        run_fread_mode(path);
    } else {
        run_read_mode(path);
        run_fread_mode(path);
    }

    SleepThread(); // Prevent exit to see final output on PS2
    return 0;
}
```

sp193:
I was referring to the comment you made in the code. There are various good reasons to enable this setting.

If you revert enough settings (including turning off core locking), you should eventually end up with an LWIP version that performs incredibly poorly on the IOP.

Personally, I feel that we need to strike a balance between using the stock code and providing a custom version. This isn't a modern game console, so the code not fitting its architecture can be a very bad thing.

Re: the PBUF alignment problem that I commented on above. I wrote a comment about how it was fixed, a long time ago. Without this change, the memory buffer will be unaligned.
```
pbuf_alloc() will no longer align the payload field, for PBUF_RAM allocations.
	This is so that frames being output will not be unaligned.

	Note: this still doesn't solve all cases. For example, generating a UDP packet will result in an unaligned final address,
	as PBUF_TRANSPORT is a general "one size fits all" option.
```

Looking at that changelog, you should be careful because some changes came about due to incompatibilities with how the SCE IOP kernel works.

uyjulian:
> Is it really a good idea, to work around unaligned memory accesses in SMAP? I recall that LWIP had some options to deal with alignment limitations of some hosts, but it has been a very long time since I touched this, indeed.
> 
> Copying with the IOP is very slow.

It avoids some difficult-to-diagnose issues such as hanging. Speed can come later.  
(I don't see a major difference in the speed in this PR, measured with iperf, anyways)

> Re: LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT
> 
> I may be wrong if things have changed, but we are not always using a mutex because semaphores are very slow on the IOP, compared to just disabling interrupts. Signalling a semaphore from an interrupt context is possible with iSignalSema().

`LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT` **enabled** (as in this PR) does **not** use a mutex and instead disables/enables interrupts.

fjtrujy:
Thanks for the careful review @sp193 — let me try to address both concerns, since I think we're actually aligned on what we want.

> Re: LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT and the comment in code

@uyjulian is right — when this option is =1, lwIP's `mem_free()` uses `SYS_ARCH_PROTECT(lev)` for the critical section, not a mutex. And on our port, `SYS_ARCH_PROTECT` resolves to `CpuSuspendIntr()` — see `iop/tcpip/tcpip-base/sys_arch.c:355`:
```
sys_prot_t sys_arch_protect(void)
{
    sys_prot_t Flags;
    CpuSuspendIntr(&Flags);
    return Flags;
}
```
So enabling this flag is exactly the **"disable interrupts instead of taking a semaphore"** path you're describing. 
With the flag=0 (the default), `lwIP` would wrap `mem_free()` in a `sys_mutex_t`, which on the `IOP` means a semaphore — that's the slow path we want to avoid.

If the in-code comment reads like we're introducing a mutex, I'm happy to reword it to make the direction explicit (e.g., "uses `SYS_ARCH_PROTECT` (interrupt disable on IOP) rather than wrapping the critical section in a sys_mutex_t/semaphore"). Let me know if that helps.

> Re: handling unaligned buffers in CopyToFIFO/CopyFromFIFO

Your historical pbuf-alignment comment is actually the reason I needed this commit. The ps2dev fork has a patch to `pbuf_alloc()` that adjusts payload alignment so that `PBUF_RAM` allocations land in a way that keeps the on-wire frame aligned. By moving to stock `lwIP` 2.0.3, that patch is gone — so there is now a real path where the PIO copy hits a halfword-aligned payload and the IOP traps an AdEL exception (reproducible: ExcCode=4, BadVAddr ending in 2/6/A/E, on the PIO fallback when len<64).

I agree pure IOP copy is slow, but:
- The aligned fast path is unchanged, so the common case has zero added cost.
- GCC turns the byte-wise word assembly into lwl/lwr, so even the slow path is reasonably tight (no per-byte function call overhead).
- @uyjulian measured with iperf and didn't see a meaningful delta.

My goal in this PR is exactly to stop patching the lwIP source so future bumps stay easy.

> Re: balance between stock and custom
Totally agree — that's the line I'm trying to walk here. The places where the IOP architecture genuinely conflicts with stock lwIP interrupt-context `pbuf_free`, `alignment`, `no <sys/time.h> from sockets.h`) are kept as accommodations outside the lwIP tree. 

The settings I pruned in the `lwipopts` cleanup commits are only the ones that already match upstream defaults or configure dead code — I verified the installed `ps2ip-nm.irx` is byte-identical (sha unchanged) before/after each cleanup commit, so nothing performance-critical was inadvertently flipped.

Happy to revisit any specific commit if you'd like — and to reword the lwipopts comment to be less ambiguous.

uyjulian (approved):
Aging tested with TCP iperf for 12 hours, works fine

fjtrujy:
@uyjulian @sp193 , you can review it again.

ee/network/tcpip/Makefile
```diff
@@ -18,10 +18,6 @@ EE_LIB_ALTNAMES ?= libsocket.a
 EE_LIB = libps2ip.a
 
 EE_CFLAGS += -Wall -fno-builtin -DLWIP_NOASSERT -DLWIP_COMPAT_MUTEX -DLWIP_COMPAT_MUTEX_ALLOWED -mno-gpopt -O2 -G0 -mno-check-zero-division
-# Upstream lwIP 2.0.3's sockets.h references struct timeval without
-# including <sys/time.h> when LWIP_TIMEVAL_PRIVATE=0. Pre-include it
-# for every compilation unit so we don't have to patch lwIP.
-EE_CFLAGS += -include sys/time.h
```
uyjulian: Might want to clean up this change as it was added in https://github.com/ps2dev/ps2sdk/pull/826/changes/a24848a50a04f9f21ea9ccfbe2dae98819b1577b#diff-30ba5e8172a8e02c1b72294258d837cf522e9e5c5055f4e851f1c9e6b81106deR24

fjtrujy: Not sure if I follow you, I think you are commenting on an old state of the PR. This change is no longer in the makefile, and it is instead in the header files

uyjulian: The change was added in https://github.com/ps2dev/ps2sdk/commit/a24848a50a04f9f21ea9ccfbe2dae98819b1577b#diff-30ba5e8172a8e02c1b72294258d837cf522e9e5c5055f4e851f1c9e6b81106deR24 and removed in this commit, both in the same PR.

fjtrujy: Ok, you mean to do a `rebase -i`

uyjulian: Yeah, when rebasing squash this commit against https://github.com/ps2dev/ps2sdk/commit/a24848a50a04f9f21ea9ccfbe2dae98819b1577b

fjtrujy: Done — squashed `2cbe8dc3a4` into `a24848a50a` via `git rebase -i fixup`, so the Makefile `-include sys/time.h` flag is no longer added then removed within the same PR. The single resulting commit puts `#include <sys/time.h>` directly in `ee/network/tcpip/src/include/arch/cc.h` and `iop/tcpip/tcpip-base/include/arch/cc.h` from the start. Force-pushed (the squashed commit is the new tip of this PR's history; only the cleanup commit is gone, all other 6 commits remain).

uyjulian (approved):
lgtm

PR #828, opened 2026-05-05
Invert condition for ACMEME status on init

israpps:
condition is ok, but original code sorrounds startup code on condition instead of returning err


uyjulian:
Nice catch

PR #830, opened 2026-05-07
fix: place exported variables in `.data` instead of bogus `section("data")`

fjtrujy:
## Summary

Two declarations in the IOP tree carried `__attribute__((section("data")))` — note the **bare "data" name with no leading dot**. The IRX linker script collects input sections that match standard names (`.text`, `.data`, `.rodata`, `.bss`, …); a `"data"` section is an *orphan*: the linker keeps it, but it lands outside any `PT_LOAD` segment, and `srxfixup` warns

> section 'data' needs allocation but not in program segment

The IRX loader doesn't allocate orphans into IOP RAM, so any write to such a variable from inside the IRX lands at a bogus address.

This PR fixes both occurrences. Two commits, one issue.

### 1. `fix: [errno] use ".data" section so the IRX loader allocates errno`

`common/include/errno.h` declared `extern int errno` with `section("data")`. This was a *latent* bug for years because lwIP 2.0.3 only wrote `errno` when `LWIP_SOCKET_SET_ERRNO=1` (our `lwipopts.h` had it `0`). lwIP commit [0ee6ad0a](https://github.com/lwip-tcpip/lwip/commit/0ee6ad0a) (Jul 2017, *Removed LWIP_SOCKET_SET_ERRNO*) deleted the escape hatch — every socket error path now writes `errno` unconditionally — so from lwIP 2.1.0 onward the corruption fires. Confirmed via `git bisect` across the 1384 commits between `STABLE-2_0_3_RELEASE` and `STABLE-2_1_0_RELEASE`: `0ee6ad0a` is the first bad commit.

Fix: `section("data")` → `section(".data")`. One character.

### 2. `cleanup: [smap] drop bogus section("data") from extern decls`

`iop/network/smap/src/main.c` declared `_exp_smap` and `_exp_smapmodu` with the same `section("data")` annotation. Same root cause.

In this case the annotation was *always* harmless because the matching `DECLARE_EXPORT_TABLE` in `iop/kernel/include/irx.h` places the actual definition in `.text` via a `section(".text\n\t#")` asm trick — and section attributes on `extern` declarations are hints only, the symbol's address is governed by its definition. The IRX layout is unchanged (verified with `nm`: `_exp_smap` is still in `.text`).

The cleanup just removes the misleading attribute so the latent bug can't manifest if the export-table macro or its placement is ever refactored. `srxfixup` no longer warns about smap.

## Test plan

- [ ] CI build of ps2sdk and the ps2-packages stack.
- [x] No `srxfixup: section 'data' needs allocation but not in program segment` warnings remain when building IOP IRX modules.
- [x] `nm smap.irx` shows `_exp_smap` in `.text` (unchanged).
- [x] Networked apps that exercise `errno` writes from inside an IRX (DHCP failures, EBADFD on a closed socket) no longer corrupt random IOP RAM.

### DIFF
--- a/common/include/errno.h
+++ b/common/include/errno.h
@@ -428,6 +428,6 @@ char *file_errors[] = {
 #define error_to_string(errnum) (file_errors[errnum * -1])
 #endif
 
-extern int errno __attribute__((section("data")));
+extern int errno __attribute__((section(".data")));
 
 #endif /* __ERRNO_H__ */
--- a/iop/network/smap/src/main.c
+++ b/iop/network/smap/src/main.c
@@ -270,12 +270,17 @@ void PS2IPLinkStateDown(void)
 #endif
 
 #if defined(BUILDING_SMAP_NETMAN) || defined(BUILDING_SMAP_NETDEV)
-// While the header of the export table is small, the large size of the export table (as a whole) places it in data instead of sdata.
-extern struct irx_export_table _exp_smap __attribute__((section("data")));
+/* DECLARE_EXPORT_TABLE in iop/kernel/include/irx.h places the export
+   table in .text via a section(".text\n\t#") asm trick, so the extern
+   declaration needs no section attribute. (The previous section("data")
+   was misleading: a bare "data" name produces an orphan ELF section the
+   IRX loader does not allocate; harmless here only because the actual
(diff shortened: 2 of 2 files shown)


uyjulian (approved):
lgtm

PR #833, opened 2026-05-07
fix: [ee/libcglue] correct `__retarget_lock_acquire_recursive` race

fjtrujy:
## Summary

`__retarget_lock_acquire_recursive` (and its `try_` sibling) had a race window between the recursion-counter bump and the underlying `WaitSema`. Under preemptive scheduling on the EE — increasingly likely now that the `MEM_LIBC_MALLOC=1` networking stack pushes multi-threaded malloc/free traffic through the same recursive lock — a higher-priority thread could observe a half-set lock state and end up double-waiting on a sema that nobody would ever signal again.

## Root cause

The previous implementation set ownership state **before** taking the gate:

```c
starting = lock->count == 0;
if (starting) lock->thread_id = thread_id;   // ← racy
if (lock->thread_id == thread_id) {
    lock->count++;
    if (starting) WaitSema(lock->sem_id);
} else {
    WaitSema(lock->sem_id);
    lock->thread_id = thread_id;
    lock->count++;
}
```

If a higher-priority thread preempted between `count++` and the *first*-acquirer's `WaitSema`, it saw `lock->count > 0` with `thread_id == self` (or matched the original owner), took the wrong branch, blocked, and never got woken.

## The fix

Standard recursive-mutex shape:

- **Fast path** — if this thread already owns the gate (`count > 0 && thread_id == self`), increment `count` and return. Safe to read those fields without the gate, because only the owning thread mutates them.
- **Slow path** — `WaitSema` first, *then* set `thread_id` + `count = 1`. By the time we touch ownership state we hold the gate, so the writes are atomic with respect to other contenders.

Same logic applied to `__retarget_lock_try_acquire_recursive` (uses `PollSema` and returns `1` on contention).

## Test plan

- [x] PCSX2 multi-thread `malloc`/`free` stress test (`samples/malloc_stress` in [`ps2_drivers`](https://github.com/fjtrujy/ps2_drivers)): 4 worker threads × 78 k operations / sec for 11 s = ~870 k mixed-size alloc/free. **0 freelist corruptions, 0 deadlocks.**
- [x] Real-hardware HTTP burst (`ps2_http`): 100/100 HTTP 200 responses without wedging.
- [x] No regression in `select_test_ee` / `tcp_echo_test_ee` / `tcp_burst_client_ee`.

### DIFF
--- a/ee/libcglue/src/lock.c
+++ b/ee/libcglue/src/lock.c
@@ -140,26 +140,26 @@ void __retarget_lock_acquire(_LOCK_T lock)
 #ifdef F___retarget_lock_acquire_recursive
 void __retarget_lock_acquire_recursive(_LOCK_T lock)
 {
-    bool starting = false;
     int32_t thread_id = GetThreadId();
-    starting = lock->count == 0;
-    if (starting) {
-        lock->thread_id = thread_id;
-    }
-    if (lock->thread_id == thread_id) {
+
+    /* Recursive case: this thread already owns the gate. We can read
+     * thread_id/count without the gate because only the owning thread
+     * mutates them while it holds the gate, and only the owning thread
+     * can take this branch. */
+    if (lock->count > 0 && lock->thread_id == thread_id) {
         lock->count++;
-        if (starting) {
-            WaitSema(lock->sem_id);
-        }
-    } else {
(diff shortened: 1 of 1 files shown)


uyjulian (approved):
lgtm

PR #834, opened 2026-05-07
fix: [ee/libcglue] pass iop_max_fd+1 to underlying select instead of n

fjtrujy:
## Summary

`libcglue/select.c` builds the IOP-side fd_sets by translating each EE-side fd to its `iop_fd` via `ps2sdk_get_iop_fd()` and setting the bit at the **iop_fd** position. It then forwarded `n` (the EE-side `nfds`) to the underlying socket-implementation's `select()`. That's wrong: the bits live at iop_fd positions, not EE fd positions, so if any `iop_fd > n - 1` the underlying select scans an empty range and returns 0 immediately — or, on impls that block, waits forever on bits it isn't watching.

## The fix

Compute `iop_max_fd` while building the IOP fdsets and pass `iop_max_fd + 1` as `nfds` to the underlying `select()`. The bits and the bound now agree.

## Drive-by robustness tightening

While in the same loop, two related improvements:

1. The post-select translation loop didn't have the `if (!in_read && !in_write && !in_except) continue` guard the pre-select loop has — added it for symmetry.
2. The pre-select loop returned `-1 / EBADF` if any single fd in the input sets failed to translate to an `iop_fd`. That made one closed/foreign fd poison the whole `select()` call. Now we skip the fd gracefully and continue building the IOP fdsets from the remaining fds.

These together are why the branch name is `…-iop-max-fd-plus-one` even though three small changes are bundled — they're tightly coupled to the same code path. Happy to split into two commits if a reviewer prefers.

## Test plan

- [x] `select_test_ee` (in [`ps2_drivers`](https://github.com/fjtrujy/ps2_drivers)) — listening-socket select probe shows `rc=0 r=0 w=0 e=0` ticks, no spurious wakes.
- [x] mongoose's poll loop (`ps2_http`) sustains 100/100 HTTP requests without `select` returning -1.
- [x] No regression observed in `tcp_echo_test_ee` (loopback echo across 100+ iterations).

### DIFF
--- a/ee/libcglue/src/select.c
+++ b/ee/libcglue/src/select.c
@@ -22,6 +22,7 @@ int	select(int n, fd_set *readfds, fd_set *writefds, fd_set *exceptfds, struct t
 {
 	int fd;
 	int ret;
+	int iop_max_fd = -1;
 	fd_set ready_readfds, ready_writefds, ready_exceptfds;
 	fd_set iop_readfds, iop_writefds, iop_exceptfds;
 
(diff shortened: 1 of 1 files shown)


fjtrujy:
Ping @F0bes as he found the issue originally

uyjulian (approved):
This codepath was not tested much originally, but this looks good now.

lgtm

PR #836, opened 2026-05-07
fix: [ps2ip_rpc] explicit fixed-layout `select_pkt` across EE/IOP RPC

fjtrujy:
## Summary

`select_pkt` (the SIF-RPC payload that `ee/rpc/tcpips` sends to `iop/tcpip/tcpips` for `select()`) used `struct fd_set` and `struct timeval` directly in the wire layout. Both types have **compiler-defined sizes that diverge between the EE and IOP compile environments**, so the EE and IOP saw different field offsets in the same struct — a silent ABI mismatch.

## The mismatch

| Field           | EE (newlib)                                    | IOP (lwIP)                                  |
|-----------------|------------------------------------------------|---------------------------------------------|
| `struct fd_set` | ~8 bytes (`FD_SETSIZE=64`, `__fd_mask=ulong=8`)| 2 bytes (packed `fd_bits[(MEMP_NUM_NETCONN+7)/8]`) |
| `struct timeval`| 16 bytes (`tv_sec`/`tv_usec` 64-bit longs)     | 8 bytes (32-bit longs)                      |

The two mismatches together meant every field of `select_pkt` after `timeout` lived at a different offset on each side. The EE wrote `pkt->writeset_p / exceptset_p / writeset / exceptset` at offsets the IOP never read; the IOP read garbage from the high half of the EE's `tv_usec` for what it thought were the `writeset_p` / `exceptset_p` pointers, ended up calling `lwip_select` with `NULL` fd-sets, and silently returned without modifying any bits.

This was harmless for **ps2link** (UDP only, never calls `select`), but broke **mongoose-based servers on the IOP-side path** (`ps2_http`): the listening socket was flagged as having an exception condition on the very first poll cycle (spurious eset bits), mongoose called `mg_error("socket error")`, and the listener was killed before the first client connection.

## The fix

Lock the wire layout to a fixed shape that's identical on both compile units:

- Define `ps2ip_rpc_fd_set` as a fixed-size byte array `unsigned char fd_bits[(MEMP_NUM_NETCONN + 7) / 8]`. Size is driven by `MEMP_NUM_NETCONN` and matches lwIP's internal `fd_set` semantics.
- Replace `struct timeval timeout` in `select_pkt` with explicit `s32 timeout_sec; s32 timeout_usec;` fields. Each side reconstructs/decomposes its own native `timeval` at the boundary.

Net result: every field of `select_pkt` is at the same byte offset on EE and IOP, regardless of toolchain header definitions.

## Files

- `common/include/ps2ip_rpc.h`     — fixed-layout struct
- `ee/rpc/tcpips/src/ps2ipc.c`     — EE side: convert local fd_set/timeval to wire shape
- `iop/tcpip/tcpips/src/ps2ips.c`  — IOP side: convert wire shape to local fd_set/timeval

## Test plan

- [x] `select_test_iop` (in [`ps2_drivers`](https://github.com/fjtrujy/ps2_drivers)) — listening-socket select probe shows `rc=0 r=0 w=0 e=0` ticks (no spurious eset bits). Pre-fix: `e=1` on the very first probe.
- [x] mongoose `ps2_http` on the IOP-side path — sustains repeated HTTP bursts on real hardware without the listener being killed by `mg_error("socket error")`.
- [x] No regression in ps2link / ps2client (UDP) which don't exercise this path.

### DIFF
--- a/common/include/ps2ip_rpc.h
+++ b/common/include/ps2ip_rpc.h
@@ -132,6 +132,23 @@ typedef struct
     u8 buffer[128];
 } setsockopt_pkt;
 
+/* On-the-wire fd_set for the select RPC. The struct fd_set seen on EE
+ * (newlib, ~128 bytes) and IOP (lwIP, MEMP_NUM_NETCONN/8 bytes) compile
+ * units differs in size, which used to scramble select_pkt's layout
+ * across the SIF RPC boundary — IOP would only touch the first few
+ * bytes of EE's wider readset and never see writeset/exceptset at the
+ * right offsets. Using a fixed-size struct here keeps select_pkt
+ * byte-identical on both sides regardless of which struct fd_set is
+ * in scope at the include site. */
+typedef struct
+{
+    unsigned char fd_bits[(MEMP_NUM_NETCONN + 7) / 8];
+} ps2ip_rpc_fd_set;
+
+/* struct timeval differs across the boundary too — newlib's tv_sec/tv_usec
+ * are 64-bit on EE while the IOP defines them as 32-bit longs. Carry
+ * explicit 32-bit fields here so the layout is byte-identical on both
+ * sides; each end reconstructs its native struct timeval. */
 typedef struct
 {
     union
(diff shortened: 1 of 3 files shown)


uyjulian (approved):
lgtm

PR #838, opened 2026-05-08
lwIP: upgrade to upstream STABLE-2_2_1_RELEASE

fjtrujy:
## Summary

Bump the vendored lwIP from `STABLE-2_0_3_RELEASE` to `STABLE-2_2_1_RELEASE` (~7 years of upstream fixes/features) on both the IOP and EE networking stacks.

This is the follow-up to the now-merged #826 (`Lwip mainstream`), which switched the dependency from the ps2dev fork to upstream lwIP 2.0.3 and prepared the `lwipopts.h` and port headers for a clean upgrade. With those prerequisites in place, this PR is **the upgrade itself, nothing else** — no tuning, no opt-in features.

## What's in this PR

A single commit with the version bump and the source-tree adjustments needed to track upstream's API and to keep the EE/IOP stacks working under 2.2.1's new defaults.

### lwipopts.h (EE + IOP)

- `LWIP_TCPIP_CORE_LOCKING=1` + `LWIP_TCPIP_CORE_LOCKING_INPUT=1` (lwIP 2.2.1 default). Socket / netconn API calls take a binary-sem mutex on the calling app thread instead of round-tripping through the tcpip thread mailbox. lwIP releases the lock before any blocking I/O wait so the tcpip thread + netif input still make progress. Saves a context switch + sem wait per API call versus message passing.
- `DEFAULT_THREAD_STACKSIZE=0x600` on IOP (matches the historical lwIP 2.0.3 budget; the deep socket call chains run on app threads under core-locking, the tcpip thread only handles timers + tcpip_callback dispatch, max ~450 B measured via `-fstack-usage`).
- `LWIP_DHCP_DOES_ACD_CHECK=0` + `LWIP_ACD=0`: no AutoIP, controlled LAN, the conflict-detection timer/code is dead weight here.
- Pruned three options removed/renamed in upstream:
  - `LWIP_SOCKET_SET_ERRNO` (gone since 2017's commit `0ee6ad0a`),
  - `DHCP_DOES_ARP_CHECK` (renamed to `LWIP_DHCP_DOES_ACD_CHECK` in 2.2.0),
  - `LWIP_DHCP_CHECK_LINK_UP` (no longer referenced anywhere in 2.2.1).

### EE `lwipopts.h` MEM_ALIGNMENT
Drop from 64 (historical "EE cache design" value, attributed to SP193) to **16**, matching the alignment newlib's `malloc` returns on EE. With `MEM_ALIGNMENT=64`, lwIP 2.2.1's `pbuf_alloc(PBUF_RAM)` computes
```c
payload = LWIP_MEM_ALIGN(p + SIZEOF_STRUCT_PBUF + offset)
```
inside an allocation sized assuming `p` is already 64-byte aligned. Newlib only guarantees 16, set in `newlib/newlib/configure.host:254`:
```
mips64r5900*)
    machine_dir=r5900
    newlib_cflags="${newlib_cflags} -DMALLOC_ALIGNMENT=16"
    ;;
```
so the payload pointer slid past the allocation by up to 48 bytes and scribbled into the next chunk's header — TLB misses inside `_malloc_r` / `_free_r` after the first close cycle of any TCP server (real hw locks up; PCSX2 logs the misses but limps on).

`PBUF_POOL` pbufs (the only ones touched by IOP→EE DMA + cache invalidate) come from memp's static pools and are 64-byte aligned regardless of `MEM_ALIGNMENT`, so the cache-design invariant the old comment cited is preserved by memp, not by `MEM_ALIGNMENT`. The IOP-side `lwipopts.h` has used `MEM_ALIGNMENT=4` forever for the same reason.

### IOP API renames

- `tcpip_callback_with_block` was promoted to `tcpip_callback` (the macro became the real function in lwIP 2.1.0); update `exports.tab`, `imports.lst`, and call sites.
- `iop/network/smap/src/imports.lst`: corresponding rename for the netif callbacks the smap driver imports from `ps2ip-nm.irx`.
- `ps2ip.c`: define `int errno` in `.data` so socket error paths have somewhere to write to (lwIP 2.2.1 always writes errno; the section attribute was already corrected in PR #830).

### IOP `tcpip-base/sys_arch.c`

- Add `sys_mbox_trypost_fromisr` stub (new in lwIP 2.2.1, called from `tcpip_input` when `LWIP_TCPIP_CORE_LOCKING_INPUT=0`; harmless in our build but the symbol must be present).

### EE `tcpip/sys_arch.c`

- Replace the previous DI/EI-based `sys_arch_protect` with a per-thread recursive semaphore. The DI/EI variant deadlocked the EE: any code path inside a `SYS_ARCH_PROTECT` region that ended up calling newlib's `malloc/free` would `WaitSema` on the heap recursive mutex with interrupts disabled. The sema-based variant lets nested waits work normally and removes the EE-specific incompatibility between lwIP and any other library that uses newlib's locks. lwIP allows `SYS_ARCH_PROTECT` to nest, so the implementation tracks the owning thread + a recursion counter and only Wait/Signal on the outermost transitions.

### IOP `tcpip/Makefile` + `ps2api_IPV4` list

- `acd.c` (new in lwIP 2.2.0) intentionally not added; `LWIP_ACD=0` makes it a no-op TU and we'd rather drop the few KB outright.

## Compatibility

- IOP IRX export tables for `ps2ip-nm.irx` are byte-stable except for the `tcpip_callback_with_block → tcpip_callback` rename. `smap.irx` rebuilds against the renamed import without code changes.
- EE socket API is unchanged for consumers (libcglue's wrappers cover the rare call-site renames).
- ps2link / ps2client unchanged; existing UDP and TCP applications continue to work.

## Test plan

- [x] `ps2link execee` and reset cycle on real hardware.
- [x] EE-side network samples in [`ps2_drivers`](https://github.com/fjtrujy/ps2_drivers): `network_ee_sample` (DHCP), `select_test_ee` (no spurious wakes), `tcp_echo_test_ee` (123 self-loopback echoes), `tcp_burst_client_ee` (20/20 to host echo), `udp_echo_server_ee` (100/100 to host echo), `tcp_server_ee` (5/5 server-side close cycles). All in PCSX2.
- [x] **`ps2_http` 100/100 burst on real hardware** with this branch only (no loopback/PCB tuning) — the original wedge that motivated the alignment fix is fully closed.
- [x] PCSX2 sweep across the same samples + ps2_http: 0 TLB misses, 0 mongoose errors.

## Follow-up

Two follow-up PRs (kept separate per review feedback so this PR is "upgrade, nothing else"):

- [`tune/ee-lwip-loopback-pcb`](https://github.com/fjtrujy/ps2sdk/tree/tune/ee-lwip-loopback-pcb) — enable `LWIP_NETIF_LOOPBACK` (~8 B static cost; needed by the `tcp_echo_test_ee` diagnostic) and bump `MEMP_NUM_TCP_PCB` to 32 (~3.8 KB BSS; defensive headroom for server-active-close workloads).

### DIFF
--- a/download_dependencies.sh
+++ b/download_dependencies.sh
@@ -16,7 +16,7 @@ fi
 ## Download LWIP (upstream, unpatched)
 LWIP_REPO_URL="https://github.com/lwip-tcpip/lwip.git"
 LWIP_REPO_FOLDER="common/external_deps/lwip"
-LWIP_BRANCH_NAME="STABLE-2_0_3_RELEASE"
+LWIP_BRANCH_NAME="STABLE-2_2_1_RELEASE"
 if test ! -d "$LWIP_REPO_FOLDER"; then
   git clone --depth 1 -b $LWIP_BRANCH_NAME $LWIP_REPO_URL "$LWIP_REPO_FOLDER"_inprogress || exit 1
   mv "$LWIP_REPO_FOLDER"_inprogress "$LWIP_REPO_FOLDER"
--- a/ee/network/tcpip/Makefile
+++ b/ee/network/tcpip/Makefile
@@ -44,6 +44,7 @@ ps2api_OBJECTS = \
 	tcpip.o
 
 ps2api_IPV4 = \
+	acd.o \
 	icmp.o \
 	ip.o \
 	ip4.o \
@@ -112,6 +113,9 @@ $(EE_OBJS_DIR)api_msg.o: $(LWIP)/src/api/api_msg.c
 	$(EE_CC) $(EE_CFLAGS) $(EE_INCS) -c $< -o $@
 
 $(EE_OBJS_DIR)api_netbuf.o: $(LWIP)/src/api/netbuf.c
+	$(EE_CC) $(EE_CFLAGS) $(EE_INCS) -c $< -o $@
+
+$(EE_OBJS_DIR)acd.o: $(LWIP)/src/core/ipv4/acd.c
 	$(EE_CC) $(EE_CFLAGS) $(EE_INCS) -c $< -o $@
 
 $(EE_OBJS_DIR)icmp.o: $(LWIP)/src/core/ipv4/icmp.c
--- a/ee/network/tcpip/src/include/lwipopts.h
+++ b/ee/network/tcpip/src/include/lwipopts.h
@@ -41,17 +41,47 @@
    ------------------------------------
 */
 /**
- * MEM_LIBC_MALLOC==1: Use malloc/free/realloc provided by the C-library
- * instead of lwIP's internal allocator. Default is 0; enabled on EE
- * because newlib's malloc is already linked in and the heap pool is
- * cheaper than a duplicate lwIP heap.
+ * MEM_LIBC_MALLOC==1: use libc's malloc/free for lwIP's mem_malloc /
+ * mem_free (which serve PBUF_RAM allocations and a handful of other
+ * sites — DHCP options, ARP, DNS, slip/ppp/zepif which we don't build).
+ * Internally lwIP's pbuf_alloc(PBUF_RAM) computes the payload pointer
+ * with `LWIP_MEM_ALIGN(p + SIZEOF_STRUCT_PBUF + offset)`, which only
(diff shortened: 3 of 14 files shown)


uyjulian (approved):
lgtm

PR #840, opened 2026-05-11
fix: [libpthreadglue] tighten error handling in OS-abstraction layer

fjtrujy:
## Summary
- `pte_osThreadCreate` used to malloc the stack and call `CreateThread`, then NULL-check the stack and check `threadId < 0` *after* `__threadInfo[threadId]` was already written — a negative id caused an OOB write. The check for the `cancelSem` semaphore was also missing, and the failure paths leaked `cancelSem` and the stack.
- Reorder so each resource is validated immediately after it is allocated and cleaned up in reverse order on failure.
- Propagate `CreateSema` failures from `pte_osMutexCreate` and `pte_osSemaphoreCreate` as `PTE_OS_NO_RESOURCES` instead of returning a negative handle through `*pHandle` as if it were a valid semaphore.

## Related
- Companion pthread-embedded PR: ps2dev/pthread-embedded#2 — adds the NULL-handle rejection and `mutex9` test referenced in the test plan.

## Test plan
- [ ] Build `ee/libpthreadglue` cleanly.
- [ ] Run pthread-embedded test suite on PS2 — exercises thread/mutex/sema creation paths and now also the NULL-handle paths from the companion pthread-embedded PR.
- [ ] Smoke-test a sample that creates pthreads (e.g. a port that already uses libpthreadglue).

### DIFF
--- a/ee/libpthreadglue/src/osal.c
+++ b/ee/libpthreadglue/src/osal.c
@@ -199,7 +199,6 @@ pte_osResult pte_osThreadCreate(pte_osThreadEntryPoint entryPoint,
   static int threadNum = 1;
   void *pTls;
   s32 threadId;
-  pte_osResult result;
   ps2ThreadData *pThreadData;
 
   if (threadNum++ > MAX_PS2_UID) {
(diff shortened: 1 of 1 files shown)


uyjulian (approved):
Lgtm

PR #841, opened 2026-05-13
Add ioprpgen library

uyjulian:
Can be used at runtime

Tested working


rickgaiser:
Is it possible to change the example to create a usefull image and also reboot the IOP using that image? By adding or replacing 1 IRX to demonstrate the libraries intended use?

uyjulian:
I added a more complete example.

PR #842, opened 2026-05-13
Use root README.md as mainpage for doxygen

uyjulian:
Allows other files with name `README.md` to be parsed correctly


rickgaiser:
This changes how the doxygen output looks?
Do you have a link to how it looks after this change?

uyjulian:
It mainly adds the missing documentation, but everything else will remain as is.

I don't have a link, but it can be run locally.

uyjulian:
The artifacts now has a generated doxygen output.

PR #843, opened 2026-05-13
Add dummy IOP module

uyjulian:
Mainly for IOPRP image no-op handling


sp193:
Aside from very specific software that need to tamper with the natural way of module loading, I don't suppose that there is much of a usecase for this.

fjtrujy (approved):

PR #846, opened 2026-05-14
Merge cdvd and xcdvd libraries

uyjulian:
Version will now be auto detected and correct RPC methods will be used

Tested working


rickgaiser:
LGTM, but I have a question.

I have a test program that runs all know IOPRP images, ranging from old BIOS to latest v3.1.0 IOPRP images. Can the library with this change automatically detect the version running? But does it also allow re-checking the version after IOP reboot or is it checked only once?

uyjulian:
It can detect the version running at `sceCdInit` since newer modules will return the version number while older modules will return 0 in place of the actual version number.
After IOP reboot use `sceCdInit` again to update the version number.

rickgaiser:
Great, thanks. Merging.

PR #848, opened 2026-05-15
Cleanup of inline assembly

uyjulian:
EE sync, cop0, and gp register access have been refactored


rickgaiser:
LGTM, but it's a big refactor. @fjtrujy ?

fjtrujy (approved):
Good clean-up!

Issue #851, opened 2026-05-22
Make fails with GCC-15/16


Pritchy96:
Hi all,

I seem to be having an issue compiling with GCC-15/16. Both give the same issue - I initially tried 16 as that's my system default, but then after seeing a [commit](https://github.com/ps2dev/ps2sdk/pull/767) for gcc-15 support I gave that a shot. Both give the issue:

```
[tom@tom-xps15 ps2sdk]$ make CC=gcc-15

...

src/romimg.c: In function ‘AddFile’:
src/romimg.c:471:27: error: initialization discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  471 |                 char* T = strrchr(fname, '.');
      |                           ^~~~~~~
cc1: all warnings being treated as errors

```

I've managed to get it moving by running `make cc=GCC-11`, but, again, it appears from the above PR that up to at least GCC-15 is supported. I'm very possibly missing something, mea culpa if so, I'm a touch rusty in terms of playing with GCC.

PR #854, opened 2026-05-24
Fix build for gcc-15/16

Pritchy96:
GCC-15/16 was still failing after #852, with a similar error to the one originally raised (#851):

```
make all
...
cc -O2 -Wall -Werror  -Isrc/ -Isrc/include -Iinclude/   -Isrc/ -Isrc/include -Iinclude/ -c src/anaarg.c -o obj/anaarg.o
src/anaarg.c: In function ‘analize_arguments’:
src/anaarg.c:131:60: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  131 |                                                         cp = strchr(opttable->option, ':');
      |                                                            ^
cc1: all warnings being treated as errors
```

This PR fixes that and GCC-15/16 now compiles cleanly.

### DIFF
--- a/tools/srxfixup/src/anaarg.c
+++ b/tools/srxfixup/src/anaarg.c
@@ -17,7 +17,7 @@
 int analize_arguments(const Opttable *dopttable, int argc, char **argv)
 {
 	Opt_strings *optstr;
-	const char *opt;
+	char *opt;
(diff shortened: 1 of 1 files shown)


uyjulian:
Lgtm

PR #855, opened 2026-05-24
Add fallback C code for inline ASM code

uyjulian:
For readability and future when MMI and VU0 get implemented in compiler

### DIFF
--- a/ee/graph/src/graph_mode.c
+++ b/ee/graph/src/graph_mode.c
@@ -94,6 +94,7 @@ static inline int __udelay(unsigned int usecs)
 
 	}
 
+#ifdef _EE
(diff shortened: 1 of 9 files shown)


ee/graph/src/graph_mode.c
```diff
@@ -106,6 +107,9 @@ static inline int __udelay(unsigned int usecs)
 				  ".set pop\n\t"
 				  :"=r" (loops_total)
 				  :"0" (loops_total), "r" (loops_end));
+#else
+	for (; loops_total < loops_end; loops_total += 1);
```
fjtrujy, line 111: I'm wondering if we should actually remove this piece of code... and use the existing `sleep` functionality...

PR #857, opened 2026-05-25
Convert kernel library asm s to c

uyjulian:
Makes it easier to read, especially with all the C preprocessor macros


fjtrujy:
I'm fine with this, @rickgaiser ?

ee/kernel/src/eenull/src/eenull.c
```diff
@@ -0,0 +1,30 @@
...
+
+	"\t" ".globl UModeCallbackDispatcher" "\n"
+	"\t" ".ent UModeCallbackDispatcher" "\n"
+	"\t" "UModeCallbackDispatcher:" "\n"
+	"\t" "\t" "lui $sp, 0x0008" "\n"
+	"\t" "\t" "jalr $v1" "\n"
+	"\t" "\t" "addiu $sp, $sp, 0x1fc0" "\n"
+	"\t" "\t" "addiu $v1, $zero, -8" "\n"
+	"\t" "\t" "syscall" "\n"
+	"\t" ".end UModeCallbackDispatcher" "\n"
```
rickgaiser, line 27: Is this more readable?

ee/kernel/src/eenull/src/eenull.s
```diff
@@ -1,26 +0,0 @@
...
-
-.globl UModeCallbackDispatcher
-.ent UModeCallbackDispatcher
-UModeCallbackDispatcher:
-	lui $sp, 0x0008
-	jalr $v1
-	addiu $sp, $sp, 0x1fc0
-	addiu $v1, $zero, -8
-	syscall
-.end UModeCallbackDispatcher
```
rickgaiser, line 23: Or is this more readable?

rickgaiser (commented):
I'm all for moving thing away from assembly and towards more readable 'C' code, but a large part of this PR just places complete assembly functions inside C and that makes no sense to me.

uyjulian:
The C version looks better to my eyes since it actually has syntax highlighting. (If there is a syntax plugin for mips asm+C preprocessor that would be nice)

I also separated each function into its own asm block for improved readability / separation, and added `.ent` / `.end` so that it will disassemble properly.

The C version can also participate in rearrangement and dropping of code (without needing `-ffunction-sections`).

rickgaiser:
I guess it's a matter of taste, becouse this makes it harder for me to read.

The line count has increased slightly (added 1.730, removed 1.651)
But each line is longer and therefore less readable:

from:
```
	jalr $v1
```
to:
```
	"\t" "\t" "jalr $v1" "\n"
```

Is there no better, more 'C' like, way to handle syscalls? Perhaps in examples from other devices or operating systems?

uyjulian:
Some example of linux mips syscalls: https://github.com/kraj/musl/blob/kraj/master/arch/mips/syscall_arch.h

Stuff writing to sp/k0/k1 may be too prone to breakage, so I'll leave those, but I plan to clean up the rest in successive PRs.  
Especially for syscalls, handling differing amount of parameters and clobbers is needed.

PR #860, opened 2026-05-30
Add xfromserv IOP library and associated RPC

uyjulian:
Access xfromman in the same manner as mcman with the new xfromserv IOP library.

Allows getting more detailed error codes


fjtrujy:
LGTM!

PR #862, opened 2026-06-01
Changes to mx4sio for ps2sdk sio2man client updates

uyjulian:
Tested working by myself

Fixes needed when using newest revision variant of mcman and padman.

Make hook simpler due to sio2man using a single semaphore  
Unhook intrman and loadcore when hooks no longer needed  
Add API for interrupt handler  
For loadcore and intrman hooks, only proceed when inner function succeeds

A major breaking change is that it now relies on the sio2man module from ps2sdk, which is smaller than the original SCE module and supports all known interfaces from 1.1 to 2.7.

### DIFF
--- a/iop/sio/mx4sio_bd/src/ioplib.c
+++ b/iop/sio/mx4sio_bd/src/ioplib.c
@@ -1,11 +1,26 @@
+/*
+# _____     ___ ____     ___ ____
+#  ____|   |    ____|   |        | |____|
+# |     ___|   |____ ___|    ____| |    \    PS2DEV Open Source Project.
+#-----------------------------------------------------------------------
+# Copyright ps2dev - http://www.ps2dev.org
+# Licenced under Academic Free License version 2.0
+# Review ps2sdk README & LICENSE files for further details.
+#
+# taken from MX4SIO driver for simplicity.
+# all credits go to maximus32
+*/
+
 #include "ioplib.h"
 #include <intrman.h>
 
-iop_library_t *ioplib_getByName(const char *name)
+int ioplib_iterateByName(const char *name, ioplib_libiterate_cb_t cb, void *userdata)
 {
     iop_library_t *libptr;
     int i;
+    int count;
 
+    count = 0;
     // Get first loaded library
     libptr = GetLoadcoreInternalData()->let_next;
     // Loop through all loaded libraries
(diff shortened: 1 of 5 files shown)


fjtrujy (approved):
LGTM!

rickgaiser:
Please test before merging.

fjtrujy:
Did you tested already??

uyjulian:
Not yet. I still need to figure out where I put my mx4sio

saildot4k:
FYI the mmceman driver was merged with this that @uyjulian PRd here:
[Changes to mmceman for ps2sdk sio2man client updates
](https://github.com/ps2-mmce/mmceman/pull/6)

Hope its not breaking anything because then as it stands does this mean that newer mmce drivers will not work until this is merged? Trying to update wLE for the new mmce drivers as rename and set card+channel was integrated which I want in wLE R3Z.

I hope to test soon since wLE R3Z allows both mmce and mx4sio in it (gated between IOP reboot) as both driver sets loaded together causes issues.

uyjulian:
Found my mx4sio. Tested working

saildot4k:
After this wle is no longer loading mx4sio, hangs on loading/browsing.
https://github.com/saildot4k/wLaunchELF_R3Z/actions/runs/27845964939

fjtrujy:
> After this wle is no longer loading mx4sio, hangs on loading/browsing. https://github.com/saildot4k/wLaunchELF_R3Z/actions/runs/27845964939

@uyjulian ??

saildot4k:
The fix for me was to use docker container from here which pcm720 informed me about:
https://github.com/ps2homebrew/ps2homebrew/pkgs/container/ps2homebrew

instead of 
https://github.com/ps2dev/ps2dev

Using the latter I still had to use mcman-1400/padman-1400 but then it broke on PSX (app wouldnt start). With the ps2homebrew container I can use mcman/padman and mmce/mx4sio work on both PS2 AND PSX.

uyjulian:
It sounds like you have an outdated ps2dev/ps2dev container

saildot4k:
> It sounds like you have an outdated ps2dev/ps2dev container

I was using latest, and for quite some time days after that GH action, until the switch to ghcr.io/ps2homebrew/ps2homebrew:main
Other problem is very hard to find this info such as the above container existing at all, or why/what differs from ps2dev/ps2dev container.

uyjulian:
Can you reproduce the issue with this minimal program? https://github.com/uyjulian/helloWorldPS2/tree/mx4sio_test

PR #867, opened 2026-06-17
Make EE data const or zero-initialized for cleaner (re-)initialization

uyjulian:
Some functions have added const arguments accordingly  
Initialization has been moved inside functions, or ctors if there is no conveient function  
Some global variables have been made local, static, or removed  
Sema, Thread, DmacHandler, IntcHandler, and fd handle 0 as invalid  

wLE tested working


fjtrujy:
Hello @uyjulian ,
I don't have anything against this PR; however, there are many conditions that were changed from `>= 0` to `> 0` and the opposite in other scenarios from `< 0` to `<= 0`, when creating semaphores, mutexes, threads, and similar.
Have you checked that all these return values were previously wrong?

uyjulian:
The return value on error is `-1`, but those functions never return the value `0`. I am using that property to initialize the variables, so I changed the conditions so that it can handle that case properly.

PR #868, opened 2026-06-17
Use iopheap functions for rom0_info

uyjulian:
Avoids fileio/file related functions inside rom0_info

### DIFF
--- a/ee/kernel/include/osd_config.h
+++ b/ee/kernel/include/osd_config.h
@@ -129,95 +129,99 @@ extern "C" {
  * @return Language value (See OSD_LANGUAGES above)
  */
 extern int configGetLanguage(void);
-extern int configGetLanguageWithIODriver(_io_driver *driver);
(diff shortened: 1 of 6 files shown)


ee/kernel/include/osd_config.h
fjtrujy, line 1: Maybe all these changes are better declared in the header as `static inline` instead of using `#define` macros?

fjtrujy:
Not sure if these changes will make the binary bigger... IIRC, I did this implementation to try to make the binary as small as possible, avoiding using the fileIO functions if we already have the fileXio in the binary

uyjulian:
These changes make wLaunchELF slightly larger by 256 bytes.  
However, in separate changes, I am working on separating fileio from kernel library.

ee/kernel/include/osd_config.h
uyjulian: Done

fjtrujy:
@uyjulian rebase, please!

PR #870, opened 2026-06-19
Changes to avoid embedded imgdrv in Iopreboot library

uyjulian:
Allows simplification of buildsystem (e.g. for CMake migration).

Tested working with ioprpgen_sample


fjtrujy:
It sounds like a good improvement for me. Anything worries you @rickgaiser ?

PR #873, opened 2026-06-20
Firewire updates

Florin9doi:


uyjulian:
Lgtm

Florin9doi:
This PR fixed the reads, but the orb’s direction should have been negated instead of always being set to write transaction. I’ll prepare a PR to correct that later.
https://github.com/ps2dev/ps2sdk/blob/f428d92211526b511d13fbabf3ea68d1ddbcc910/iop/iLink/IEEE1394_disk/src/scsi.c#L321

PR #876, opened 2026-06-21
bdm: added read16/write16 to support large disks

Florin9doi:
- more tests are needed

### DIFF
--- a/iop/fs/bdm/src/bdm.c
+++ b/iop/fs/bdm/src/bdm.c
@@ -51,7 +51,8 @@ void bdm_connect_bd(struct block_device *bd)
 {
     int i;
 
-    M_PRINTF("connecting device %s%dp%d size=%luMB id=0x%x\n", bd->name, bd->devNr, bd->parNr, (u32)bd->sectorCount / ((1000 * 1000) / bd->sectorSize), bd->parId);
+    M_PRINTF("connecting device %s%dp%d size=%luMB id=0x%x\n", bd->name, bd->devNr, bd->parNr,
+            (u32)(bd->sectorCount / ((1000 * 1000) / bd->sectorSize)), bd->parId);
 
     for (i = 0; i < MAX_CONNECTIONS; ++i) {
         if (g_mount[i].bd == NULL) {
--- a/iop/iLink/IEEE1394_bd/src/include/sbp2_disk.h
+++ b/iop/iLink/IEEE1394_bd/src/include/sbp2_disk.h
@@ -121,7 +121,7 @@ struct CommandDescriptorBlock
     struct sbp2_pointer DataDescriptor;
 
     u32 misc;
-    unsigned char CDBs[12];
+    unsigned char CDBs[16];
 } __attribute__((packed));
 
 struct SBP2Device
--- a/iop/usb/usbmass_bd/src/scsi.c
+++ b/iop/usb/usbmass_bd/src/scsi.c
@@ -24,6 +24,7 @@ typedef struct _inquiry_data
     u8 product[16];
     u8 revision[4];
 } inquiry_data;
+static_assert(sizeof(inquiry_data) == 36);
 
 typedef struct _sense_data
 {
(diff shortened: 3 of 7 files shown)


Florin9doi:
I tested using both of my SATA adapters capable of powering a 3.5 inch drive: a JMicron JMS567 and a JMS578, and both work as expected, including accessing data above the previous 32 bit limit. I have more adapters, but none of them are compatible with 3.5 inch drives, and I don't have any 2.5 inch drive larger than 2TB. Rebased and ready for review

iop/iLink/IEEE1394_bd/src/include/sbp2_disk.h
```diff
@@ -121,7 +121,7 @@ struct CommandDescriptorBlock
     struct sbp2_pointer DataDescriptor;
 
     u32 misc;
-    unsigned char CDBs[12];
+    unsigned char CDBs[16];
```
rickgaiser, line 124: What does this fix?

iop/fs/bdm/src/bdm.c
```diff
@@ -51,7 +51,8 @@ void bdm_connect_bd(struct block_device *bd)
 {
     int i;
 
-    M_PRINTF("connecting device %s%dp%d size=%luMB id=0x%x\n", bd->name, bd->devNr, bd->parNr, (u32)bd->sectorCount / ((1000 * 1000) / bd->sectorSize), bd->parId);
+    M_PRINTF("connecting device %s%dp%d size=%luMB id=0x%x\n", bd->name, bd->devNr, bd->parNr,
+            (u32)(bd->sectorCount / ((1000 * 1000) / bd->sectorSize)), bd->parId);
```
rickgaiser, line 55: This fixes a 64bit printing issue?

iop/usb/usbmass_bd/src/scsi.c
```diff
@@ -166,15 +210,39 @@ static int scsi_warmup(struct block_device *bd)
...
+
+    u64 sectorCount = bd->sectorCount;
+    U64_2XU32(sectorCount);
+    M_PRINTF("rc10 = 0x%08x%08x %u-byte logical blocks: (%lu MB / %lu MiB)\n", sectorCount_u32[1], sectorCount_u32[0], bd->sectorSize,
+        (u32)(bd->sectorCount / ((1000 * 1000) / bd->sectorSize)), (u32)(bd->sectorCount / ((1024 * 1024) / bd->sectorSize)));
+
+    // Cannot rely exclusively on READ_CAPACITY16 because it causes issues on smaller/older disks:
+    // - Kingmax Super Stick 4GB stops responding after receiving RC16
+    // - Kingmax Super Stick 8GB returns success, but the last LBA value is garbage (ff<<48)
+    // Fingers crossed - hopefully this approach works
```
rickgaiser: As far as I know READ CAPACITY(10)/READ(10)/WRITE(10) can be used for all drives. Only for > 2TB drives the (16) versions can be used after the drive told you it's possible by returning `sectorCount == 0xffffffff`. It seems like your code is doing that and it should be compatible with all older drives.
```
send READ CAPACITY(10)
if returned_LBA == 0xFFFFFFFF:
    send READ CAPACITY(16)   # get real 64-bit LBA + block size
    # READ(16)/WRITE(16) is possible for large LBA
else:
    # returned_LBA is the real max LBA
    # READ(10)/WRITE(10) is sufficient
```

rickgaiser (commented):
The SCSI files should be the same. Originally I wanted to have them in a separate library instead of this duplicated code but that's the way they are now. Please compare your changes to both SCSI files and try to keep them as much the same as possible.

iop/iLink/IEEE1394_bd/src/include/sbp2_disk.h
Florin9doi: It allows storage of the new 16-byte long commands: 
https://github.com/ps2dev/ps2sdk/pull/876/changes#diff-5c54b798a7dc5d91227744349fb24de21232d9bb4fec07892430d1770669c04dR148

iop/fs/bdm/src/bdm.c
Florin9doi: Yes, the u32 cast is made after division. Initially, only the 32 bits were used for division.

iop/usb/usbmass_bd/src/scsi.c
Florin9doi: Yes, that's what it does. I've removed the comment.

rickgaiser:
Ready for merge?

Florin9doi:
Yes, from my point of view

PR #878, opened 2026-06-24
elf-loader: add no-IOP-reset loader variant and NoReset entry point

KrahJohlito:
Adds LoadELFFromFileWithPartitionNoReset() via a runtime reset flag passed to the loader through argv, so an ELF launched from USB/BDM can still reach its boot device's storage modules (one blob; small change to the stub's argv layout).

Alternative to #877

### DIFF
--- a/ee/elf-loader/include/elf-loader.h
+++ b/ee/elf-loader/include/elf-loader.h
@@ -50,6 +50,8 @@ extern int LoadELFFromFile(const char *filename, int argc, char *argv[]);
  */
 extern int LoadELFFromFileWithPartition(const char *filename, const char *partition, int argc, char *argv[]);
 
+extern int LoadELFFromFileWithPartitionNoReset(const char *filename, const char *partition, int argc, char *argv[]);
+
 #ifdef __cplusplus
 }
 #endif
--- a/ee/elf-loader/src/elf.c
+++ b/ee/elf-loader/src/elf.c
@@ -54,13 +54,13 @@ static void wipe_bramMem(void) {
 #endif
 }
 
-int LoadELFFromFileWithPartition(const char *filename, const char *partition, int argc, char *argv[]) {
+static int LoadELFFromFileResetCommon(const char *filename, const char *partition, int reset, int argc, char *argv[]) {
 	u8 *boot_elf;
 	elf_header_t *eh;
 	elf_pheader_t *eph;
 	void *pdata;
 	int i;
-	int new_argc = argc + 2;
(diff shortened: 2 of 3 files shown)


fjtrujy:
Hello,
@KrahJohlito, please rebase your branch as there are now conflicts.
I like this PR more than #877, so let's keep this one and close the other

saildot4k:
I dont understand why you dont just call iop reset or not, instead of needing another loader variant. 

If you go down this path which i dont recommend, make sure the arg isnt passed to next elf but rather internal to the elf loader. 

In wle R3Z have a branch though that moves from ps2sdk elf loader to pcm720s osdmenu elf loader...so far its launching everything that wouldnt launch in other versions of wLE both p2homebrew version and isras
PCM720s elf loader consumes last to first, and handles internal args first then passes the rest as needed

KrahJohlito:
> I dont understand why you dont just call iop reset or not, instead of needing another loader variant.
> 
> If you go down this path which i dont recommend, make sure the arg isnt passed to next elf but rather internal to the elf loader.
> 
> In wle R3Z have a branch though that moves from ps2sdk elf loader to pcm720s osdmenu elf loader...so far its launching everything that wouldnt launch in other versions of wLE both p2homebrew version and isras PCM720s elf loader consumes last to first, and handles internal args first then passes the rest as needed

Because I would rather link to a lib than embed a loader.. feels cleaner, also being the sdk it should be at the developers discretion as to whether or not they reset iop (maybe certain apps need an iop reset? idk).. the flag is internal to the loader and not passed on to the loaded elf.. that being said if he wants to pr his loader changes I am sure they would be reviewed and considered.

saildot4k:
Yes often old homebrew expects inherited iop state. But new homebrew should avoid this and even reboot iop once the target is launched. Ie elf loader reboots iop before handoff. Next app reboots iop on load.

Being you dont know if next app needs iop reboot or not, user definable choice should exist, but default should be yes, reboot iop unless defined otherwise.

KrahJohlito:
> Yes often old homebrew expects inherited iop state. But new homebrew should avoid this and even reboot iop once the target is launched. Ie elf loader reboots iop before handoff. Next app reboots iop on load.
> 
> Being you dont know if next app needs iop reboot or not, user definable choice should exist, but default should be yes, reboot iop unless defined otherwise.

This pr achieves that, just call LoadELFFromFileWithPartitionNoReset() to not reset iop otherwise LoadELFFromFileWithPartition() will.. same loader lib not a build variant

PR #881, opened 2026-07-02
Changes to avoid strncpy in EE and host code

uyjulian:
`strncpy` has major flaws:
* Will zero initialize extra memory after NULL terminator
* Will not NULL terminate if written bytes is equal to buffer size


AKuHAK:
doh, isnt it supposed to be rather fixed in strncpy implementation?

uyjulian:
The strncpy in newlib is the C standard specification. The implementation is correct, but the specification has footguns

AKuHAK:
got it, thanks

PR #885, opened 2026-07-05
SIF DMA alignment cleanup

uyjulian:
Makes it easier to determine which buffers are used for SIF DMA

### DIFF
--- a/common/include/fileXio.h
+++ b/common/include/fileXio.h
@@ -21,6 +21,7 @@
 
 #include <tamtypes.h>
 #include <iox_stat.h>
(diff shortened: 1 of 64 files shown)


fjtrujy:
Hi @uyjulian,
Nice improvement, however, I have a question, there are some variables that now don't have any alignment, are these changes right?

uyjulian:
Those changes are correct, as the alignment has been moved to the type/structure so it applies when the type is used. Also, if alignment is applied to a member of the struct, the structure inherits that alignment.

fjtrujy (approved):

PR #904, opened 2026-07-23
fix(acmeme): zeroed irx status to avoid init abort

israpps:
before this change ACMEME.IRX returns -16 all the time bc of  acMemeModuleStatus() check on startup


uyjulian:
Lgtm

PR #905, opened 2026-07-24
add acram module with 128mb support

israpps:
based on wangan midnight, the only game to use double RAM64 PCB

this driver only makes sense for SYSTEM246 Rack A


uyjulian:
Lgtm

PR #906, opened 2026-07-26
scr_printf: implement color management via custom escape code

israpps:
this PR makes scr_printf coloring simpler, via a custom escape sequence

this is a hybrid between actual ansi escapes, and windows `color` command from cmd.exe just to keep impl small and the sequence as small as posible

there are 16 default colors (that the program can customize)

the sequence is \x1b followed by two hex chars

to avoid C compiler messing up, its better to use the octal version of the escape

as an example:
```c
scr_printf("hello people \03304 This is red text! \0330F Now back to white!\n");
```

### DIFF
--- a/ee/debug/src/scr_printf.c
+++ b/ee/debug/src/scr_printf.c
@@ -13,6 +13,7 @@
  * EE UGLY DEBUG ON SCREEN
  */
 
+#include <ctype.h>
 #include <stdio.h>
 #include <tamtypes.h>
 #include <sifcmd.h>
@@ -26,6 +27,8 @@ static short int X, Y;
 static const short int MX = 80, MY = 40;
 static u32 bgcolor, fontcolor, cursorcolor;
 static short int cursor;
+
+static u32 defcols[16];
 
 struct t_setupscr
 {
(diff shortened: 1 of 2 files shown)


ee/debug/src/scr_printf.c
```diff
@@ -128,6 +148,36 @@ void scr_setfontcolor(u32 color)
...
+
+void scr_change_defaultcolor(int index, u32 color) {
+    defcols[index] = color;
+}
+
+void scr_setfontcolorescape(char n1, char n2)
+{
+    int bg = hex2int(n1);
+    int fg = hex2int(n2);
+    printf("changing col t0 %c %c (%X %X)\n", n1, n2, bg, fg);
```
uyjulian: Leftover debug

ee/debug/src/scr_printf.c
```diff
@@ -27,6 +28,25 @@ static const short int MX = 80, MY = 40;
 static u32 bgcolor, fontcolor, cursorcolor;
 static short int cursor;
 
+static u32 defcols[16] = {
```
uyjulian: Move the initialization of this array to inside `init_scr` to move it to `.bss` section

uyjulian:
Lgtm

PR #907, opened 2026-07-28
Add definitions for ATA vendor specific commands to command table

uyjulian:
For Western Digital Marvell and FC1307 based platforms


sp193:
What is/are the usecase(s) for this change? If the commands are vendor-specific, when would they be used?

uyjulian:
The use case is to flash custom firmware on these devices without needing to take them out or desolder (for 70k) them.

Issue #908, opened 2026-07-28
USBD rewrite (b1f7ff96) breaks Crash Bandicoot: Wrath of Cortex over USB in OPL


horizonends:
### Summary

After the **2024-09-04 USBD feature update** (`b1f7ff96`), Open PS2 Loader builds that embed the current `usbd_mini` / `usbmass_bd` IRX fail to run **Crash Bandicoot: The Wrath of Cortex** from USB on real hardware.

Rebuilding the same modern OPL with USB IRX from **immediately before** that rewrite (`2dc6b32f`) restores normal behaviour.

This looks like a **ps2sdk USBD (and/or usbmass_bd pairing) regression**, not an OPL game-code bug.

### Symptoms (hardware)

On broken (post-rewrite) USB stack:
- Load to game menu can take **up to ~5 minutes** (good case is ~**10 seconds**)
- Starting play then goes to a **black screen**

On pre-rewrite USB stack (same OPL otherwise):
- ~10s to menu, **playable**

### Reproduction

- Consoles: PAL slim PS2s (SCPH-70003 / 75003 / 77003)
- Device: **USB** ISO
- Games: NTSC `SLUS_202.38` and PAL `SLES_503.86` (both fail the same way)
- Host loader: [Open PS2 Loader](https://github.com/ps2homebrew/Open-PS2-Loader) (embeds SDK IRX at build time)

OPL stock bisect (same USB/ISOs):
- Last good OPL Beta build era: **Beta-2125**
- First bad: **Beta-2127**
- That window lines up with OPL builds that baked in post-rewrite USBD

### Confirmed isolation

| Build | Result |
|-------|--------|
| Modern OPL + **current** SDK `usbd_mini` / `usbmass_bd` | Broken (slow load → black screen) |
| Modern OPL (Beta-2245 tree) + USB IRX from ps2sdk **`@2dc6b32f`** (pre-`b1f7ff96`) | **Works** (~10s, playable) |
| Stock OPL Beta-2245 | Broken |

So only swapping the embedded USB IRX back to pre-rewrite fixes it.

### Suspected commit

- Introduced: [`b1f7ff96`](https://github.com/ps2dev/ps2sdk/commit/b1f7ff96) — *USBD feature update* (2024-09-04)  
  (“usbd 1.2 API now exposed, module unloading, isochronous scheduling”)
[...]

PR #661, opened 2024-09-04
usbd feature update

uyjulian:
Most notable feature is module unloading.

My tests consist of wLE loading itself.

PR #911, opened 2026-08-07
usbd_mini: restore FreeUsbd for WoC USB regression (#908)

horizonends:
## Summary
- Restore **FreeUsbd** (ps2sdk `@2dc6b32f`, pre-`b1f7ff96`) as the implementation of **`usbd_mini` only**
- Leave full `iop/usb/usbd` on the SCE rewrite (HID report descriptors, multi-isochronous, etc.)
- Keep export table at **usbd 1.2** with stubs for `usbdReboot` / `sceUsbdGetReportDescriptor` / `sceUsbdMultiIsochronousTransfer` (`USB_RC_NOSUPPORT` / no-op)

Fixes the hardware regression in [#908](https://github.com/ps2dev/ps2sdk/issues/908): Crash Bandicoot: Wrath of Cortex over USB in OPL (embeds `usbd_mini`). Symptom: mid-transfer hang with `fail_count=0` (TD submitted, completion never wakes). Standalone usbstress does not reproduce.

### Isolation
| usbd | usbmass | WoC USB |
|--|--|--|
| rewrite | modern | fail |
| FreeUsbd (pre-`b1f7ff96`) | old | **works** |
| FreeUsbd | modern | **works** |

So modern `usbmass_bd` is fine; the rewrite’s `usbd_mini` path is not. Many FreeUsbd-vs-rewrite A/Bs on the rewrite were tried (HCD bring-up, WDH/DoneHead, dual TD queues, embedded ED, hub flush, SetAddress timing, attach/`transferedBytes`, timer `delay-=1`, etc.) — none restore play after the cutscene. Dead list on #908.

This does **not** claim to fix the rewrite; it restores the known-good driver for the mini module OPL actually ships.

### Related
- OPL: https://github.com/ps2homebrew/Open-PS2-Loader/issues/1751
- OPL pin PR (workaround until this lands): https://github.com/ps2homebrew/Open-PS2-Loader/pull/1752

## Test plan
- [x] `make -C iop/usb/usbd_mini` builds
- [x] OPL Beta-2245 tree + this `usbd_mini` + SDK `usbmass_bd_mini` → `OPNPS2LD-freeusbd-mini.ELF` staged for hardware
- [ ] Hardware: WoC USB past cutscene on PAL slim (same as known-good old-usbd pin)
- [ ] Spot-check USB mass still enumerates in OPL menu


horizonends:
Hardware confirm: OPL with this `usbd_mini` + current SDK `usbmass_bd_mini` — Crash Bandicoot: Wrath of Cortex over USB on PAL slim plays past the cutscene (same as the known-good old-usbd pin).

sp193:
Normally, the app is designed and built with a specific SDK version in mind, which is not really what our homebrew world does because our SDK contains modules that are based on modules from different SDK versions. Sony does not change the default IOP module versions, even when they release new PS2 models.

But I don't think we should be maintaining another project's dependencies in the SDK. Otherwise, we could have N copies of the source code, for N homebrew software.

I excused myself from working on PS2 homebrew software for about 8 years now. I recall that the USB module code was originally in OPL's repository. Then the new maintainers probably relocated them here, to avoid duplicating the code - which IMO is a good move. But if we are going to undo that, then the correct way may be to store the code back in OPL's repository, not here.

A greater concern may be whether this is a sign of instability.

When I fixed the original USBD for correct operation on some console (I forgot what), I did that my comparing the logic against the assembly of the similar module. One problem is that we're not the hardware developers and built the module without Sony's support. Any hardware and software bugs or limitations are undocumented. So my solution was to make the SDK libraries work as similarly as possible to the original, to avoid such issues.

rickgaiser:
There have been issues ever since this PR (https://github.com/ps2dev/ps2sdk/pull/661) from 2 years ago.

It's a complete rewrite / reverse engineered usbd module. Perhaps it's better to revert the entire commit, becouse the old usbd was carefully fixed over many years, but the new usbd seems totally unsupported. @uyjulian what's your thoughts?

For neutrino I use my own ps2sdk neutrino branch where I revert (https://github.com/ps2dev/ps2sdk/commit/309ebcdb3c53edf34f153e1b92d02b2ae4c3f5e9) it.

sp193:
The only thing I really didn't like about that PR, was how the structure changed so much. So much that we cannot easily compare the 2, to have an idea what the differences are. Unless this actually meant that Sony possibly did a major rewrite too (which is possible, seeing the way MCMAN worked before and after some point at SDK 2.x).

I wrote the "instability" remark because the author mentioned some hang on some internal logic, in the OPL-side PR. However, not every malfunction necessarily means we wrote bad code because games could have undetected bugs too.

If @uyjulian strongly believes that this new structure and version is very much better to have, I guess we can still try to move forward with it. We just need to be sure that the I/O logic (previously in hcd.c and a little more) are correct. The PS2 has a OHCI USB controller so most of the USB logic is in hardware.

sp193:
If we deem that this version of USBD is good and the issue lies with this game, then I would say either:
1. The game should be debugged and a patch should be made.
2. OPL should go back to having its own version of USBD.

uyjulian:
It fixed some things (e.g. games requiring the isosynchronous interfaces) but also broke some things.

It would be nice to figure out what exactly is the issue. I suspect thread stack space due to the added features but not entirely sure. Needs further debugging. (but my DTL-T system doesn't work consistently right now so slowly working on improving alternatives)

10522:
@rickgaiser @sp193 @uyjulian @horizonends i don't have this regression (more details [here](https://github.com/ps2dev/ps2sdk/issues/908#issuecomment-5155348825)).

horizonends:
Thanks for the feedback — agreed this shouldn’t turn into “N copies of USBD for N apps” in the SDK.

Quick clarification on what we proved on hardware (PAL slim, USB WoC):
- Regression tracks **modern `usbd_mini` alone** (hybrid: FreeUsbd/`pre-b1f7ff96` usbd + **current** `usbmass` works)
- Hang looks like **TD submitted, completion never wakes** (`fail_count=0`); usbstress does not repro
- Cheap FreeUsbd-vs-rewrite A/Bs on the rewrite are exhausted (dead list on #908)

Given that, I’m fine closing **this** PR if the preferred paths are:
1. **Revert the rewrite** (as @rickgaiser suggested / Neutrino already does), or keep chasing a real rewrite fix, and/or
2. **OPL vendors FreeUsbd/`usbd_mini`** (ps2homebrew/Open-PS2-Loader#1752) until the rewrite is solid

This PR was only meant as a narrow SDK-side workaround for the module OPL embeds. If maintainers prefer (1) or (2), say which and I’ll close #911 and follow that.

@uyjulian — re stack space: we did try **16 KiB MINI HCD/CB stacks** on hardware; still black after cutscene. Happy to try other stack/debug experiments if you have a specific build in mind.

PR #913, opened 2026-08-27
atad: fix zero-progress loop for 65536-sector LBA48 transfers

LAP87:
`ata_device_sector_io64` narrows the per-iteration transfer length to `u16` before using it for the loop accounting. A request of 65536 sectors or more therefore computes `len == 0`, passes `blkcount` 0 to `sceAtaExecCmd`, and advances `buf`/`lba`/`nsectors` by zero — an infinite loop that never makes progress.

This change keeps the iteration length in a `u32` (`chunk`) and only narrows to `u16` for the sector-count register, where 0 legitimately means 65536. The 28-bit path is expressed the same way for consistency; its behaviour is unchanged (max 256).

Found while characterizing large GPT/exFAT drives on clone network adapters (measurements and test tooling: https://github.com/LAP87/ps2-gamestar-pio), but the defect is independent of adapter type — any BDM/APA caller requesting ≥ 32 MiB in one call hits it.

### DIFF
--- a/iop/dev9/atad/src/ps2atad.c
+++ b/iop/dev9/atad/src/ps2atad.c
@@ -1026,12 +1026,14 @@ int ata_device_sector_io64(int device, void *buf, u64 lba, u32 nsectors, int dir
     USE_SPD_REGS;
     int res = 0, retries;
     u16 sector, lcyl, hcyl, select, command, len;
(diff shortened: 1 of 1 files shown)


uyjulian:
Looks good to me

PR #914, opened 2026-08-27
atad: allow opting out of the DVRP workaround for clone adaptors

LAP87:
The DVRP (PSX) workaround activates whenever `SPD_CAPS_DVR` is set. The widespread third-party/clone network adaptors (Spartan-3 `XC3S50AN` FPGA + JMicron `JM20330` bridge — sold as GameStar, SSPH-10350, BitFunx and similar) assert that bit even though they are not DVRP-based: verified on real hardware, they address 48-bit LBA correctly over PIO (`READ SECTOR EXT` at LBA `0x10000000`+ returns CRC-verified correct data). With the workaround active their reported capacity is truncated to 28-bit LBA and large GPT drives are rejected outright — this is the common "big drive not detected on a clone adapter" failure.

This adds an `ATA_DISABLE_DVRP_WORKAROUND` build option (Makefile pattern matching the existing `ATA_GAMESTAR_WORKAROUND`) so builds targeting such adaptors can opt out. Default behaviour is completely unchanged, and the comment block documents why the escape hatch exists.

Runtime discrimination between a real DVRP and a clone asserting the bit would be nicer than a build flag; happy to iterate toward that if anyone with DESR hardware can help characterize a reliable probe. Measurements, boundary-test tooling and the full clone characterization live at https://github.com/LAP87/ps2-gamestar-pio.


uyjulian:
Please try out https://github.com/ps2dev/ps2sdk/pull/915 to see if it a solution to the issue

LAP87:
Superseded by #915, whose runtime detection is now hardware-verified on a clone adapter (see the test report there) — a better solution than this build-time flag. Thanks @uyjulian!

Issue #924, opened 2026-09-21
ERL: four defects - a vacuous HI16/LO16 guard, a dropped export, three dangling dependencies, six erl_id/filename mismatches


rixnobis:
Four defects in the EE ERL machinery, found while porting ERL to the PS1. All re-verified against `d317f8f0` today; `ee/erl/src/erl.c` has the same blob sha at `61a4b3f1` and `d317f8f0`, and none of the commits in between touch `ee/erl/`, `ee/erl-loader/` or any `erl-support.c`.

**1. `ee/erl/src/erl.c:856` reads the wrong struct, which makes the HI16/LO16 pairing guard vacuous.**

```c
851             sym_n = reloc.r_info >> 8;
...
855             next_reloc = *((struct elf_reloc_t *) (reloc_section + ((j+1) * sec[i].sh_entsize)));
856             next_sym_n = reloc.r_info >> 8;
```

`next_reloc` is fetched on line 855 and never consulted for the symbol index; line 856 is a byte-identical copy of 851. The consequence is at line 883:

```c
882             (has_next_reloc && ((next_reloc.r_info & 255) == R_MIPS_LO16)) &&
883             (sec[sym[sym_n].st_shndx].sh_addr == sec[sym[next_sym_n].st_shndx].sh_addr) &&
```

883 is only reached when `has_next_reloc` is true, which is the same path on which 856 ran, and `reloc` is not written in between - so `next_sym_n == sym_n` always and the section-address comparison is unconditionally true. The paired-section check does nothing. No uninitialised read: the short-circuit covers it.

Last touched by `8f16bbc0` ("fix: recalculate addresses when lower bits are greater than 0x8000", 2025-03-19) on both lines.

**2. `ee/erl-loader/src/build-exports.sh` drops the last symbol from both the forward declarations and the export table.**

```sh
 6  sort -n |
 7  awk ' { t[NR] = $0 }
 9        for (i = 1; i < NR; i++)      # forward decls
15        for (i = 1; i < NR; i++)      # export_list[] body
```

`t[NR]` is written and never read, so whichever symbol sorts last gets neither a prototype nor an `export_list[]` entry. Separately, line 6 applies `sort -n` to C identifiers coming out of `awk '{print $8}'`, so every key sorts as 0 and the order is whatever order the input arrived in. *Which* symbol gets dropped is therefore not stable.

**3. Three declared ERL dependencies have no provider anywhere in the tree.**

[...]

PR #925, opened 2026-09-22
erl: use next_reloc for the paired symbol index, and stop dropping the last export

rixnobis:
Defects 1 and 2 from #924. Three lines, both verified at source on `d317f8f0`.

`erl.c` line 856 fetched `next_reloc` and then read `reloc.r_info` for the symbol index, so `next_sym_n` always equalled `sym_n` and the `sec[sym[sym_n].st_shndx].sh_addr == sec[sym[next_sym_n].st_shndx].sh_addr` term at line 883 was unconditionally true. The HI16/LO16 pairing guard did nothing.

`build-exports.sh` stored `t[NR]` and iterated `i < NR` in both loops, so the last symbol got neither a forward declaration nor an `export_list[]` entry.

Correcting one thing I said in #924 while I was here: I claimed the `sort -n` on line 6 makes every key compare as 0 so the order is the input order, and that the dropped symbol is therefore unstable. That is wrong. GNU sort falls back to a whole-line comparison when the keys tie, so `sort -n` on identifiers gives the same order as plain `sort` - `printf 'zeta\nalpha\nmike\nbravo\n' | sort -n` yields `alpha bravo mike zeta`, and only `sort -n -s` preserves input order. So the drop is deterministic: it is always the lexically last symbol. That makes the bug more findable rather than less, and I have left the `sort -n` alone here since it is not doing any harm.

Defects 3 and 4 are untouched. For 4 I have your ruling that the buildsystem name is the intended one and will follow it.


fjtrujy:
@uyjulian ?

uyjulian:
Lgtm

